PR #2, opened 2024-05-02
gpio: tegra186: Fix tegra186_gpio_is_accessible() check

jamieNguyenNVIDIA:
The controller has several register bits describing access control information for a given GPIO pin. When SCR_SEC_[R|W]EN is unset, it means we have full read/write access to all the registers for given GPIO pin. When SCR_SEC[R|W]EN is set, it means we need to further check the accompanying SCR_SEC_G1[R|W] bit to determine read/write access to all the registers for given GPIO pin.

This check was previously declaring that a GPIO pin was accessible only if either of the following conditions were met:

  - SCR_SEC_REN + SCR_SEC_WEN both set

    or

  - SCR_SEC_REN + SCR_SEC_WEN both set and SCR_SEC_G1R + SCR_SEC_G1W both set

Update the check to properly handle cases where only one of SCR_SEC_REN or SCR_SEC_WEN is set.

Fixes: b2b56a163230 ("gpio: tegra186: Check GPIO pin permission before access.")

Acked-by: Thierry Reding <treding@nvidia.com>
(cherry-picked from commit d806f474a9a7993648a2c70642ee129316d8deff linux-next)


jamieNguyenNVIDIA:
Validated on CG4 by checking line 70:
```
$ sudo gpioinfo 0
...
	line  70:      "PL.02" "ACPI:Event"   input  active-high [used]
	line  71:      "PL.03"       kernel   input  active-high [used]
	line  72:      "PL.04"       kernel   input  active-high [used]
	line  73:      "PL.05"       unused   input  active-high 
```

In the broken case, this reports:
```
$ sudo gpioinfo 0
...
    line  70:      "PL.02"       kernel   input  active-high [used]
    line  71:      "PL.03"       kernel   input  active-high [used]
    line  72:      "PL.04"       kernel   input  active-high [used]
    line  73:      "PL.05"       unused   input  active-high
```

PR #10, opened 2024-06-06
NVMe/NVMeoF: Export nvfs register and unregister functions as GPL

sourabgupta3:
On linux kernel 6.6 and above, __symbol_get() on the registration functions from nvidia-fs was failing as a GPL modules are no longer allowed to __symbol_get() on non-gpl exported symbols. This change fixes that issue.


nvmochs:
Already pulled, closing PR.

❯ git lgo 0fb9df2f1465 -1
0fb9df2f1465 NVIDIA: SAUCE: NVMe/NVMeoF: Export nvfs register and unregister functions as GPL

❯ git br -r --contains 0fb9df2f1465| grep origin
  origin/22.04_linux-nvidia-6.8
  origin/24.04_linux-nvidia

FYI, confirmed this is also present in the 6.8 tech preview and the 6.11 kernels.

PR #13, opened 2024-06-06
NFS: Export nvfs register and unregister functions as GPL

sourabgupta3:
On linux kernel 6.6 and above, __symbol_get() on the registration functions from nvidia-fs was failing as a GPL modules are no longer allowed to __symbol_get() on non-gpl exported symbols. This change fixes that issue for nfs.


nvmochs:
Already pulled, closing PR.

❯ git lgo 55eef858801d -1
55eef858801d NVIDIA: SAUCE: NFS: Export nvfs register and unregister functions as GPL

❯ git br -r --contains 55eef858801d | grep origin
  origin/22.04_linux-nvidia-6.8
  origin/24.04_linux-nvidia

FYI, verified these are also present in 6.8 tech preview and 6.11 kernels.

PR #31, opened 2024-11-13
cpufreq/cppc: Don't compare desired_perf in target()

jamieNguyenNVIDIA:
There is a corner case where the desired_perf is exactly same as the old perf, but the actual current freq is not.

This happens during S3 while the cpufreq governor is set to powersave. During cpufreq resume process, the booting CPU's new_freq obtained via .get() is the highest frequency, while the policy->cur and cpu->perf_ctrls.desired_perf are set to the lowest level (powersave governor). This causes the warning: "CPU frequency out of sync:", and the cpufreq core sets policy->cur to new_freq.

Then the governor->limits() calls cppc_cpufreq_set_target() to configures the CPU frequency and returns directly because the desired_perf converted from target_freq is same as the cpu->perf_ctrls.desired_perf and both are the lowest_perf.

Since target_freq and policy->cur have been already compared in __cpufreq_driver_target(), there's no need to compare them again here.

Drop the comparison.

[ Viresh: Updated commit message / subject ]

(cherry picked from commit 90e4ed6bb02ad93663f17411d17e8e714a765a6b)

### DIFF
--- a/drivers/cpufreq/cppc_cpufreq.c
+++ b/drivers/cpufreq/cppc_cpufreq.c
@@ -291,15 +291,10 @@ static int cppc_cpufreq_set_target(struct cpufreq_policy *policy,
 	struct cppc_cpudata *cpu_data = policy->driver_data;
 	unsigned int cpu = policy->cpu;
 	struct cpufreq_freqs freqs;
-	u32 desired_perf;
 	int ret = 0;
 
-	desired_perf = cppc_khz_to_perf(&cpu_data->perf_caps, target_freq);
-	/* Return if it is exactly the same perf */
-	if (desired_perf == cpu_data->perf_ctrls.desired_perf)
-		return ret;
-
-	cpu_data->perf_ctrls.desired_perf = desired_perf;
+	cpu_data->perf_ctrls.desired_perf =
+			cppc_khz_to_perf(&cpu_data->perf_caps, target_freq);
 	freqs.old = policy->cur;
 	freqs.new = target_freq;
 


jamieNguyenNVIDIA:
Internal bug: b4915974
Internal JIRA: DGX-10617

Tested on UTSkinnyJoe-DVT-46 and lego-cg1-qct-034 using the following script (which starts stress tests and then checks for frequency readings that are less than 70% of nominal):
```
#!/bin/bash

trap 'signal_term; exit 1' SIGTERM SIGINT SIGQUIT

function signal_term()
{
    echo "Kill bg pids..."
    for bg_pid in $bg_pids; do
        kill -9 $bg_pid > /dev/null 2>&1
    done
}

bg_pids=""
errcode=0
max_MHz=`lscpu | awk '/CPU max MHz/ {print $NF}' | awk -F '.' '{print $1}'`
# if we use "$[max_MHz*0.7]" , we need install bc.
exp_MHz=$[max_MHz*7]
core_num=`lscpu | grep "^CPU(s):" | awk '{print $NF}'`
echo "Start stress process..."
for i in `seq 1 $core_num`; do
    taskset -c $[i-1] yes > /dev/null &
    #yes > /dev/null &
    bg_pids="$bg_pids $!"
done
echo "Start stress process done."

echo "Waiting for 3s ..."
sleep 3

echo "Start Check CPU Freq..."
for i in `seq 1 $core_num`; do
    cpu_freq=`cat /proc/cpuinfo | awk '/cpu MHz/ {print $NF}' | awk -F '.' '{print $1}' | sed -n "$i"p`
    # for arm cpu
    [ -z "$cpu_freq" ] && cpu_freq=`cat /sys/devices/system/cpu/cpu$[i-1]/cpufreq/scaling_cur_freq` && cpu_freq=$[cpu_freq/1000]
    # we use $[max_MHz*7] before, so ...
    if [ -z "$cpu_freq" ] || [ "$[cpu_freq*10]" -lt $exp_MHz ]; then
        echo "CPU $i[1-$core_num] freq check fail! $cpu_freq < $max_MHz * 0.7"
        let errcode++
    else
        
        echo "CPU $i: $cpu_freq > $max_MHz * 0.7"
    fi
done

signal_term

if [ $errcode -gt 0 ]; then
    echo "CPU Freq check fail !"
else
    echo "CPU Freq check pass !"
fi
exit $errcode
```

clsotog (approved):
Acked-by: Carol L Soto (csoto@nvidia.com)

nvmochs (approved):
Acked-by: Matthew R. Ochs <mochs@nvidia.com>

PR #60, opened 2025-02-25
arm: AMU-based average frequency readings (for 6.11 HWE)

jamieNguyenNVIDIA:
Backport the following series:
```
[1]: https://lore.kernel.org/all/20250131162439.3843071-1-beata.michalska@arm.com/
[2]: https://lore.kernel.org/all/20250131155842.3839098-1-beata.michalska@arm.com/
[3]: https://lore.kernel.org/all/20240827154818.1195849-1-ionela.voinescu@arm.com/
[4]: https://lore.kernel.org/all/20250220091015.2319901-1-beata.michalska@arm.com/
```
[1] and [2] are what we really want; [3] we need to take as a dependency, and [4] we should take as a fix.

PR #67, opened 2025-03-11
Patches to support CPPCv3 Autonomous Frequency Control

nvidia-ishah:
Patch series to enable support for the CPPCv3 optional feature of Autonomous Frequency Control.
This will allow for the hardware to control CPU Frequency. The kernel will be able to input a preference for "performance" or "energy savings" via the "energy_perf" sysfs knob.

Additionally bringing over patches to support the "cpuinfo_avg_freq" sysfs knob by Beata from ARM. This is part of the same patchseries as this is the recommended knob to read back frequency with hardware-based cpu-frequency control.


nvmochs:
These patches are already carried in 6.11 (backported with [PR 60](https://github.com/NVIDIA/NV-Kernels/pull/60)):

2dfd12502a24 NVIDIA: SAUCE: arm64: Update AMU-based freq scale factor on entering idle
7abdf3cdf709 NVIDIA: SAUCE: arm64: Provide an AMU-based version of arch_freq_get_on_cpu
1ba538698d3f arm64: amu: Delay allocating cpumask for AMU FIE support
a06e2d2fa82b NVIDIA: SAUCE: cpufreq: Introduce an optional cpuinfo_avg_freq sysfs entry
488afa30f410 NVIDIA: SAUCE: cpufreq: Allow arch_freq_get_on_cpu to return an error

So they can be excluded.

nvidia-ishah:
ACK nvmochs.
Should I remove them from the branch, or is this a note for future reviewers?

nvmochs:
> ACK nvmochs. Should I remove them from the branch, or is this a note for future reviewers?

Leave them for now, we'll figure out if we need to rebase after reviewing. @clsotog and I are reviewing this branch. I'm making notes as I go along.

nvmochs:
Looking more closely at this branch, we only need to review these 5 SAUCE patches:
8af0daab270a NVIDIA: SAUCE: cpufreq: CPPC: Add cppc_cpufreq_epp instance for Autonomous mode
ac91475a5cdc NVIDIA: SAUCE: Documentation: ACPI: add autonomous mode ctrls info in cppc_sysfs.txt
0a9a49186040 NVIDIA: SAUCE: ACPI: CPPC: support updating epp, auto_sel and {min|max_perf} from sysfs
d127e15e032b NVIDIA: SAUCE: ACPI: CPPC: expand macro to create store acpi_cppc sysfs node
676012eae049 NVIDIA: SAUCE: ACPI: CPPC: add read perf ctrls api and rename few existing

All of the other patches are already present in Canonical's [nvidia-6.11-next noble branch](https://git.launchpad.net/~canonical-kernel/ubuntu/+source/linux-nvidia/+git/noble/log/?h=nvidia-6.11-next).

clsotog:
I tried the patches in a grace system and I see a stack trace when enabling cppc_cpufreq.auto_sel_mode=1
[   17.493260] ------------[ cut here ]------------
[   17.493264] UBSAN: shift-out-of-bounds in drivers/acpi/cppc_acpi.c:1261:9
[   17.500403] shift exponent 64 is too large for 64-bit type 'long unsigned int'
[   17.500406] CPU: 7 UID: 0 PID: 1248 Comm: (udev-worker) Not tainted 6.11.0+ #2
[   17.500408] Hardware name:  /P3880, BIOS 24032321 20240324
[   17.500409] Call trace:
[   17.500410]  dump_backtrace+0xa4/0x150
[   17.500416]  show_stack+0x24/0x50
[   17.507990]  dump_stack_lvl+0x80/0x140
[   17.507994]  dump_stack+0x1c/0x38
[   17.507995]  __ubsan_handle_shift_out_of_bounds+0x154/0x260
[   17.508001]  cpc_write+0x468/0x490
[   17.508005]  cppc_set_enable+0x90/0x1b8
[   17.508007]  cppc_cpufreq_init+0xdc/0xff8 [cppc_cpufreq]
[   17.508012]  do_one_initcall+0x64/0x388
[   17.508013]  do_init_module+0xa4/0x2d0
[   17.508017]  load_module+0x2488/0x25b8
[   17.508019]  init_module_from_file+0x98/0x118
[   17.508020]  __arm64_sys_finit_module+0x27c/0x378
[   17.508022]  invoke_syscall+0x74/0x128
[   17.508024]  el0_svc_common.constprop.0+0x114/0x140
[   17.508025]  do_el0_svc+0x28/0x58
[   17.508025]  el0_svc+0x44/0x1b0
[   17.508028]  el0t_64_sync_handler+0x148/0x158
[   17.508029]  el0t_64_sync+0x1b0/0x1b8
[   17.508036] ---[ end trace ]---

Maybe we need upstream patch:60949b7b805 ACPI: CPPC: Fix MASK_VAL() usage
Not sure if we will need this one too:2388b266c9fcc ACPI: CPPC: Fix _CPC register setting issue since its a fix for the previous one.

nvmochs:
8af0daab270a NVIDIA: SAUCE: cpufreq: CPPC: Add cppc_cpufreq_epp instance for Autonomous mode

Commit message:
Engergy -> Energy

I’m confused, cppc_auto_sel is added to kernel parameter documentation but then no support is added for the parameter? Is it supposed to be “auto_sel_mode”?

Nit: comment blocks in cppc_cpufreq_epp_cpu_init() need improved formatting

cppc_cpufreq_epp_cpu_init() could likely be refactored to share code with cppc_cpufreq_cpu_init(), but will let that slide for now.

Nit: In cppc_cpufreq_epp_set_policy(), just return cppc_cpufreq_epp_update_perf();

In cppc_cpufreq_epp_cpu_exit(), where is the complementary call site to cppc_cpufreq_put_cpu_data(policy)?

Why is cppc_cpufreq_auto_sel_enable() taking a boolean parameter? Is there a case when you would call it with false? Also, if it is to stay, please name it something other than the static global used for the module parameter.



ac91475a5cdc NVIDIA: SAUCE: Documentation: ACPI: add autonomous mode ctrls info in cppc_sysfs.txt

Nit: increase /decrease -> increase / decrease

0a9a49186040 NVIDIA: SAUCE: ACPI: CPPC: support updating epp, auto_sel and {min|max_perf} from sysfs

Commit message fixups:
updation -> update
Engergy -> Energy


store_min_perf are very similar and probably could be refactored into a common routine with a named front-end, but I’m willing to let that slide.

Why is store_energy_perf() using u64 for epp and input (and related code)? The energy_perf member is a u32, no?

In cppc_set_epp_perf(), in the new !CPC_IN_PCC leg, is it correct to only write auto_sel_reg and epp_set_reg when both are supported? Or should they be checked/written independently?

In cppc_get_auto_sel_caps(), don’t you need to check CPC_SUPPORTED(auto_sel_reg) in the CPC_IN_PCC leg? Or maybe keep the check that was removed and don’t add the check in the !CPC_IN_PCC case?

How are perf_limited and auto_activity_window used? I don’t see support for them.



d127e15e032b NVIDIA: SAUCE: ACPI: CPPC: expand macro to create store acpi_cppc sysfs node

Which of these attributes are supposed to support rw?

Also, it looks like this only supports “show” and not “store” semantics?

For example, this will define a rw attribute, but where is the code for the store?

```
#define define_one_cppc_rw(_name)              \
+       static struct kobj_attribute _name =    \
+__ATTR(_name, 0644, show_##_name, store_##_name)
```

After looking at the next commit (0a9a49186040 NVIDIA: SAUCE: ACPI: CPPC: support updating epp, auto_sel and {min|max_perf} from sysfs), I see you are adding them there. I would update the commit message to clarify that while it enables support for a store node, it does not actually add any store nodes.






676012eae049 NVIDIA: SAUCE: ACPI: CPPC: add read perf ctrls api and rename few existing

There are other users of cppc_set_perf (drivers/cpufreq/amd-pstate.c) that need to be fixed up in the name change. Also, I think the name change should be a standalone patch.

Is there a reason why you feel it necessary to remove energy_perf from the caps structure? I would just leave it for now as it seems it may not have been implemented correctly (comparing caps and ctrls) and could be fixed someday. Or if you want to remove it, it should be in a standalone cleanup patch.

The logic in cppc_get_perf_ctrls looks correct. This should be added in a standalone patch.

clsotog:
Is there test results with this code? Maybe my lego system does not support this but when I did this 
lego-cg1-dvt-63:/sys/devices/system/cpu/cpu0/acpi_cppc# echo 1 > auto_sel
It crashed here store_auto_sel+0x5c let me debug it but maybe my guess is that cppc_cpufreq_driver is null so maybe need to redo more the function cppc_cpufreq_init.

clsotog:
Please add these 2 patches I mentioned 
60949b7b805 ACPI: CPPC: Fix MASK_VAL() usage
2388b266c9fcc ACPI: CPPC: Fix _CPC register setting issue
It will fix the UBSAN: shift-out-of-bounds I saw. 
Also at my system the cppc_set_enable is failing with -14 so need to redo the function cppc_cpufreq_init if cppc_cpufreq_auto_sel_enable fails maybe put a printk it fails and maybe allow to run cppc_cpufreq_driver.

nvmochs:
Since the branch will need to be updated to resolve issues mentioned above, please rebase it on top of the latest 24.04_linux-nvidia-6.11-next (HEAD should be b0eb77506189676728cfbee0094f791cd6c01ba5) to avoid this PR showing commits that are already present in the main branch.

nvidia-ishah:
@clsotog Grace systems do NOT support this feature. This requires compatible HW and ACPI definitions.
cppc_set_enable failing with -14 (EFAULT) is likely failing when we write to the set_enable address. That implies it is present but bad, since it is passing previous checks.
I would expect a failure & some log in dmesg to explain why, but this shouldn't work anyways - Grace ACPI tables shouldn't define this register.
But I will take your advice & pickup those patches to fix the UBSAN oob error.

> Since the branch will need to be updated to resolve issues mentioned above, please rebase it on top of the latest 24.04_linux-nvidia-6.11-next (HEAD should be https://github.com/NVIDIA/NV-Kernels/commit/b0eb77506189676728cfbee0094f791cd6c01ba5) to avoid this PR showing commits that are already present in the main branch.

Ack. I will include this as part of the update.

> Is there test results with this code? Maybe my lego system does not support this but when I did this
Yes, there are test results that can be shared offline. This was tested with a supported system.
Hardware support for this feature will be required on top of ACPI table changes.

I will address nvmoch's other comments as I update the code.

clsotog:
> @clsotog Grace systems do NOT support this feature. This requires compatible HW and ACPI definitions. cppc_set_enable failing with -14 (EFAULT) is likely failing when we write to the set_enable address. That implies it is present but bad, since it is passing previous checks. I would expect a failure & some log in dmesg to explain why, but this shouldn't work anyways - Grace ACPI tables shouldn't define this register. But I will take your advice & pickup those patches to fix the UBSAN oob error.
Even if lego-cg1 does not support this feature, the code is not handling correctly this situation. I think you need something like this 
if (auto_sel_mode || caps.auto_sel) {
                ret = cppc_cpufreq_auto_sel_enable(true);
                if (ret)
                {
                        pr_err("failed to cppc_cpufreq_auto_sel_enable setting cppc_cpufreq_driver\n");
                        current_cppc_cpufreq_driver = &cppc_cpufreq_driver;
                } else {
                        current_cppc_cpufreq_driver = &cppc_cpufreq_epp_driver;
                }
To avoid the crash I saw, when doing the echo.

nvidia-ishah:
> https://github.com/NVIDIA/NV-Kernels/commit/676012eae04942eacbadcdaeec71ea0f3f2e9568 NVIDIA: SAUCE: ACPI: CPPC: add read perf ctrls api and rename few existing
>
> There are other users of cppc_set_perf (drivers/cpufreq/amd-pstate.c) that need to be fixed up in the name change. Also, I think the name change should be a standalone patch.
>
> Is there a reason why you feel it necessary to remove energy_perf from the caps structure? I would just leave it for now as it seems it may not have been implemented correctly (comparing caps and ctrls) and could be fixed someday. Or if you want to remove it, it should be in a standalone cleanup patch.
>
> The logic in cppc_get_perf_ctrls looks correct. This should be added in a standalone patch.

In light of these comments - Splitting this into two commits.
First commit will just perform the rename, and the second will implement the logic in cppc_get_perf_ctrls.
I will drop the energy_perf change. It's not a high priority.

---

> https://github.com/NVIDIA/NV-Kernels/commit/d127e15e032be1f6f671684d0827ac7087dd604a NVIDIA: SAUCE: ACPI: CPPC: expand macro to create store acpi_cppc sysfs node

Updating the commit message to reflect that we are not using this new feature at this time.
I don't think there's other issues with this.

---

> https://github.com/NVIDIA/NV-Kernels/commit/0a9a49186040fa5ae0519cf1e49929e7c1bc7e82 NVIDIA: SAUCE: ACPI: CPPC: support updating epp, auto_sel and {min|max_perf} from sysfs

Taking commit message fixes, thank you.

> Why is store_energy_perf() using u64 for epp and input (and related code)? The energy_perf member is a u32, no?

Fixing, thank you.

> In cppc_set_epp_perf(), in the new !CPC_IN_PCC leg, is it correct to only write auto_sel_reg and epp_set_reg when both are supported? Or should they be checked/written independently?

Yes, both need to be supported. Otherwise, epp doesn't really play a role, since you also need auto_sel to be featured.
epp would be be regulating the balance between energy-savings & performance. This is documented in the next patch when we add the sysfs node docs.
I see that this doesn't align with the existing code, where we assume these two can be written separately in PCC.
Can this alignment be deferred until the formal upstreaming discussion is done? We know from the ACPI documentation that you would need both, so I'm not sure why the original implementation would separate these two out.

> In cppc_get_auto_sel_caps(), don’t you need to check CPC_SUPPORTED(auto_sel_reg) in the CPC_IN_PCC leg? Or maybe keep the check that was removed and don’t add the check in the !CPC_IN_PCC case?

Ack, good callout. Adding the CPC_SUPPORTED check into the if() logic.

> How are perf_limited and auto_activity_window used? I don’t see support for them.

They are RO knobs, so they will be handled by the existing `cppc_get_perf_fb_ctrs` and `cppc_get_perf_ctrls` functions.

---

> https://github.com/NVIDIA/NV-Kernels/commit/ac91475a5cdc12f5704707f1c779a073fc19ce06 NVIDIA: SAUCE: Documentation: ACPI: add autonomous mode ctrls info in cppc_sysfs.txt

Making the recommended fix, updating commit message to .rst and not .txt to reflect the actual file.

---

> https://github.com/NVIDIA/NV-Kernels/commit/8af0daab270a6d988beb01563676b3409ff04b30 NVIDIA: SAUCE: cpufreq: CPPC: Add cppc_cpufreq_epp instance for Autonomous mode

Fixing commit message, comment nits. Fixing return statement strangeness.

> I’m confused, cppc_auto_sel is added to kernel parameter documentation but then no support is added for the parameter? Is it supposed to be “auto_sel_mode”?

Yes, apologies. This was changed recently and we forgot to update documentation. Fixed.

> In cppc_cpufreq_epp_cpu_exit(), where is the complementary call site to cppc_cpufreq_put_cpu_data(policy)?

This is shared between epp & non epp implementation. This is also shared in the _init() functions.

> Why is cppc_cpufreq_auto_sel_enable() taking a boolean parameter? Is there a case when you would call it with false? Also, if it is to stay, please name it something other than the static global used for the module parameter.

The long-term I believe would be to allow for some init where we say "auto selection is possible, so setup the minimum HW as such, but do not init by default with auto selection enabled". This would support a future lifecycle where we can properly switch between HW and SW frequency control on the fly (and at boot).
As for naming, renamed slightly. Please comment again.

RE: clsotog's error, fixed. Good callout, we should fail better. I see how this would result in improper driver setup.

clsotog:
I checked the new commits and I do not see the USBAN and I see this print when failed to setup:
CPPC Cpufreq:Failed to enable cppc_cpufreq_epp_driver. Defaulting to cppc_cpufreq_driver.
I tried the echo 1 > auto_sel and no crash. 
Thanks at least those were my concerns. Unfortunately do not have a system that has this feature to try more experiments.

clsotog (approved):
Acked-by: Carol L Soto (csoto@nvidia.com)

Matt still reviewing, so if he has more comments I can review again.

nvmochs:
571fdaa06ab2 iommu/arm-smmu-v3: Implement IOMMU_HWPT_ALLOC_NEST_PARENT

This should not be a part of this PR, it is already in the branch.

6a942884a25d ACPI: CPPC: Fix MASK_VAL() usage

This is already present in the 6.11 HWE branch, no need to include again. I spoke with Carol and she thinks she was looking at the wrong origin branch. For these 2 CPPC fixes.

70b0541eb647 ACPI: CPPC: Fix _CPC register setting issue

Ditto, already present, no need to include.

< **To avoid confusion, please rebase without the above 3 patches** >

e7c07b337747 NVIDIA: SAUCE: ACPI: CPPC: add read perf ctrls api

Function comment nit: cppc_get_perf -> cppc_get_perf_ctrls

915b8a4242d2 NVIDIA: SAUCE: ACPI: CPPC: support updating epp, auto_sel and {min|max_perf} from sysfs

> I see that this doesn't align with the existing code, where we assume these two can be written separately in PCC.
> Can this alignment be deferred until the formal upstreaming discussion is done? We know from the ACPI documentation that you would need both, so I'm not sure why the original implementation would separate these two out.

Ok, I don’t have an issue with this as it’s existing.

> They are RO knobs, so they will be handled by the existing cppc_get_perf_fb_ctrs and cppc_get_perf_ctrls functions.

I’m sorry, I am not getting it. Where will they be handled by the existing routines? This patch is adding them to struct cpc_reg and  struct cppc_perf_fb_ctrs, yet I do not see where those members are populated.

I was expecting to see some code like:

```
perf_lim = &cpc_desc->cpc_regs[PERF_LIMITED];

…

perf_fb_ctrs->limted_perf = perf_lim;
```

01e09bf12d59 NVIDIA: SAUCE: cpufreq: CPPC: Add cppc_cpufreq_epp instance for Autonomous mode

Regarding cppc_cpufreq_enable_auto_sel_support(), my issue is that the parameter name is easy to confuse with the global variable of the same name. Maybe instead name the parameter “use_auto_sel_mode”?

nvidia-ishah:
> < To avoid confusion, please rebase without the above 3 patches >

Done.

> https://github.com/NVIDIA/NV-Kernels/commit/e7c07b337747c0a91bf8c67f4691935250e6a343 NVIDIA: SAUCE: ACPI: CPPC: add read perf ctrls api

Addressed function comment - good catch.

> https://github.com/NVIDIA/NV-Kernels/commit/915b8a4242d2d2cfbe865bc16cc715663904d661 NVIDIA: SAUCE: ACPI: CPPC: support updating epp, auto_sel and {min|max_perf} from sysfs

I see your point on the RO knobs. Confirming that this is a miss, I mistakenly thought this was already handled. Updating function to fix this.

> https://github.com/NVIDIA/NV-Kernels/commit/01e09bf12d591f02bd57d432be9a2e0921147b88 NVIDIA: SAUCE: cpufreq: CPPC: Add cppc_cpufreq_epp instance for Autonomous mode
> Regarding cppc_cpufreq_enable_auto_sel_support(), my issue is that the parameter name is easy to confuse with the global variable of the same name. Maybe instead name the parameter “use_auto_sel_mode”?

I see. Updated param name with your request.

> I checked the new commits and I do not see the USBAN and I see this print when failed to setup:
> CPPC Cpufreq:Failed to enable cppc_cpufreq_epp_driver. Defaulting to cppc_cpufreq_driver.
> I tried the echo 1 > auto_sel and no crash.
> Thanks at least those were my concerns. Unfortunately do not have a system that has this feature to try more experiments.

Thanks for testing this, I appreciate it. We did not catch this as we did not have a Grace available for testing.

clsotog:
Question I see the first commit "remove nvidia-fs from dkms-versions" and yesterday I thought maybe was a rebase issue.  Looking more to this,  it does not look like a rebase issue but then Im confused the name says remove nvidia-fs but the commit is removing mstflint. What is the intention of this commit?

nvidia-ishah:
> Question I see the first commit "remove nvidia-fs from dkms-versions" and yesterday I thought maybe was a rebase issue. Looking more to this, it does not look like a rebase issue but then Im confused the name says remove nvidia-fs but the commit is removing mstflint. What is the intention of this commit?

I think this is a rebase issue from changing the base branch of my working branch. This commit is not explicitly part of this particular merge-request.
I will remove this. If it is necessary for building working images, bfigg can open a merge request as such.

nvmochs (approved):
Thanks for addressing my concerns...no further feedback from me.

clsotog (approved):
Acked-by: Carol L Soto (csoto@nvidia.com)

nvmochs:
Pulled and present in https://git.launchpad.net/~canonical-kernel/ubuntu/+source/linux-nvidia/+git/noble/log/?h=nvidia-6.11-next

Closing PR.

PR #81, opened 2025-04-11
MEDIATEK: PCI: disable SBR on Gen5 ports

aaronp24:
Add quirk function to skip pcie secondary bus reset. PCIe link in occasionally(1/300 fail rate) drop to lower link width (ex. 4 lane drops to 2 lane or less) when secondary bus reset is applied. This happens because CDR lock is lost randomly due to mismatch between RC & EP speed (ex. RC in Gen1 and EP in Gen5).


nvmochs (approved):

jamieNguyenNVIDIA:
Where did this patch originate from?

If it hasn't been upstreamed, we normally preface the one-liner with "NVIDIA: SAUCE"
```
NVIDIA: SAUCE: MEDIATEK: PCI: disable SBR on Gen5 ports
```

PR #129, opened 2025-06-04
iommu/io-pgtable-arm: backport contiguous bit support

aritger:
iommu/io-pgtable-arm: Support contiguous bit in translation tables

    The contiguous bit in translation table entries can be used as a hint to
    SMMU that a group of adjacent translation table entries have consistent
    attributes and point to a contiguous and properly aligned output address
    range. This enables SMMU to predict the properties of the remaining
    translation table entries in the same group without accessing them. It
    also allows an SMMU implementation to make more efficient use of its TLB
    by using a single TLB entry to cover all translation table entries in
    the same group.

    In the case of 4KB granule size, there are 16 translation table entries
    in one group.

    This change sets the contiguous bit for such groups of entries that are
    completely covered by a single call to map_pages. As it stands, the code
    wouldn't set the contiguous bit if a group of adjacent descriptors is
    completed by separate calls to map_pages.

    Signed-off-by: Daniel Mentz <danielmentz@google.com>

from https://lore.kernel.org/linux-iommu/20250430231924.1481493-1-danielmentz@google.com/

For the backport: updated the __arm_lpae_init_pte() callsite in arm_lpae_split_blk_unmap() (arm_lpae_split_blk_unmap() is no longer present at top of tree where the original commit was written).

### DIFF
--- a/drivers/iommu/io-pgtable-arm.c
+++ b/drivers/iommu/io-pgtable-arm.c
@@ -76,6 +76,7 @@
 
 #define ARM_LPAE_PTE_NSTABLE		(((arm_lpae_iopte)1) << 63)
 #define ARM_LPAE_PTE_XN			(((arm_lpae_iopte)3) << 53)
+#define ARM_LPAE_PTE_CONT		(((arm_lpae_iopte)1) << 52)
 #define ARM_LPAE_PTE_DBM		(((arm_lpae_iopte)1) << 51)
 #define ARM_LPAE_PTE_AF			(((arm_lpae_iopte)1) << 10)
 #define ARM_LPAE_PTE_SH_NS		(((arm_lpae_iopte)0) << 8)
@@ -298,6 +299,27 @@ static void __arm_lpae_sync_pte(arm_lpae_iopte *ptep, int num_entries,
 				   sizeof(*ptep) * num_entries, DMA_TO_DEVICE);
 }
 
+static int arm_lpae_cont_ptes(int lvl, struct arm_lpae_io_pgtable *data)
+{
+	switch (ARM_LPAE_GRANULE(data)) {
+	case SZ_4K:
+		if (lvl >= 1)
+			return 16;
+		break;
+	case SZ_16K:
+		if (lvl == 2)
+			return 32;
+		else if (lvl == 3)
+			return 128;
+		break;
+	case SZ_64K:
+		if (lvl >= 2)
+			return 32;
+		break;
+	}
+	return 1;
+}
+
 static void __arm_lpae_clear_pte(arm_lpae_iopte *ptep, struct io_pgtable_cfg *cfg)
 {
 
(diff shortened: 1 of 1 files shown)


clsotog:
@aritger couple questions:
Is there a nvbug related to this commit? 
Your version of this patch will be submitted upstream? 
We usually add the key word NVIDIA: SAUCE: to the commit to identify patches that are not upstream.

aritger:
Thanks for looking at this.
 
The nvidia-internal bug is 5268190.  Should that be cited in the commit message?  I wasn't sure.
 
This patch was initially proposed upstream, but rejected in favor of a more general solution that is still under development.  So, I guess the "NVIDIA:SAUCE:" tag should be prepended to the commit message.

I've updated the commit message with the sauce prefix.  I'm new to this github workflow, so please let me know if I haven't followed the right conventions.

aritger:
Pushed again, hopefully with an even better commit message.

clsotog:
Thanks for sharing the nvbug. It helps. 
Do we know if Jason series will get approved soon or not? 
This patch, do you guys tested with 4k or also with 64K? Not sure what is the symptom if we do not have this patch from nvbug. Slower allocations with 4k?

aritger:
>  Do we know if Jason series will get approved soon or not?

Jason's series:
https://lore.kernel.org/linux-iommu/0-v2-5c26bde5c22d+58b-iommu_pt_jgg@nvidia.com/
is not enough to address the SMMU contig bit for ARM, by itself.  But as I understand it, that series would provide a better infrastructure on which to build SMMU contig_bit support.

So, I don't know, yet, when there would be an upstream replacement for the patch in this pull request.  It is possible we would also need this patch in the downstream 6.14-based kernel, if that is the reason for the question.

> This patch, do you guys tested with 4k or also with 64K?

Yes, it has been tested with 4k and 64k.  I also developed a selftest patch to exercise this patch with a variety of page sizes.
(the selftest patch is attached to bug 5268190; I didn't include that patch in the pull request, because the BaseOS kernel does not have the selftest framework enabled in the kconfig)

> Not sure what is the symptom if we do not have this patch from nvbug. Slower allocations with 4k?

The motivation is performance: without this patch, we can only get good TLB cache capacity on n1x with a 64kb pagesize kernel.  With this patch, we can get the same TLB cache capacity as 64k pagesize, but with 4kb pagesize in the kernel.

With poor TLB cache capacity, GPU workloads that access large amounts of memory will be slower.

It is desirable to use 4kb pagesize in the kernel, if possible, because:

* Changing from 4kb to 64kb shortly before product launch would be a risky change.
* 4kb has better compatibility with other software in the ecosystem.

clsotog:
Thanks so much for the answers.

clsotog (approved):

jamieNguyenNVIDIA:

jamieNguyenNVIDIA (approved):

jamieNguyenNVIDIA:
This has been pulled: https://git.launchpad.net/~canonical-kernel/ubuntu/+source/linux-nvidia/+git/noble/commit/?h=nvidia-6.11-next&id=1eae00a9aec7383cb9b8cad4eeef0d1329884b99.  Closing.

PR #126, opened 2025-06-04
Config acs and pasid 24.04 linux nvidia 6.11 next

tdavenvidia:
Backport Upstream patches to linux-nvidia 6.11 for GPUDirect

1) config_acs fix patch - https://lore.kernel.org/r/20250207030338.456887-1-tdave@nvidia.com
2) iommu skip pasid patch-   https://lore.kernel.org/all/20250520011937.3230557-1-tdave@nvidia.com/

PR #130, opened 2025-06-05
iommu: Skip PASID validation for devices without PASID capability

tdavenvidia:
Generally PASID support requires ACS settings that usually create single device groups, but there are some niche cases where we can get multi-device groups and still have working PASID support. The primary issue is that PCI switches are not required to treat PASID tagged TLPs specially so appropriate ACS settings are required to route all TLPs to the host bridge if PASID is going to work properly.

pci_enable_pasid() does check that each device that will use PASID has the proper ACS settings to achieve this routing.

However, no-PASID devices can be combined with PASID capable devices within the same topology using non-uniform ACS settings. In this case the no-PASID devices may not have strict route to host ACS flags and end up being grouped with the PASID devices.

This configuration fails to allow use of the PASID within the iommu core code which wrongly checks if the no-PASID device supports PASID.

Fix this by ignoring no-PASID devices during the PASID validation. They will never issue a PASID TLP anyhow so they can be ignored.

Fixes: c404f55c26fc ("iommu: Validate the PASID in iommu_attach_device_pasid()") Cc: stable@vger.kernel.org
(backported from commit b3f6fcd8404f9f92262303369bb877ec5d188a81)

Reviewed-by: Lu Baolu <baolu.lu@linux.intel.com>
Reviewed-by: Vasant Hegde <vasant.hegde@amd.com>
Link: https://lore.kernel.org/r/20250520011937.3230557-1-tdave@nvidia.com

### DIFF
--- a/drivers/iommu/iommu.c
+++ b/drivers/iommu/iommu.c
@@ -3353,9 +3353,11 @@ static int __iommu_set_group_pasid(struct iommu_domain *domain,
 	int ret;
 
 	for_each_group_device(group, device) {
-		ret = domain->ops->set_dev_pasid(domain, device->dev, pasid);
-		if (ret)
-			goto err_revert;
+		if (device->dev->iommu->max_pasids > 0) {
+			ret = domain->ops->set_dev_pasid(domain, device->dev, pasid);
+			if (ret)
+				goto err_revert;
+		}
 	}
 
 	return 0;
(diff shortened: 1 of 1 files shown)


kobak2026:
Acked-by: Koba Ko [kobak@nvidia.com](mailto:kobak@nvidia.com)

clsotog:
@tdavenvidia 
why I do not see the commit PCI/ACS: Fix 'pci=config_acs=' parameter in this pull request. I will expect to be similar to PR (https://github.com/NVIDIA/NV-Kernels/pull/126)

clsotog:
> @tdavenvidia why I do not see the commit PCI/ACS: Fix 'pci=config_acs=' parameter in this pull request. I will expect to be similar to PR (#126)

Discard this message I see it at the kernel tree.

clsotog:

jamieNguyenNVIDIA:
Pulled into `24.04_linux-nvidia-adv-6.11-next`.  Closing PR.

PR #139, opened 2025-06-11
Build kernel on each PR to make sure as sanity test

nirmoy:
Building the kernel before merging should be a bare minimal we should have for a repo. I haven't played with github action
before but this seems to work in my [local PR](https://github.com/nirmoy/NV-Kernels/pull/5)

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -1,0 +1,45 @@
+name: Kernel Build
+
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


nirmoy:
converting it to draft for more input and I would also need a .config file which is specific for our usecase.

jamieNguyenNVIDIA:

clsotog:

nirmoy:
We don't want to touch sync repos so for  now going to target  only internal and adv branches https://github.com/NVIDIA/NV-Kernels/pull/146

PR #147, opened 2025-07-02
Build kernel on each PR to make sure as sanity test

nirmoy:
Building the kernel before merging should be a bare minimal sanity test we should have for a repo. This uses github actions to build kernel for x86 and arm64 on github provided machines. In the future this can be extedned to run kselftests.

Demo: https://github.com/nirmoy/NV-Kernels/pull/5

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -1,0 +1,45 @@
+name: Kernel Build
+
+on:
+  pull_request:
+    branches: [ "*" ]
+
+jobs:
+  build:
+    name: ${{ matrix.arch == 'amd64' && 'x86_64' || 'ARM64' }} - ${{ matrix.flavour }}
+    runs-on: ${{ matrix.runs-on }}
+    strategy:
+      matrix:
+        include:
+          - runs-on: ubuntu-latest
+            arch: amd64
+            flavour: nvidia
+          - runs-on: ubuntu-24.04-arm
+            arch: arm64
+            flavour: nvidia
+          - runs-on: ubuntu-24.04-arm
+            arch: arm64
+            flavour: nvidia-64k
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Install Dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y build-essential libncurses-dev bison flex libssl-dev libelf-dev python3
+
+      - name: Configure Kernel
+        run: |
+          debian/scripts/misc/annotations --arch ${{ matrix.arch }} --flavour ${{ matrix.flavour }} --export > .config
+          make oldconfig
+          # Ensure WERROR is disabled
+          ./scripts/config --disable CONFIG_WERROR
+          # Disable certificate verification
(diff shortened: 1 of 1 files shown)


nvidia-bfigg:
I know how this is going to sound but I'm honestly not in favor of this. We've been doing development the way we have without this and we've not had a single build failure that I can think of. Before implementing this I want to see actual data that a problem is being solved. 

Developers that are submitting patches are doing their own santiy builds and testing. The patches we are applying are getting a lot of review and more test builds. Having github perform more automated builds and tests just isn't adding real value.

nirmoy:
This patch will immediately free us from repetitively building the kernel for each PR, which I think we must do if we aren't already. This alone gives it enough value to be considered.

Developers build and test on the latest kernel. We are backporting here, so it's not exactly the same. Maybe I'm not aware of existing CI we have that's doing more tests. Developers also might not use our Kconfig, which is very specific as far as I understand.

I'm not surprised that there weren't many build failures (IIRC, Matt told me there was a build failure discovered with a different Kconfig for Ubuntu), which is expected given our workflow of applying reviewed and tested patches. However, that's not a reason not to add an extra layer of checking, as we're backporting patches and using a different Kconfig, which might not tested by upstream developers.

nvidia-bfigg:
No, every developer must be building kernels with their patches before submitting PRs, THIS is good practice. There's no excuse for not building test kernels with your PR.

I'm not buying your other arguments that it may not build with other Kconfig either, if you wanted to test with that we'd need builds that tested with lots of different Kconfigs which isn't practical.

Ubuntu is our development distribution, that's what people should be building and testing with, if they aren't then they need to change. I'm not going to be convinced otherwise.

nvmochs:
> Matt told me there was a build failure discovered with a different Kconfig for Ubuntu

It wasn't necessarily a different Kconfig, but rather issues with the annotations:

- Developer encountered a kernel build failure due to the annotations check and decided to "handle it later" by using the skip option to bypass the check. Then the developer forgot to address the error and submitted the PR. The annotations issue was not found until either a reviewer built the kernel or [worse] after the code has been merged and a build is attempted.
- Developer only attempted a build for arm64 and x86 encounters an annotations failure.

Both of these scenarios have occurred more than once. I agree that if a developer does everything correctly up front then this is not an issue. But the CI would help catch the cases I've listed here consistently and earlier in the PR/merge process.

nirmoy:
> No, every developer must be building kernels with their patches before submitting PRs, THIS is good practice. There's no excuse for not building test kernels with your PR.

I'm referring to the person who is merging a pull request: they should be building the kernel first, and only then merge the PR. Sorry for the confusion.

> 
> I'm not buying your other arguments that it may not build with other Kconfig either, if you wanted to test with that we'd need builds that tested with lots of different Kconfigs which isn't practical.
True, we can't do everything but should atleast try to test with arm-64k, arm-4k, x86-def config.

nvidia-bfigg:
OK, for tech preview kernels, I have no problem with adding this CI process improvement. You have my ACK.

clsotog (commented):

nvmochs (approved):

nvmochs:
Merged, closing PR.

PR #148, opened 2025-07-02
vfio/nvgrace-gpu: Avoid resmem pfn unregistration (6.11-adv)

nvmochs:
When porting the ECC/EGM series, to this branch, some logic that avoids the “resmem” region for systems that do not use it (i.e. Grace-Blackwell) was dropped. This was due to the Grace-Blackwell support implementation changing during the upstreaming process. The 6.8 branches have the pre-upstream version, and therefore their ECC/EGM patches did not take into account some of the changes that went into the upstream version.

This patch adds a check to only unregister the resmem pfnmap when it was created. It also adds a sanity check in the mmap handler to call out a potential bug (trying to register resmem pfnmap on systems where it is not required).

Tested on GH and GB by passing through a GPU to a VM:
- While running a [bpftrace](https://drive.google.com/file/d/1MZVqWtwINPG-9WBSMX7MpHc1nzMkQ-oe/view?usp=drive_link) on the host to evaluate pfnmap registrations on boot/shutdown
- Evaluating the kernel log after VM shutdown to ensure the print mentioned in the commit is no longer present.

### DIFF
--- a/drivers/vfio/pci/nvgrace-gpu/main.c
+++ b/drivers/vfio/pci/nvgrace-gpu/main.c
@@ -239,7 +239,8 @@ static void nvgrace_gpu_close_device(struct vfio_device *core_vdev)
 	mutex_destroy(&nvdev->remap_lock);
 
 #ifdef CONFIG_MEMORY_FAILURE
-	unregister_pfn_address_space(&nvdev->resmem.pfn_address_space);
+	if (nvdev->resmem.memlength)
+		unregister_pfn_address_space(&nvdev->resmem.pfn_address_space);
 	unregister_pfn_address_space(&nvdev->usemem.pfn_address_space);
 #endif
 	vfio_pci_core_close_device(core_vdev);
@@ -320,9 +321,10 @@ static int nvgrace_gpu_mmap(struct vfio_device *core_vdev,
 #ifdef CONFIG_MEMORY_FAILURE
 	vma->vm_ops = &nvgrace_gpu_vfio_pci_mmap_ops;
 
-	if (index == VFIO_PCI_BAR2_REGION_INDEX)
+	if (index == VFIO_PCI_BAR2_REGION_INDEX) {
+		WARN_ON_ONCE(!nvdev->has_mig_hw_bug);
 		ret = nvgrace_gpu_vfio_pci_register_pfn_range(&nvdev->resmem, vma);
-	else
+	} else
 		ret = nvgrace_gpu_vfio_pci_register_pfn_range(&nvdev->usemem, vma);
 #endif
 


clsotog (approved):

drivers/vfio/pci/nvgrace-gpu/main.c
```diff
@@ -239,7 +239,8 @@ static void nvgrace_gpu_close_device(struct vfio_device *core_vdev)
 	mutex_destroy(&nvdev->remap_lock);
 
 #ifdef CONFIG_MEMORY_FAILURE
-	unregister_pfn_address_space(&nvdev->resmem.pfn_address_space);
+	if (nvdev->resmem.memlength)
+		unregister_pfn_address_space(&nvdev->resmem.pfn_address_space);
```
nirmoy, line 243: unregister_pfn_address_space() has a check if the address space is NULL then it will return without any error. 
So it means there was a successful call  to nvgrace_gpu_vfio_pci_register_pfn_range(). Is this double call of 
the unregister_pfn_address_space() ?

nirmoy: I thought `resmem` probably wasn't zeroed at init but that is not the case AFAIU reading `vfio_alloc_device` which is using zero alloc function.

nvmochs: In this case pfn_address_space is a structure embedded within "resmem", so it will never be non-NULL. So the parameter pointer check inside unregister_pfn_address_space is essentially a no-op from this call site.

Instead, we need to use some other means to determine the validity of resmem. The original implementation in the 6.8 branch used the "has_mig_hw_bug" flag, as that is what was used to configure resmem. Since resmem and usemem are both unconditionally initialized, I think it's more accurate to use "memlength" to indicate if resmem was in use. There are other places in the code where that field is used in a similar manner.

nirmoy: I see the issue : to get `resource: Trying to free nonexistent resource <0x0000000000000000-0x000000000000ffff>`  this must be running on 64k page_size machine where:  `pfn_space->node.last == pfn_space->node.start = 0`
`void unregister_pfn_address_space(struct pfn_address_space *pfn_space) -->
        release_mem_region(pfn_space->node.start << PAGE_SHIFT,
                           (pfn_space->node.last - pfn_space->node.start + 1) << PAGE_SHIFT);`

nirmoy (approved):

drivers/vfio/pci/nvgrace-gpu/main.c
nvmochs: Exactly, thanks for the ack.

nirmoy: I missed up code formatting which I can't seem to make it work.

nvmochs:
Merged, closing PR.

PR #149, opened 2025-07-02
vfio/nvgrace-gpu: Avoid resmem pfn unregistration (6.11-internal)

nvmochs:
When porting the ECC/EGM series, to this branch, some logic that avoids the “resmem” region for systems that do not use it (i.e. Grace-Blackwell) was dropped. This was due to the Grace-Blackwell support implementation changing during the upstreaming process. The 6.8 branches have the pre-upstream version, and therefore their ECC/EGM patches did not take into account some of the changes that went into the upstream version.

This patch adds a check to only unregister the resmem pfnmap when it was created. It also adds a sanity check in the mmap handler to call out a potential bug (trying to register resmem pfnmap on systems where it is not required).

Tested on GH and GB by passing through a GPU to a VM:
- While running a [bpftrace](https://drive.google.com/file/d/1MZVqWtwINPG-9WBSMX7MpHc1nzMkQ-oe/view?usp=drive_link) on the host to evaluate pfnmap registrations on boot/shutdown
- Evaluating the kernel log after VM shutdown to ensure the print mentioned in the commit is no longer present.

### DIFF
--- a/drivers/vfio/pci/nvgrace-gpu/main.c
+++ b/drivers/vfio/pci/nvgrace-gpu/main.c
@@ -239,7 +239,8 @@ static void nvgrace_gpu_close_device(struct vfio_device *core_vdev)
 	mutex_destroy(&nvdev->remap_lock);
 
 #ifdef CONFIG_MEMORY_FAILURE
-	unregister_pfn_address_space(&nvdev->resmem.pfn_address_space);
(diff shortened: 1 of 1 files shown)


clsotog (approved):

nirmoy (approved):

nvmochs:
Merged, closing PR.

PR #163, opened 2025-07-14
NVIDIA: SAUCE: efi: Export screen_info

nmahale:
The NVIDIA driver requires access to the screen_info data to restore the console, so you need to export the screen_info data.

### DIFF
--- a/drivers/firmware/efi/efi-init.c
+++ b/drivers/firmware/efi/efi-init.c
@@ -62,7 +62,7 @@ extern __weak const efi_config_table_type_t efi_arch_tables[];
  */
 #if !defined(CONFIG_X86) && (defined(CONFIG_SYSFB) || defined(CONFIG_EFI_EARLYCON))
 struct screen_info screen_info __section(".data");
-EXPORT_SYMBOL_GPL(screen_info);
+EXPORT_SYMBOL(screen_info);
 #endif
 
 static void __init init_screen_info(void)


clsotog:
@nmahale:
Is this for digits or also for GB200/GB300? 
Is there a nvbug for reference?
Can you add the NVIDIA: SAUCE: tag if this is not upstream?
Thanks

jamieNguyenNVIDIA:
@nmahale: is this something that can go upstream?

nvmochs:
@nmahale - Can you provide more information on why this is required? Is this related to a bug?

nmahale:
> Is this for digits or also for GB200/GB300?
It is required for digit to support console restore with proprietary driver modules.

>  Is there a nvbug for reference?
5106852

> Can you add the NVIDIA: SAUCE: tag if this is not upstream?
This is not for upstream.

nirmoy:
I don't think we can do this and distribute such kernel to end user without breaking GPL law. 
Ref: https://lwn.net/Articles/154603

arighi:
I agree with @nirmoy, also why not using the DRM / fbdev APIs to resume the console? This would be a much cleaner and more maintainable approach.

nmahale:
It’s okay if we can’t pull this patch. It simply means that console restore will only be supported with OpenRM, which is the default. We’re working to enable nvidia-drm fbdev, but progress is slow and we’re encountering multiple issues along the way.

nvidia-bfigg:
Then please consider this PR as NAK'd.

jamieNguyenNVIDIA:
Closing based on comments above.

PR #187, opened 2025-08-12
[RFC][linux-nvidia-6.14] Add nvidia-fs as an in tree module

ianm-nv:
This PR adds the gds-nvidia-fs source as an in tree kernel module.  Since we've had issues with handling nvidia-fs-dkms as a separate package, including as a module feels like it is worthy of a discussion and investigation to see if it would provide value and limit some error cases we've previously encountered. 

The high level design of the PR is to create a top level nvidia directory with the nvidia-fs src directory as a child.  It adds a top level Kconfig for nvidia drivers with nvidia-fs as a sub config.  The various nvidia-fs config options are also added as Kconfig options.  Having the top level nvidia directory could also be used to store additional nvidia modules that aren't upstream.

Additional details are in the [LP BugLink](https://bugs.launchpad.net/bugs/2119679)

For a prebuilt test, there is an unsigned kernel build in this ppa
 add-apt-repository ppa:ian-may/nvidia-fs-kernel-test-1
 apt install linux-headers-6.14.0-1008-nvidia
 apt install linux-image-unsigned-6.14.0-1008-nvidia
 apt install linux-modules-extra-6.14.0-1008-nvidia
 apt install nvidia-dkms-570-server-open
 reboot
 modinfo nvidia-fs


nvidia/nvidia-fs/Makefile
```diff
@@ -0,0 +1,53 @@
...
+nvidia-fs-$(CONFIG_FAULT_INJECTION) += nvfs-fault.o
+
+# Main module target
+obj-$(CONFIG_NVIDIA_FS) += nvidia-fs.o
+
+# Compiler flags
+ccflags-y += -Wall
+
+# NVIDIA driver headers path (required for nv-p2p.h)
+NVIDIA_VERSION := $(shell ls -1d /usr/src/nvidia-* 2>/dev/null | sort -V | tail -1 | xargs basename 2>/dev/null | sed 's/nvidia-//')
```
jamieNguyenNVIDIA, line 22: Ouch -- does this mean that the NVIDIA GPU driver kernel source will need to be present to correctly build linux-modules-$(uname -r)?

ianm-nv: Yes, definitely not ideal. Technically this dependency is already resolved in the debian sources build environment. In order to generate the 'linux-modules-nvidia-fs' package the 'debian.nvidia-6.14/control.stub.in' contains the following lines in the Build-Depends section
` nvidia-dkms-kernel [amd64 arm64] <!stage1>,
 nvidia-kernel-source [amd64 arm64] <!stage1>,`
 
But it does potentially add an additional step of fetching the GPU driver source for users wanting to build on their own with a mechanism like 'make modules' or 'debian/rules binary'

jamieNguyenNVIDIA: Does the `nvidia-fs` build not care which `nvidia-dkms-kernel` or `nvidia-kernel-source` you have installed?   i.e. R570 vs. R580?

ianm-nv: Correct, looking through the [gds-nvidia-fs](https://github.com/NVIDIA/gds-nvidia-fs) source I was unable to find any indication of specific nvidia driver version checks.
[https://github.com/NVIDIA/gds-nvidia-fs/blob/master/src/Makefile](https://github.com/NVIDIA/gds-nvidia-fs/blob/master/src/Makefile#L23)
[https://github.com/NVIDIA/gds-nvidia-fs/blob/master/src/create_nv.symvers.sh](https://github.com/NVIDIA/gds-nvidia-fs/blob/master/src/create_nv.symvers.sh)

So from what I can tell, this lack of version locking exists today even with `nvidia-fs-dkms`
Here are the imported symbols defined in [nvfs-p2p.h](https://github.com/NVIDIA/gds-nvidia-fs/blob/master/src/nvfs-p2p.h)
`nvidia_p2p_get_pages_persistent
nvidia_p2p_put_pages_persistent
nvidia_p2p_get_pages
nvidia_p2p_put_pages
nvidia_p2p_dma_map_pages
nvidia_p2p_dma_unmap_pages
nvidia_p2p_free_page_table
nvidia_p2p_free_dma_mapping`

jamieNguyenNVIDIA: That behavior seems more palatable for DKMS users since they'd be building nvidia-fs.ko against whatever GPU driver they have installed on the system at the time.

But in our case, I'm worried about this type of pattern:
1. We build `linux-modules-$(uname -r)` against GPU driver X
2. The user is running GPU driver Y
3. The `nvidia_p2p_*` symbols change between GPU driver X vs GPU driver Y

Wouldn't this lead to symbols disagreements when trying to load our in-tree version?

ianm-nv: Agree with everything you stated. I think a similar version problem possibly exists today if a user installs the `nvidia-driver-dkms` and the `linux-modules-nvidia-fs` package.

For example doing
`sudo apt install linux-nvidia`

$ sudo apt show `linux-nvidia` | grep Depends
Depends: `linux-image-nvidia` (= 6.8.0-1035.38), linux-headers-nvidia (= 6.8.0-1035.38)

$ sudo apt show `linux-image-nvidia` | grep Depends
Depends: linux-image-6.8.0-1035-nvidia, linux-modules-extra-6.8.0-1035-nvidia, linux-firmware, intel-microcode, amd64-microcode, `linux-modules-nvidia-fs-6.8.0-1035-nvidia`

Using the `linux-nvidia` meta would install the prebuilt nvidia-fs, but I don't think we have a guarantee of which nvidia driver is installed.

I'm thinking we would want a depends on the compiled against nvidia driver version here
$ sudo apt show `linux-modules-nvidia-fs-6.8.0-1035-nvidia` | grep Depends
Depends: linux-image-6.8.0-1035-nvidia | linux-image-unsigned-6.8.0-1035-nvidia

jamieNguyenNVIDIA: Based on this discussion, my opinion is that it would be better to leave nvidia-fs out of tree.  What do you think, Ian?

ianm-nv: @jamieNguyenNVIDIA I think it was worth investigating and discussing, but ultimately I agree that the external dependency will eventually cause a problem for us and we are better off keeping it out of tree.

ianm-nv:
Per discussions, we've decided not to merge this PR.

PR #212, opened 2025-09-25
drm/ast: Use msleep instead of mdelay for edid read

nirmoy:
The busy-waiting in mdelay() can cause CPU stalls and kernel timeouts during boot.

https://bugs.launchpad.net/ubuntu/+source/linux-nvidia-6.14/+bug/2125667
https://nvbugspro.nvidia.com/bug/5541552


nvmochs (approved):

nirmoy:
6.11 is used only by Spark which doesn't have this AST. So closing it.. Thanks @clsotog for noticing it

PR #238, opened 2025-11-11
iommu: Skip PASID validation for devices without PASID capability

tdavenvidia:
Generally PASID support requires ACS settings that usually create single device groups, but there are some niche cases where we can get multi-device groups and still have working PASID support. The primary issue is that PCI switches are not required to treat PASID tagged TLPs specially so appropriate ACS settings are required to route all TLPs to the host bridge if PASID is going to work properly.

pci_enable_pasid() does check that each device that will use PASID has the proper ACS settings to achieve this routing.

However, no-PASID devices can be combined with PASID capable devices within the same topology using non-uniform ACS settings. In this case the no-PASID devices may not have strict route to host ACS flags and end up being grouped with the PASID devices.

This configuration fails to allow use of the PASID within the iommu core code which wrongly checks if the no-PASID device supports PASID.

Fix this by ignoring no-PASID devices during the PASID validation. They will never issue a PASID TLP anyhow so they can be ignored.

Fixes: c404f55c26fc ("iommu: Validate the PASID in iommu_attach_device_pasid()") Cc: stable@vger.kernel.org
(backported from commit b3f6fcd8404f9f92262303369bb877ec5d188a81)

Reviewed-by: Lu Baolu <baolu.lu@linux.intel.com>
Reviewed-by: Vasant Hegde <vasant.hegde@amd.com>
Link: https://lore.kernel.org/r/20250520011937.3230557-1-tdave@nvidia.com

### DIFF
--- a/drivers/iommu/iommu.c
+++ b/drivers/iommu/iommu.c
@@ -3341,9 +3341,11 @@ static int __iommu_set_group_pasid(struct iommu_domain *domain,
 	int ret;
 
 	for_each_group_device(group, device) {
-		ret = domain->ops->set_dev_pasid(domain, device->dev, pasid);
-		if (ret)
-			goto err_revert;
+		if (device->dev->iommu->max_pasids > 0) {
+			ret = domain->ops->set_dev_pasid(domain, device->dev, pasid);
+			if (ret)
+				goto err_revert;
+		}
 	}
 
 	return 0;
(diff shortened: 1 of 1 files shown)


khfeng (changes requested):
The content doesn't exactly match the upstream commit, so it should use "backported from" instead of "cherry picked from".

tdavenvidia:
> The content doesn't exactly match the upstream commit, so it should use "backported from" instead of "cherry picked from".

Done. Updated commit text to "backported from".

kobak2026:
LGTM
Reviewed-by: kobak@nvidia.com

nvmochs (approved):

nvmochs:
Merged, closing PR.

arighi (commented):
I usually like when there's a short line describing what changed from the original commit, unless it's just context adjustments, in this case we could add something like this:
```
[v6.12: updated set_dev_pasid() API, old domain argument added in v6.13]
```

But apart than that LGTM.

PR #248, opened 2025-11-19
[linux-nvidia-6.14-next] iommufd: Avoid including VM_PFNMAP memory from RLIMIT_MEMLOCK

NathanChenNVIDIA:
Device memory regions such as I/O device BARs do not consume host RAM and should not count against the RLIMIT_MEMLOCK limit for how much memory can be locked/pinned.

When a VM with device passthrough is launched by an unprivileged process (e.g., Libvirt launches qemu process as an unprivileged user), IOMMUFD's unified memory accounting causes ENOMEM errors for large device BARs that exceed the process memlock limit. VFIO without IOMMUFD does not check device memory against the limit and therefore does not show the same behavior.

This commit sets device memory account_mode to IOPT_PAGES_ACCOUNT_NONE, bypassing the RLIMIT_MEMLOCK limit to restore the behavior of VFIO without IOMMUFD for device memory.

LP: https://bugs.launchpad.net/ubuntu/+source/linux-nvidia/+bug/2131970

Testing:
Launch a GPU passthrough Libvirt-qemu VM with iommufd backing and observe ENOMEM from GPU BAR mapping operations. Subsequent VM launches fail. Add CAP_IPC_LOCK capability to the qemu binary and observe successful BAR mappings + multiple VM launch.

Upstream plans:
Working on different implementation for upstream memory accounting infrastructure that differs from what we carry here.

### DIFF
--- a/drivers/iommu/iommufd/pages.c
+++ b/drivers/iommu/iommufd/pages.c
@@ -1410,6 +1410,18 @@ struct iopt_pages *iopt_alloc_user_pages(void __user *uptr,
 		return pages;
 	pages->uptr = uptr_down;
 	pages->type = IOPT_ADDRESS_USER;
+
+	if (pages->account_mode == IOPT_PAGES_ACCOUNT_USER) {
+		struct vm_area_struct *vma;
+
+		mmap_read_lock(pages->source_mm);
+		vma = find_vma(pages->source_mm, (unsigned long)uptr_down);
+		if (vma && (vma->vm_flags & VM_PFNMAP)) {
+			pages->account_mode = IOPT_PAGES_ACCOUNT_NONE;
+		}
+		mmap_read_unlock(pages->source_mm);
+	}
+	
 	return pages;
 }
 


clsotog:
we will need this for 6.17?

NathanChenNVIDIA:
> we will need this for 6.17?

Yes, 6.17 PR here: https://github.com/NVIDIA/NV-Kernels/pull/249

clsotog (approved):

nvmochs:
@NathanChenNVIDIA This seems like it is allowing for a privilege escalation and could open an attack vector. Why isn't the admin-controlled WAR (adding CAP_IPC_LOCK capability to the qemu binary) sufficient until a proper upstream solution can be developed?

NathanChenNVIDIA:
> @NathanChenNVIDIA This seems like it is allowing for a privilege escalation and could open an attack vector. Why isn't the admin-controlled WAR (adding CAP_IPC_LOCK capability to the qemu binary) sufficient until a proper upstream solution can be developed?

These changes shouldn't allow for a privilege escalation - this commit exempts device memory from being subject to the memory pinning limit, but that device memory already requires privileged access to map in the first place (controlled by access to VFIO device file permissions). Setting the CAP_IPC_LOCK capability is less secure because it allows the qemu process to bypass the memory pinning limit entirely for all memory types including guest RAM. Following this commit's approach only applies the memory pinning limit exemption to VM_PFNMAP VMAs that are not meant to count against the limit.

NathanChenNVIDIA:
After an offline discussion we have decided to close this PR and pursue a long term fix upstream, as we are unblocked by setting the CAP_IPC_LOCK capability on the qemu binary.

NathanChenNVIDIA:
After an offline discussion we have decided to close this PR and pursue a long term fix upstream, as we are unblocked by setting the CAP_IPC_LOCK capability on the qemu binary.

PR #253, opened 2025-12-01
NVMe/NVMEeoF: Patch NVMe/NVMeoF driver to support GDS on Linux 6.17 Kernel

sourabgupta3:
With this change, the NVMe and NVMeOF driver would be enabled to support GPUDirectStorage(GDS). NVMe driver introduced a way to use the blk_rq_dma_map API to DMA map requests instead of scatter gather lists. With these changes, GDS path also adopts a similar framework where we introduce blk based APIs(nvfs_blk_rq_dma_map_iter_start and nvfs_blk_rq_dma_map_iter_next) to map a DMA request. The NVMeoF path remains the same as previous releases.

### DIFF
--- a/drivers/nvme/host/Makefile
+++ b/drivers/nvme/host/Makefile
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: GPL-2.0
 
 ccflags-y				+= -I$(src)
-
+ccflags-y				+= -DCONFIG_NVFS
 obj-$(CONFIG_NVME_CORE)			+= nvme-core.o
 obj-$(CONFIG_BLK_DEV_NVME)		+= nvme.o
 obj-$(CONFIG_NVME_FABRICS)		+= nvme-fabrics.o
@@ -20,10 +20,11 @@ nvme-core-$(CONFIG_NVME_HWMON)		+= hwmon.o
 nvme-core-$(CONFIG_NVME_HOST_AUTH)	+= auth.o
 
 nvme-y					+= pci.o
-
+nvme-y					+= nvfs-dma.o
 nvme-fabrics-y				+= fabrics.o
 
 nvme-rdma-y				+= rdma.o
+nvme-rdma-y				+= nvfs-rdma.o
 
 nvme-fc-y				+= fc.o
 
--- a/drivers/nvme/host/nvfs-dma.c
+++ b/drivers/nvme/host/nvfs-dma.c
@@ -1,0 +1,51 @@
+/* SPDX-License-Identifier: GPL-2.0 */
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms and conditions of the GNU General Public License,
+ * version 2, as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+ * more details.
+ */
+#ifdef CONFIG_NVFS
+#define NVFS_USE_DMA_ITER_API
+#define MODULE_PREFIX nvme_v2
+#include "nvfs.h"
+
+struct nvfs_dma_rw_blk_iter_ops *nvfs_ops = NULL;
+
+atomic_t nvfs_shutdown = ATOMIC_INIT(1);
+
+DEFINE_PER_CPU(long, nvfs_n_ops);
+
+#define NVIDIA_FS_COMPAT_FT(ops) \
+	(NVIDIA_FS_CHECK_FT_BLK_DMA_MAP_ITER_START(ops) && NVIDIA_FS_CHECK_FT_BLK_DMA_MAP_ITER_NEXT(ops))
+
+// protected via nvfs_module_mutex
+int REGISTER_FUNC(struct nvfs_dma_rw_blk_iter_ops *ops)
+{
+       if (NVIDIA_FS_COMPAT_FT(ops)) {
+             nvfs_ops = ops;
+             atomic_set(&nvfs_shutdown, 0);
+             return 0;
+       } else
+             return -EOPNOTSUPP;
+
+}
+EXPORT_SYMBOL_GPL(REGISTER_FUNC);
+
+// protected via nvfs_module_mutex
+void UNREGISTER_FUNC(void)
+{
+        (void) atomic_cmpxchg(&nvfs_shutdown, 0, 1);
+        do {
+                msleep(NVFS_HOLD_TIME_MS);
+        } while(nvfs_count_ops());
+        nvfs_ops = NULL;
+}
+EXPORT_SYMBOL_GPL(UNREGISTER_FUNC);
+#endif
--- a/drivers/nvme/host/nvfs-dma.h
+++ b/drivers/nvme/host/nvfs-dma.h
@@ -1,0 +1,180 @@
+/* SPDX-License-Identifier: GPL-2.0 */
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms and conditions of the GNU General Public License,
+ * version 2, as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+ * more details.
+ */
+
+#ifndef NVFS_DMA_H
+#define NVFS_DMA_H
+
+/* Forward declarations for functions from pci.c that we need */
+static blk_status_t nvme_pci_setup_data_prp(struct request *req,
+		struct blk_dma_iter *iter);
+static blk_status_t nvme_pci_setup_data_sgl(struct request *req,
+		struct blk_dma_iter *iter);
+static inline struct dma_pool *nvme_dma_pool(struct nvme_queue *nvmeq,
+		struct nvme_iod *iod);
+static inline dma_addr_t nvme_pci_first_desc_dma_addr(struct nvme_command *cmd);
+
+static inline bool nvme_nvfs_unmap_sgls(struct request *req)
+{
+	struct nvme_iod *iod = blk_mq_rq_to_pdu(req);
+	struct nvme_queue *nvmeq = req->mq_hctx->driver_data;
+	struct device *dma_dev = nvmeq->dev->dev;
+	dma_addr_t sqe_dma_addr = le64_to_cpu(iod->cmd.common.dptr.sgl.addr);
+	unsigned int sqe_dma_len = le32_to_cpu(iod->cmd.common.dptr.sgl.length);
+	struct nvme_sgl_desc *sg_list = iod->descriptors[0];
+	enum dma_data_direction dir = rq_dma_dir(req);
+        
+	if (iod->nr_descriptors) {
+                unsigned int nr_entries = sqe_dma_len / sizeof(*sg_list), i;
+
+                for (i = 0; i < nr_entries; i++) {
+			nvfs_ops->nvfs_dma_unmap_page(dma_dev, 
+					              iod->nvfs_cookie, 
+						      le64_to_cpu(sg_list[i].addr), 
+						      le32_to_cpu(sg_list[i].length), 
+						      dir);
+		}
+        } else
+		nvfs_ops->nvfs_dma_unmap_page(dma_dev, iod->nvfs_cookie, sqe_dma_addr, sqe_dma_len, dir);
+        
+	
+	
+	return true;
+}
+
+static inline bool nvme_nvfs_unmap_prps(struct request *req)
+{
+	struct nvme_iod *iod = blk_mq_rq_to_pdu(req);
+	struct nvme_queue *nvmeq = req->mq_hctx->driver_data;
+	struct device *dma_dev = nvmeq->dev->dev;
+	enum dma_data_direction dma_dir = rq_dma_dir(req);
+	unsigned int i;
+
+	/* Check if dma_vecs was allocated - if setup failed early, it might be NULL */
+	if (!iod->dma_vecs)
+		return true;
+
+	/* Unmap all DMA vectors - pass page pointer from dma_vecs */
+	for (i = 0; i < iod->nr_dma_vecs; i++) {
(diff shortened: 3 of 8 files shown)


nvmochs:
@sourabgupta3 Couple of comments/questions after looking at this PR:

drivers/nvme/host/Makefile
- Nit: Please use tabs instead of spaces

<hr>

drivers/nvme/host/nvfs-dma.c
- The copyright data differs from this same file in linux-nvidia-6.14
- In linux-nvidia-6.14 REGISTER_FUNC returns -EOPNOTSUPP but here it is returning -ENOTSUPP. Is that intended? 

<hr>

drivers/nvme/host/nvfs-dma.h
- Same comment about copyright data.


<hr>
drivers/nvme/host/nvfs.h
- Same comment about the copyright data.

<hr>

drivers/nvme/host/pci.c
- Are the forward declarations for nvme_pci_sgl_set_data_sg and nvme_pci_sgl_set_data_seg needed?
- In nvme_pci_prp_iter_next(), the page is extracted into bvec, but then not used?
- In nvme_map_data(), is it the intended behavior to not check the return code from nvme_nvfs_map_data?

drivers/nvme/host/nvfs-rdma.h
```diff
@@ -0,0 +1,89 @@
...
+				DMA_ATTR_NO_WARN);
+
+		if (unlikely((*count == NVFS_IO_ERR))) {
+			nvfs_put_ops();
+			return -EIO;
+		}
+
+		if (unlikely(*count == NVFS_CPU_REQ)) {
+			nvfs_put_ops();
+			BUG();
```
nirmoy: This BUG() too agressive here. This needs a proper comment. Also in this case and the previous one return is -EIO but BUG() is only used here which is inconsistent.  `BUG()` is generally discouraged as it will hang the system.

drivers/nvme/host/nvfs-dma.h
```diff
@@ -0,0 +1,182 @@
...
+	/* Check if this was an NVFS I/O by checking the IOD_NVFS_IO flag */
+	if (!(iod->flags & IOD_NVFS_IO))
+		return false;
+
+	/* Clear the NVFS flag */
+	iod->flags &= ~IOD_NVFS_IO;
+
+	if (!nvfs_ops || !nvfs_ops->nvfs_dma_unmap_page) {
+		nvfs_put_ops();
+		return false;
```
nirmoy: `nvfs_put_ops()` is confusing here. Please move it to the caller of nvme_nvfs_unmap_data() so the flow of ref counting is more readable.

drivers/nvme/host/pci.c
```diff
@@ -711,12 +723,28 @@ static void nvme_free_sgls(struct request *req)
...
+
 static void nvme_unmap_data(struct request *req)
 {
 	struct nvme_iod *iod = blk_mq_rq_to_pdu(req);
 	struct nvme_queue *nvmeq = req->mq_hctx->driver_data;
 	struct device *dma_dev = nvmeq->dev->dev;
 
+#ifdef CONFIG_NVFS
+	/* Check if this was an NVFS I/O and handle unmapping */
+	if (nvme_nvfs_unmap_data(req))
```
nirmoy, line 738: There are lots of usages of this `ifdef CONFIG_NVFS`. How about moving those to the header file.  Something like: 
```
#ifndef NVFS_DMA_H
#define NVFS_DMA_H

#ifdef CONFIG_NVFS
static inline bool nvme_nvfs_unmap_data(struct request *req)
{
	struct nvme_iod *iod = blk_mq_rq_to_pdu(req);


.....
}
#else /* !CONFIG_NVFS */
static inline bool nvme_nvfs_unmap_data(struct request *req)
{
	return false;
}

```

nirmoy (commented):
Also make sure to run `./scripts/checkpatch.pl <patch file>`

drivers/nvme/host/nvfs-dma.c
```diff
@@ -0,0 +1,43 @@
...
+
+atomic_t nvfs_shutdown = ATOMIC_INIT(1);
+
+DEFINE_PER_CPU(long, nvfs_n_ops);
+
+#define NVIDIA_FS_COMPAT_FT(ops) \
+	(NVIDIA_FS_CHECK_FT_BLK_DMA_MAP_ITER_START(ops) && NVIDIA_FS_CHECK_FT_BLK_DMA_MAP_ITER_NEXT(ops))
+
+// protected via nvfs_module_mutex
+int REGISTER_FUNC (struct nvfs_dma_rw_blk_iter_ops *ops)
```
jamieNguyenNVIDIA: There's an extra space character in front of the `(`.

drivers/nvme/host/nvfs-dma.c
```diff
@@ -0,0 +1,43 @@
...
+             atomic_set(&nvfs_shutdown, 0);
+             return 0;
+       } else
+             return -ENOTSUPP;
+
+}
+EXPORT_SYMBOL_GPL(REGISTER_FUNC);
+
+// protected via nvfs_module_mutex
+void UNREGISTER_FUNC (void)
```
jamieNguyenNVIDIA: There's an extra space character in front of the `(`.

drivers/nvme/host/nvfs-dma.c
```diff
@@ -0,0 +1,43 @@
...
+{
+        (void) atomic_cmpxchg(&nvfs_shutdown, 0, 1);
+        do{
```
jamieNguyenNVIDIA: Missing space character before `{`.

drivers/nvme/host/nvfs-dma.h
```diff
@@ -0,0 +1,182 @@
...
+                unsigned int nr_entries = sqe_dma_len / sizeof(*sg_list), i;
+
+                for (i = 0; i < nr_entries; i++) {
+			nvfs_ops->nvfs_dma_unmap_page(dma_dev, 
+					              iod->nvfs_cookie, 
+						      le64_to_cpu(sg_list[i].addr), 
+						      le32_to_cpu(sg_list[i].length), 
+						      dir);
+		}
+        } else {
```
jamieNguyenNVIDIA: Looks like you can get rid of the `{` and `}` here.

drivers/nvme/host/pci.c
jamieNguyenNVIDIA: Similarly, it seems a couple of the `#include`'s can be moved to the top to reduce these:

1. `#include "nvfs-dma.h"` in `pci.c`
2. `#include "nvfs-rdma.h"` in `rdma.c`

drivers/nvme/host/nvfs-rdma.h
sourabgupta3: Fixed this one.

drivers/nvme/host/nvfs-dma.h
sourabgupta3: Realized that this wasn't needed, so removed it.

drivers/nvme/host/pci.c
sourabgupta3: This is the similar pattern we have been using for the prior versions and that's why keeping it similar. But are you asking to move it to nvfs-dma.h and always calling nvme_nvfs_unmap_data()? That way, won't we always need nvme_nvfs_unmap_data() to be defined even if we are not compiling with nvfs.

nvmochs:
@sourabgupta3 Thanks for address these comments. A few more items...

> drivers/nvme/host/pci.c

> * In nvme_map_data(), is it the intended behavior to not check the return code from nvme_nvfs_map_data?

Can you respond to this? Is the idea here that if is_nvfs_io is false then we should continue on and try the non-NVFS path?


<hr> 

`/* Store DMA mapping AND page pointer (extracted before iterator moved) */`

Is this comment still accurate?

<hr>

Once the review cycle is complete, can you squash so that there is just a single commit?

drivers/nvme/host/pci.c
jamieNguyenNVIDIA: In the example Nirmoy gave, `nvme_nvfs_unmap_data()` would always be defined.  But when `CONFIG_NVFS` is not set, then this function would return without doing anything.

sourabgupta3: Agreed but if we define these in nvfs-dma.h, then we would always need to include nvfs-dma.h as well.

drivers/nvme/host/nvfs-dma.c
sourabgupta3: Fixed

drivers/nvme/host/nvfs-dma.c
sourabgupta3: Fixed

drivers/nvme/host/nvfs-dma.h
sourabgupta3: Fixed

drivers/nvme/host/nvfs-dma.c
sourabgupta3: Fixed

drivers/nvme/host/Makefile
```diff
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: GPL-2.0
 
 ccflags-y				+= -I$(src)
-
+ccflags-y                               += -DCONFIG_NVFS
```
khfeng: Is there any reason to change this line?

drivers/nvme/host/nvfs-dma.c
khfeng: Snake case is preferred.

drivers/nvme/host/nvfs-rdma.c
```diff
@@ -0,0 +1,52 @@
...
+{
+       if (NVIDIA_FS_COMPAT_FT(ops)) {
+             nvfs_ops = ops;
+             atomic_set(&nvfs_shutdown, 0);
+             return 0;
+       } else
+             return -EOPNOTSUPP;
+
+}
+EXPORT_SYMBOL_GPL(REGISTER_FUNC);
```
khfeng, line 40: Duplicated exported symbol name.

khfeng (commented):
The second commit needs to be squashed.

The subject "NVMe/MVMEeOF" is a typo.

drivers/nvme/host/Makefile
sourabgupta3: There was a comment on this PR to use tabs instead on spaces, that's why. If you look at the current changes, it looks in sync with rest of the file.

drivers/nvme/host/nvfs-dma.c
sourabgupta3: Historically, we have been using this name for the last 6-7 years, don't want to change that now.

drivers/nvme/host/nvfs-rdma.c
sourabgupta3: I don't think this is a duplicate, REGISTER_FUNC is a macro and it would resolve to different name for NVMe and NVMeoF driver.

drivers/nvme/host/pci.c
sourabgupta3: @jamieNguyenNVIDIA We have these inclusions at these places because, there are some static methods in pci.c that nvfs-dma.h/nvfs-rdma.h relies on and including the file at this place ensures those are defined prior to compiling content in nvfs-dma.h

drivers/nvme/host/Makefile
khfeng: My bad. I thought `ccflags-y				+= -I$(src)` was removed. Nevertheless, it should be `obj-$(CONFIG_NVFS) += nvfs_dma.o nvfs_rdma.o`.

drivers/nvme/host/nvfs-rdma.c
khfeng: It's not a macro, and both nvfs-dma.c and nvfs-rdma.c export REGISTER_FUNC.

sourabgupta3: #define REGSTR2(x) x##_register_nvfs_dma_ops
#define REGSTR(x)  REGSTR2(x)

#define UNREGSTR2(x) x##_unregister_nvfs_dma_ops
#define UNREGSTR(x)  UNREGSTR2(x)

#define REGISTER_FUNC REGSTR(MODULE_PREFIX)
#define UNREGISTER_FUNC UNREGSTR(MODULE_PREFIX)
This is how it's defined and MODULE_PREFIX is different for nvfs-dma.c and nvfs-rdma.c and that would make the names different for these files.
#define MODULE_PREFIX nvme_v2. --->>> nvfs-dma.c
#define MODULE_PREFIX nvme_rdma_v1 --->nvfs_rdma.c

nvmochs (approved):
With the two commits squashed into a single commit:

nirmoy:
One more thing, we also need to hide this behind annotations. Something like https://github.com/NVIDIA/NV-Kernels/commit/ae9e27cdd9582782abdea6684b61b591c43f1563

jamieNguyenNVIDIA:
@sourabgupta3 : Can you please take care of this typo in your commit message?  `MVMEeoF`

clsotog (approved):

jamieNguyenNVIDIA:
@sourabgupta3: I still see some issues in the commit message:
```
commit 276ef4a60479e212ff39f26b5509e7960fdce6a8 (HEAD -> sg_24.04_linux-nvidia-6.17-next, sourab/sg_24.04_linux-nvidia-6.17-next)
Author: Sourab Gupta <sougupta@nvidia.com>
Date:   Thu Nov 20 18:10:03 2025 +0000

    NVIDIA: SAUCE: NVMe/MVMEeoF: Patch NVMe/NVMeoF driver to support GDS on Linux 6.17 Kernel
    
    With this change, the NVMe and NVMeoF driver would be enabled to support GPUDirectStorage(GDS).
    NVMe driver introduced a way to use the blk_rq_dma_map API to DMA map requests instead of
    scatter gather lists. With these changes, GDS path also adopts a similar framework where we
    introduce blk based APIs(nvfs_blk_rq_dma_map_iter_start and nvfs_blk_rq_dma_map_iter_next)
    to map a DMA request. The NVMeoF path remains the same as previous releases.
    
    Signed-off-by: Sourab Gupta <sougupta@nvidia.com>
    Reviewed-by: Kiran Modukuri <kmodukuri@nvidia.com>
```

Please remove the extra "NVMe/MVMEeoF: Patch" from the title so that it just says, "NVIDIA: SAUCE: NVMe/NVMeoF driver to support GDS on Linux 6.17 Kernel".  Also please limit the lines in the body to 75 characters

jamieNguyenNVIDIA:
Thanks, Sourab.

jamieNguyenNVIDIA:
LP: https://bugs.launchpad.net/ubuntu/+source/linux-nvidia-6.17/+bug/2134960

nvmochs:
Merged in https://git.launchpad.net/~canonical-kernel/ubuntu/+source/linux-nvidia/+git/noble/log/?h=nvidia-6.17-next

Closing PR.

PR #259, opened 2025-12-02
NVIDIA: SAUCE: Downgrade "xhci_endpoint_init rsv" message level

khfeng:
The message should use pr_debug() instead of pr_err().


nvmochs:
@khfeng - Is this for a particular nvbug? Also, 6.11 is dead.

nirmoy:
Can you also send this to upstream?

clsotog:
This printk is not upstream. This print was introduced by a patch from brad did at beginning for 6.11 kernel. That's why he is added as reviewer. 
There is no specific nvbug but I see these printks a lot from the GB300WS dmesgs so I agree for make it pr_debug.

nirmoy:
> This printk is not upstream. This print was introduced by a patch from brad did at beginning for 6.11 kernel. That's why he is added as reviewer. There is no specific nvbug but I see these printks a lot from the GB300WS dmesgs so I agree for make it pr_debug.

Ok. Thanks for that info. Yes the patch is fine.

clsotog:
Closing we finally got confirmation to revert the patch that introduced this pr_err. 
Will create the PRs to revert the patch. 
Thanks @khfeng for awareness for this issue.

PR #319, opened 2026-02-12
NVIDIA: VR: SAUCE: Introduce full.config and full-64k.config

jamieNguyenNVIDIA:
Introduce full.config and full-64k.config files:

- full.config: Full config (nvidia-6.17 + generic-6.18 new options, env excluded). Deduplicated so options already in nvidia.config are omitted. Merge with nvidia.config to build.

- full-64k.config: 64K page size variant of full.config. Same dedup rule. 64K-related options taken from nvidia-6.17-64k reference (page size, PGTABLE_LEVELS, VA/PA bits, ARCH_MMAP_RND, etc.).


jamieNguyenNVIDIA:
The full configs were generated using the following formula:

- config-6.17.0-1008-nvidia from noble + config-6.18.0-9-generic from resolute
- stripped out environment-based configs
- deduplicated configs that were already in nvidia.config
- set a few additional ones to keep ensure the compilation was prompt-less

arighi (approved):
LGTM.

To verify this I tried to re-generate the configs using noble:linux-nvidia-6.17 as a baseline + the configs from the latest resolute:linux kernel from Ubuntu for the new options and I produced similar results.

A general comment (not specific to this PR): it would be nice to have a reference (nvbug / LP bug / generic URL) for each NVIDIA option defined in `arch/arm64/configs/nvidia.config`. Some configs already include an LP bug (which is great), while others only have a brief comment. Sometimes having a proper reference makes much easier to understand the full context and rationale behind enabling (or disabling) a particular option. Maybe something to keep in mind for the future configs.

nvmochs:
@jamieNguyenNVIDIA 

Previously (with just nvidia.confg) we were saying to merge defconfig + nvidia.confg. Because of that, I think we concluded that we did not need to add "CONFIG_NVIDIA_TEGRA410_CMEM_LATENCY_PMU=m" to nvidia.config. However since we'd now be recommending to merge the full*.config + nvidia.config, I think we need to sure that any defconfig items that we want are captured in either full*.config or nvidia.config. i.e. we need to add CONFIG_NVIDIA_TEGRA410_CMEM_LATENCY_PMU=m to nvidia.config. I looked at others but this was the one that stood out.

I assume we'd need to potentially refresh full*.config when stable updates are applied? We might want to capture that methodology in the commit message.

jamieNguyenNVIDIA:
> @jamieNguyenNVIDIA
> 
> Previously (with just nvidia.confg) we were saying to merge defconfig + nvidia.confg. Because of that, I think we concluded that we did not need to add "CONFIG_NVIDIA_TEGRA410_CMEM_LATENCY_PMU=m" to nvidia.config. However since we'd now be recommending to merge the full*.config + nvidia.config, I think we need to sure that any defconfig items that we want are captured in either full*.config or nvidia.config. i.e. we need to add CONFIG_NVIDIA_TEGRA410_CMEM_LATENCY_PMU=m to nvidia.config. I looked at others but this was the one that stood out.
> 
> I assume we'd need to potentially refresh full*.config when stable updates are applied? We might want to capture that methodology in the commit message.

Good points.

I'll add `CONFIG_NVIDIA_TEGRA410_CMEM_LATENCY_PMU` to nvidia.config.  I believe this is the only one missing.

And instead of hand-crafting the full.config/full-64k.config, I'll check in a script that can regenerate these files when given the right configs (i.e. `config-6.17.0-1008-nvidia` + `config-6.18.0-9-generic` or `config-6.17.0-1008-nvidia-64k` + `config-6.18.0-9-generic-64k`)

nvmochs (commented):
Thanks Jamie for the script - that will help with future updates!

I confirmed that using the script and example provided in the commit message generates the full configs that are added by this commit.

nvmochs (approved):

scripts/kconfig/merge_full_configs.py
```diff
@@ -0,0 +1,208 @@
...
+    for key in base_order:
+        merged[key] = None  # placeholder to preserve order
+
+    for cfg_path in config_paths:
+        d = full_config_to_dict(cfg_path)
+        for key, line in d.items():
+            if is_env_key(key):
+                continue
+            merged[key] = line
+        # Add keys only in this config (not in base order)
```
arighi: Is this block really needed? We just did `merged[key] = line` so in this block key will be always in merged. I guess we can remove this block entirely.

scripts/kconfig/merge_full_configs.py
```diff
@@ -0,0 +1,208 @@
...
+
+def parse_full_config(path: Path):
+    """Parse full .config: CONFIG_X=val or # CONFIG_X is not set. Returns ordered (key, line)."""
+    entries = []
+    for line in open(path, encoding="utf-8", errors="replace"):
+        s = line.rstrip("\r\n")
+        if not s.strip():
+            entries.append((None, s))
+            continue
+        m = re.match(r'^# CONFIG_([A-Za-z0-9_]+) is not set\s*$', s.strip())
```
arighi: Minor improvement (feel free to ignore): instead of compiling the regex at every line, we could pre-compile them at the beginning as:
```
RE_NOTSET = re.compile(r'^# CONFIG_([A-Za-z0-9_]+) is not set\s*$')
RE_SET = re.compile(r'^(CONFIG_([A-Za-z0-9_]+)=(.*))$')
```

And then use `RE_NOTSET.match()` and `RE_SET.match()`. But this is just speed, not correctness.

arighi (approved):
Left a couple of comments, but overall LGTM. 

jamieNguyenNVIDIA:
Rebased and addressed Andrea's comments.

scripts/kconfig/merge_full_configs.py
jamieNguyenNVIDIA: Hi @arighi,

I believe you're right.  This should be addressed in the latest iteration.

scripts/kconfig/merge_full_configs.py
jamieNguyenNVIDIA: Good suggestion.  I believe I've addressed this in the latest iteration.

nvmochs:

arighi (approved):
LGTM
```
```

About `merge_full_configs.py` I asked AI to review it (they all look minor things). Here's the report:

## Review: `scripts/kconfig/merge_full_configs.py`

Overall the logic looks correct — merge order (base + overlays), dedup against fragment, and env exclusion all make sense. A few notes:

---

### 🐛 File handle leak

Lines 48 and 75 use `open()` without closing:
```python
for line in open(path, encoding="utf-8", errors="replace"):
```

Should use a context manager:
```python
with open(path, encoding="utf-8", errors="replace") as f:
    for line in f:
```

Not critical for a short-lived script, but bad practice and will trigger linter warnings.

---

### 📝 Minor: Redundant path resolution

Lines 157-158 re-resolve `dedup_path` inside the output section:
```python
dedup_path = args.dedup if args.dedup.is_absolute() else repo / args.dedup
```

But it was already resolved at line 126. Could just reuse that variable.

---

### ✅ Looks good otherwise

The placeholder pattern (`merged[key] = None`) is guarded correctly before output, so no issue there. Logic is sound.

jamieNguyenNVIDIA:
PR merged: https://github.com/NVIDIA/NV-Kernels/commit/f2c220d67255f235e05825578b562fd827c15b61

Thanks all!

PR #329, opened 2026-02-24
NVIDIA: VR: SAUCE: [Config] Update configs for virtualization support

ianm-nv:
Add kernel configuration options required for libvirt/virt-install networking and virtualization device support.

In kernel 6.18, CONFIG_IP_NF_IPTABLES was split into a nft-based variant and IP_NF_IPTABLES_LEGACY. Legacy iptables modules (IP_NF_FILTER, IP_NF_NAT, IP_NF_TARGET_MASQUERADE, etc.) now depend on the LEGACY variant. Without it, these modules are silently dropped by Kconfig, breaking libvirt's default iptables firewall backend.

Enable both firewall backends:
- nftables: NF_TABLES, NFT_NAT, NFT_MASQ, NFT_CT, NFT_REJECT with IPv4/IPv6 family support
- legacy iptables: NETFILTER_XTABLES_LEGACY, IP_NF_IPTABLES_LEGACY, IP6_NF_IPTABLES_LEGACY

Also add:
- Traffic control: NET_SCH_HTB, NET_CLS_U32, NET_ACT_CSUM for libvirt bridge QoS and DHCP filtering
- VHOST_NET for virtio-net host backend
- VIRTIO_IOMMU for guest DMA isolation
- VIRT_DRIVERS to enable the virtualization drivers subsystem

Without these configs enabled I'm hitting network errors when attempting to use virt-install.
"ERROR    Requested operation is not valid: network 'default' is not active"


ianm-nv:
These configs are addressed in https://github.com/NVIDIA/NV-Kernels/pull/319
closing PR.

PR #298, opened 2026-01-28
[linux-nvidia-6.17] Add ARM CCA host support

ianm-nv:
[BugLink](https://bugs.launchpad.net/ubuntu/+source/linux-nvidia-6.17/+bug/2139249)

[ Impact ]

ARM Confidential Compute Architecture (CCA) provides hardware-enforced isolation for confidential virtual machines called "Realms" on ARM64 platforms. This patch series enables CCA support for NVIDIA Vera platforms.

This series is based on the ARM KVM RME host support patches (v10), rebased for the 6.17 kernel:
https://lore.kernel.org/linux-coco/20250820145606.180644-1-steven.price@arm.com/

This series enables:
  -KVM host support for creating and managing Realms via the Realm Management Extension (RME)
  -MECID (Memory Encryption Context ID) for improved isolation between Realms
  -Required CCA kernel configuration options

[ Test Plan ]

Deploy and test on NVIDIA Vera platform with RMM firmware
Verify Realm guest VMs boot and run successfully
CCA testing requires specialized hardware and firmware. Testing performed by NVIDIA CCA team.

[ Where problems could occur ]

Bugs in the KVM/RME integration could cause Realm guest failures or host instability. Issues would be limited to CCA-enabled platforms running Realm workloads.

[ Other Info ]

Patch summary:
43 patches for upstream v10 KVM/RME host support - marked as SAUCE because not in upstream kernel yet.
3 upstream cherry-picks:
  arm64: realm: ioremap: Allow mapping memory as encrypted
  arm64: acpi: Enable ACPI CCEL support
  arm64: Enable EFI secret area Securityfs support
4 SAUCE patches:
  NVIDIA: VR: SAUCE: KVM: arm64: Expose KVM_CAP_ARM_RME via module parameter
  NVIDIA: VR: SAUCE: arm64: RME: Add MECID support
  NVIDIA: VR: SAUCE: arm64: RME: Add bounds check
  NVIDIA: VR: SAUCE: [Config] Update ARM CCA annotations

PR #332, opened 2026-03-03
[linux-nvidia-6.18] Add ARM CCA host support

ianm-nv:
This PR adds host support for ARM CCA to the linux-nvidia-6.18 kernel.

The bulk of these patches (42 of 47) are derived from ARM's CCA v11 patch series on the "cca-host/v11" branch of https://gitlab.arm.com/linux-arm/linux-cca. ARM announced this series on LKML [1] but did not submit the patches inline. Five commits required minor adjustments during cherry-pick and are marked as backported with explanations.

In addition to ARM's Host patches there are four additional commits:
arm64: RME: Add MECID support -- MECID allocation for realm guests
arm64: RME: Add bounds check -- bounds validation for RME operations
KVM: arm64: Expose KVM_CAP_ARM_RMI via module parameter
arm64: rsi: Move RSI init to fix Realm guest boot on BBML2
config: Enable EFI secret area for CCA realm guests

BBML2 boot fix
Realm guests running a 6.18+ kernel on BBML2-capable hardware fail to boot. The root cause is a timing issue: arm64_rsi_init() ran after paging_init(), so is_realm_world() was always false when the kernel linear map was created. On BBML2 hardware, this caused block mappings to be used instead of page-level mappings, which cannot be split later when set_memory_decrypted() is called for SWIOTLB/ITS memory. The fix moves RSI initialization earlier in setup_arch() and reorders the checks in force_pte_mapping() so that the realm check is not bypassed by the BBML2 early-return path.

KVM_CAP_SYNC_MMU allowlist
ARM's v10 series used a blocklist model for realm KVM capabilities. The v11 refactor switched to an allowlist (kvm_realm_ext_allowed()), but KVM_CAP_SYNC_MMU was accidentally omitted. This capability is required for guest_memfd private memory allocation, so it has been added to the allowlist in the backported "Don't expose unsupported capabilities for realm guests" commit.

[1] https://lore.kernel.org/linux-coco/461fa23f-9add-40e5-a0d0-759030e7c70b@arm.com/


jamieNguyenNVIDIA:
@ianm-nv do you need to port [these config changes](https://github.com/NVIDIA/NV-Kernels/pull/298/changes/d0f7f41e249ce9faf3f55795cf3165a02d8cc459) to this branch as well?

arch/arm64/include/uapi/asm/kvm.h
```diff
@@ -459,6 +461,17 @@ struct arm_rmi_config {
 			__u32	hash_algo;
 		};
 
+		/* cfg == ARM_RME_CONFIG_MEC */
```
jamieNguyenNVIDIA: A couple nitpicks on this comment and the one on line 470: the comments say `ARM_RME_CONFIG_MEC` and `ARM_RME_CONFIG_MEC_QUERY`, but the defines are `ARM_RMI_CONFIG_MEC` and `ARM_RMI_CONFIG_MEC_QUERY`.

jamieNguyenNVIDIA (commented):
Some of the pure SAUCE patches seem like they might be worth sending to the ARM CCA tree:
```
arm64: RME: Add MECID support -- MECID allocation for realm guests
arm64: RME: Add bounds check -- bounds validation for RME operations
arm64: rsi: Move RSI init to fix Realm guest boot on BBML2
```

Are there any plans to do that?

ianm-nv:
> @ianm-nv do you need to port [these config changes](https://github.com/NVIDIA/NV-Kernels/pull/298/changes/d0f7f41e249ce9faf3f55795cf3165a02d8cc459) to this branch as well?

Thanks!  In 6.18 the PRIVATE_MEM configs were replaced with CONFIG_KVM_GUEST_MEMFD and along with CONFIG_KVM_GENERIC_MEMORY_ATTRIBUTES is now auto selected when CONFIG_KVM=y.  

We still do still want `CONFIG_EFI_SECRET=y` and I've included an additional patch to add that to nvidia.config

arch/arm64/include/uapi/asm/kvm.h
ianm-nv: Fixed and pushed, thanks!

ianm-nv:
> Some of the pure SAUCE patches seem like they might be worth sending to the ARM CCA tree:
> 
> ```
> arm64: RME: Add MECID support -- MECID allocation for realm guests
> arm64: RME: Add bounds check -- bounds validation for RME operations
> arm64: rsi: Move RSI init to fix Realm guest boot on BBML2
> ```
> 
> Are there any plans to do that?

Yes, we notify ARM on bugs/issues we encounter.  These should all be upstream in the future.

jamieNguyenNVIDIA:
Thanks for the quick responses and fixes!

nvmochs:
@ianm-nv Couple of comments / questions...

2d2ebca0b180 NVIDIA: VR: CCA: SAUCE: arm64: RMI: Provide accurate register list

- I don’t follow the commit message comment about the "timer register list changes already present in 6.18” as I’m not seeing them.

<hr>

2fd30b0a9817 NVIDIA: VR: CCA: SAUCE: arm64: RMI: Allow VMM to set RIPAS

- This should have a backported tag and context note about the end bit assignment change in kvm_realm_unmap_range().
- Was this fix also reported upstream?

<hr>

dc37bd90fdaa NVIDIA: VR: CCA: SAUCE: KVM: arm64: Support timers in realm RECs
- I don't follow the commit message comment about dropping the realm guard because it is already present in 6.18, I’m not seeing the kvm_is_realm() in timer_set_offset() in linux-nvidia-6.18.

ianm-nv:
@nvmochs Thanks, I should've included a better explanation in my PR message.

> @ianm-nv Couple of comments / questions...
> 
> [2d2ebca](https://github.com/NVIDIA/NV-Kernels/commit/2d2ebca0b180005513ff87a3ba9e45fe127f63f4) NVIDIA: VR: CCA: SAUCE: arm64: RMI: Provide accurate register list
> 
> * I don’t follow the commit message comment about the "timer register list changes already present in 6.18” as I’m not seeing them.

In 6.18 the timer registers were consolidated.  The timer register code in guest.c was moved into `sys_reg_descs`, so instead of having to hide timer registers in two places we only need to hide it in the one place where `kvm_realm_sys_reg_hidden_user` is implemented.

Updated context adjustement:
`[ianm: drop timer register list changes from guest.c; 6.18 moved
     timer registers into sys_reg_descs where they are covered by
     kvm_realm_sys_reg_hidden_user()]`

> 
> [2fd30b0](https://github.com/NVIDIA/NV-Kernels/commit/2fd30b0a9817eef264285d07b527ef2351005281) NVIDIA: VR: CCA: SAUCE: arm64: RMI: Allow VMM to set RIPAS
> 
> * This should have a backported tag and context note about the end bit assignment change in kvm_realm_unmap_range().
> * Was this fix also reported upstream?

Good catch, I've updated to include the same backport tag used in 6.17.  This has been reported to the upstream author who acknowledged the bug and agreed with the adjustment.

> 
> [dc37bd9](https://github.com/NVIDIA/NV-Kernels/commit/dc37bd90fdaa671134cb0d4cba326b26ecd1728f) NVIDIA: VR: CCA: SAUCE: KVM: arm64: Support timers in realm RECs
> 
> * I don't follow the commit message comment about dropping the realm guard because it is already present in 6.18, I’m not seeing the kvm_is_realm() in timer_set_offset() in linux-nvidia-6.18.

Another fail at brevity, `timer_set_offset` was moved into a header file `include/kvm/arm_arch_timer.h` which means in order to call `kvm_is_realm` we need to have `arm_arch_timer.h` include `kvm_emulate.h`, but `kvm_emulate.h` already includes `arm_arch_timer.h`, so we find ourselves in a circular dependency.

I looked into adding a "wrapper" function in `arch_timer.c` that would call `kvm_is_realm` and including that in `arm_arch_timer.h` but that would add overhead to the inline timer_set_offset.  Better for readability, but worse for performance, and would affect all VMs.

Since the goal with this patch is give RMM control of timers, instead of trying to resolve this dependency chain, I decided it was best to leverage the existing vm_offset check in `timer_set_offset` and `timer_get_offset`.

Updated context adjustement:
`[ianm: drop timer_set_offset() realm guard; vm_offset is set to
     NULL for realms in timer_context_init(), handled by existing NULL
     check in 6.18's inline timer_set_offset()]`

nvmochs (approved):
@ianm-nv Thanks for clarifying and updating the commit notes!

No further issues from me.

jamieNguyenNVIDIA:
My ACK still stands:

nirmoy (approved):
LGTM

clsotog (approved):

jamieNguyenNVIDIA:
Content merged -- thanks all.

https://github.com/NVIDIA/NV-Kernels/commit/707df2508d09d9dfffae6a22656a9bb314aee0a1

PR #351, opened 2026-03-31
CSD lockup during kexec due to unbounded busy-wait in pl011_console_write_atomic 

nirmoy:
Backport  ```9bd18e1262c0 "printk/nbcon: Restore IRQ in atomic flush after each emitted record"``` to fix: 
https://lore.kernel.org/all/sqwajvt7utnt463tzxgwu2yctyn5m6bjwrslsnupfexeml6hkd@v6sqmpbu3vvu/

There are two more patches
```
ce2bba89566b "mm/kfence: add reboot notifier to disable KFENCE on shutdown"
9bc9ccbf4c93 "mm/kfence: fix potential deadlock in reboot notifier"
```

 which are of interest but we disable CONFIG_KFENCE_STATIC_KEYS so those fixes can be ignored.

LP: https://bugs.launchpad.net/ubuntu/+source/linux-nvidia-6.17/+bug/2146955
https://nvbugspro.nvidia.com/bug/5983492


nirmoy:
I should also add https://github.com/NVIDIA/NV-Kernels/commit/d5d399efff65773ed82ddaf6c11a0fcfdb5eb029

nirmoy:
We don't have kfence so closing

PR #356, opened 2026-04-13
Add patchscan workflow to check PRs for missing upstream fixes

nirmoy:
## Summary

Adds patchscan support to \`24.04_linux-nvidia-6.17-next\` to begin with —
more branches can be added to the trigger list as needed.

The workflow runs on every PR targeting \`24.04_linux-nvidia-6.17-next\` and
checks that cherry-picked commits are not missing upstream \`Fixes:\` patches.
It posts a comment on the PR listing any missing fixes and fails the check,
or posts an all-clear if none are found.

## How it works

The workflow lives entirely on the \`github-actions\` default branch — no
changes needed on target branches. On each PR it:

1. Checks out the PR head (full history)
2. Fetches \`origin/linux\` as the upstream reference
3. Runs \`patchscan <base>..<head> origin linux --no-update\`
4. Posts a warning comment listing any missing fix commits, or an all-clear otherwise

## Files added

| File | Description |
|------|-------------|
| \`.github/workflows/patchscan.yml\` | The workflow |
| \`.github/scripts/patchscan\` | The patchscan script (from @jmartin-lc) |
| \`.github/scripts/requirements.txt\` | Python deps (\`colorama\`, \`GitPython\`) |

## Test

Validated on a fork with two scenarios:

- **No missing fixes** — nirmoy/NV-Kernels#7: all-clear comment posted ✅
- **Missing fix detected** — nirmoy/NV-Kernels#8: backport of \`bd23f293a0d5\` without its fix \`83da212b7fca\` correctly flagged ✅

Co-developed-by: Claude AI

### DIFF
--- a/.github/scripts/patchscan
+++ b/.github/scripts/patchscan
@@ -1,0 +1,173 @@
+#!/usr/bin/python3
+# Author: Jacob Martin <jacob.martin@canonical.com>
+from argparse import ArgumentParser
+from colorama import Fore
+import git
+import os
+import re
+
+parser = ArgumentParser(
+        description="Checks that cherry-picked or backported commits actually exist upstream.")
+parser.add_argument("check_range", help="Git commit range to check, e.g. HEAD~10..HEAD")
+parser.add_argument("upstream_remote", help="Remote to consider as the upstream with potential fixes, e.g. torvalds")
+parser.add_argument("upstream_branch", help="Branch to consider as the upstream with potential fixes, e.g. master")
+parser.add_argument("--no-update", action='store_true')
+parser.add_argument("--no-fixes", action='store_true')
+parser.add_argument("--url", help="URL prefix to use for commit links", default="https://github.com/torvalds/linux/commit/")
+args = parser.parse_args()
+
+def format_commit_hex12_msg(commit):
+    firstline = commit.message.split('\n')[0]
+    return f"{commit.hexsha[0:12]} (\"{firstline}\")"
+
+def format_hyperlink(url, text):
+    return f"\33]8;;{url}\a{text}\33]8;;\a"
+
+def format_commit_with_hyperlink(commit, url_prefix):
+    return format_hyperlink(f"{url_prefix}{commit.hexsha}", format_commit_hex12_msg(commit))
+
+def get_src_commit_strs(msg):
+    # Match common backport/cherry-pick annotations:
+    #   (cherry picked from commit <sha>)
+    #   (backported from commit <sha>)
+    #   upstream commit <sha>
+    #   Upstream commit <sha>
+    matches = list(re.finditer(r"(?:from commit|[Uu]pstream commit) ([a-fA-F0-9]+)", msg))
+    return list(map(lambda m: m.group(1) if len(m.groups()) == 1 else None, matches))
+
+def references_upstream(commit):
+    return len(get_src_commit_strs(commit.message)) > 0
+
+def is_upstream(upstream, commit):
+    assert commit
+
+    repo = commit.repo
+
+    for commit_str in get_src_commit_strs(commit.message):
+        pick_title = commit.message.split('\n')[0]
+
+        try:
+            src_commit = repo.commit(commit_str)
+        except Exception as exc:
+            print("\nE: {} refers to invalid 'upstream' SHA {}: {}".format(pick_title, commit_str, exc))
+            return False
+
+        src_title = src_commit.message.split('\n')[0]
+        if src_title != pick_title:
+            print("\nE: Commit messages differ: upstream '{}' != pick '{}'".format(src_title, pick_title))
+            return False
+
+        if src_commit.author != commit.author:
+            print("\nE: Commit authors differ: upstream '{}' != pick '{}'".format(src_commit.author, commit.author))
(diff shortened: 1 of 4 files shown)


nvmochs:
@nirmoy This looks good.

Is this triggered on every push of the branch? Using your second scenario where a missing fix was detected, if you force push the branch again with the missing fix, does the scanner run again to verify that the fix is now present (and no other fixes are found).

Is there a way to trigger this manually? e.g. We can enter a PR number and it will scan that PR?

After this is approved / merged, we should enable this for all of our other active branches as well.

nirmoy:
Thanks for the review @nvmochs!

> Is this triggered on every push of the branch?

This trigger on every PR of the branch 24.04_linux-nvidia-6.17-next

> Is there a way to trigger this manually?

Yes, it already has a `workflow_dispatch` trigger with a `pr_number` input. You can trigger it from the Actions tab → "Patchscan - Check Missing Fixes" → Run workflow, and enter any PR number.

> After this is approved / merged, we should enable this for all of our other active branches as well.

Yes, that is the plan. Want to polish one branch 1st

nvmochs:
> Thanks for the review @nvmochs!
> 
> > Is this triggered on every push of the branch?
> 
> This trigger on every PR of the branch 24.04_linux-nvidia-6.17-next
> 
> > Is there a way to trigger this manually?
> 
> Yes, it already has a `workflow_dispatch` trigger with a `pr_number` input. You can trigger it from the Actions tab → "Patchscan - Check Missing Fixes" → Run workflow, and enter any PR number.
> 
> > After this is approved / merged, we should enable this for all of our other active branches as well.
> 
> Yes, that is the plan. Want to polish one branch 1st

Sounds like everything is covered by this PR, no issues from me.

nirmoy:
Update the Readme file to include this check

nirmoy:
Addressed all Copilot review comments:

**Workflow:**
- **Security**: Replaced PR head checkout with base branch checkout (`persist-credentials: false`) + explicit `git fetch` of the PR SHAs. The patchscan script only operates on git objects so no working tree from the PR is needed.
- **Unused `exit_code`**: Now used to fail the job immediately on patchscan runtime errors (exit non-zero before the fix-detection grep).
- **Error detection**: Added `^E:` to the grep pattern so upstream verification failures also trigger the warning comment rather than a false all-clear.
- **Comment pagination**: Both comment-posting steps now use `github.paginate` with `per_page: 100`.

**patchscan script:**
- Fixed typo: "backported comments" → "backported commits".
- Replaced bare `except:` with `except Exception as exc:` in both `is_upstream` and `check_fixes`, including the exception message in the output.
- Renamed inner loop variable `commit` → `fix_commit` in `check_fixes` to avoid shadowing the function parameter.
- `main()` now calls `sys.exit(1)` when any upstream verification fails, so the workflow detects it via `exit_code`.

nirmoy:
Addressed the three new comments:

- **`exit_code` captures `tee` exit code**: Switched to `PIPESTATUS[0]` with `set -o pipefail` to reliably capture patchscan's exit code.
- **Fetch by SHA unreliable for forks**: Changed to `refs/pull/<pr_number>/head` for the head ref, which works for both same-repo and forked PRs.
- **Range should use merge-base**: Now computes `git merge-base base head` and uses `merge_base..head_sha` as the range, so only PR-introduced commits are scanned regardless of whether the branch is behind.

nirmoy:
Addressed the three latest comments:

- **`check_fixes()` performance**: Replaced `iter_commits()` over the full range with `git log --grep "Fixes: <sha>"` via GitPython's `repo.git.log()`. This only walks commits that match the pattern rather than all commits between src and upstream head.
- **Pin Python version**: Changed `3.x` to `3.12` and added pip caching with `cache-dependency-path` pointing to `requirements.txt`.
- **Markdown injection via output**: Escape triple-backtick sequences in the output (`replace(/\`{3,}/g, '\` \` \`')`), use a 4-backtick fence to avoid conflict, and truncate output at 60,000 chars.

Regarding the `issues: write` permission comment — `pull-requests: write` is sufficient for creating/updating comments on pull requests via the Issues API endpoint. This was confirmed by the test runs (nirmoy/NV-Kernels#7 and #8) which posted comments successfully with the current permissions. No change needed there.

nirmoy:
Addressed all three:

- **`assert commits` on empty range**: Replaced with a graceful `if not commits: print(...); return` so an empty-range scan exits cleanly with 0 rather than raising.
- **Regex misses `upstream commit` pattern**: Updated `get_src_commit_strs()` regex from `from commit <sha>` to `(?:from commit|[Uu]pstream commit) <sha>`, covering `cherry picked from commit`, `backported from commit`, `upstream commit`, and `Upstream commit` formats.
- **Error exit skips PR comment**: Removed the early `exit $exit_code` — the error is now recorded as `fixes_found=error`, a new "Post PR comment on patchscan error" step posts the output to the PR before the job fails, and the final fail step handles both `error` and `true` cases with appropriate `::error::` / `::warning::` annotations.

ianm-nv (approved):
This will be nice to have.  LGTM!

`Signed-off-by: Ian May <ianm@nvidia.com>`

PR #359, opened 2026-04-14
[6.14] RDMA/core: Fix stale RoCE GIDs during netdev events at registration

nirmoy:
BugLink: https://bugs.launchpad.net/ubuntu/+source/linux-nvidia-6.17/+bug/2148311

## Summary

Cherry-pick of upstream commit 9af0feae8016ba.

RoCE GID entries become stale when netdev properties (e.g., MAC address)
change during the IB device registration window. This causes VFs to show
GIDs derived from an old/random MAC rather than the configured one,
leading to wrong GID index assignment.

The root cause is a race: `ib_enum_all_roce_netdevs()` only iterates
devices marked `DEVICE_REGISTERED`, but that mark is set late — after
the GID cache is already populated. `NETDEV_CHANGEADDR` events arriving
in this window are silently dropped.

Fix introduces a new xarray mark `DEVICE_GID_UPDATES` set immediately
after GID table allocation. `ib_enum_all_roce_netdevs()` uses this mark
instead of `DEVICE_REGISTERED`, so in-flight netdev events are no longer
lost. Also fixes the same race for IP address events.

Upstream commit: https://git.kernel.org/pub/scm/linux/kernel/git/rdma/rdma.git/commit/?id=9af0feae8016ba

### DIFF
--- a/drivers/infiniband/core/cache.c
+++ b/drivers/infiniband/core/cache.c
@@ -927,6 +927,13 @@ static int gid_table_setup_one(struct ib_device *ib_dev)
 	if (err)
 		return err;
 
+	/*
+	 * Mark the device as ready for GID cache updates. This allows netdev
+	 * event handlers to update the GID cache even before the device is
+	 * fully registered.
+	 */
+	ib_device_enable_gid_updates(ib_dev);
+
 	rdma_roce_rescan_device(ib_dev);
 
 	return err;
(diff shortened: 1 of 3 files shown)


nirmoy:
We don't have build 6.14 anymore so closing we have that for 6.17 too https://github.com/NVIDIA/NV-Kernels/pull/358

PR #419, opened 2026-05-14
Please merge MPAM branch 26.04 linux nvidia.glue.others.cpu less.hardlimit into 7.0 hwe

fyu1:
This MPAM branch is based on 7.0 hwe:

1. The first 47 patches are cherry picked from 7.1-rc upstream for MPAM part 2 glue series;
2. The patch 48 enable RESCTRL_FS in annotations
3. The patches 49-114 are VR features (iommu, cmin/cmax, cpu-less and hardlimit) and are forward ported from 6.17 hwe kernel.
4. The patch 115 fixes a couple of bugs on Grace.

Please review the PR and consider merging it into 7.0 hwe.

Thank you!

-Fenghua


fyu1:
Force pushing the PR with the fixes of the following errors:

Cherry-pick digest:
E: d20c0e3f9b89 ("NVIDIA: SAUCE: arm_mpam: resctrl: Make r"): subject differs from upstream "arm_mpam: resctrl: Make resctrl_mon_ctx_"
E: 69355752b96a ("NVIDIA: SAUCE: arm_mpam: resctrl: Fix th"): subject differs from upstream "arm_mpam: resctrl: Fix the check for no "
E: 2ccd4dc534b8 ("NVIDIA: SAUCE: arm_mpam: resctrl: Fix MB"): subject differs from upstream "arm_mpam: resctrl: Fix MBA CDP alloc_cap"

clsotog:
@fyu1 still looking at the patches but question with this untested commit c673322fc7253fe9702fe5db1d1eaeb4b3672716? is what this patch 525443bd5c05 did not have but you have a commit https://github.com/NVIDIA/NV-Kernels/commit/40e0b0792745d65ea76f7b28f2642c590fe4dd9a (at 6.17 kernel)

drivers/resctrl/mpam_devices.c
jamieNguyenNVIDIA, line 2063: In James Morse's tree, `mpam_cpu_offline()` calls the helper: `mpam_reset_msc()`: https://git.kernel.org/pub/scm/linux/kernel/git/morse/linux.git/tree/drivers/resctrl/mpam_devices.c?id=6becd89512fd69089045ca7654fdf70261c973e6#n2002

`mpam_reset_msc()` takes the outer lock: https://git.kernel.org/pub/scm/linux/kernel/git/morse/linux.git/tree/drivers/resctrl/mpam_devices.c?id=6becd89512fd69089045ca7654fdf70261c973e6#n1870

However, in our tree, the `mpam_reset_msc()` helper is gone and the code is in-lined into `mpam_cpu_offline()`.  I think you need to take the outer lock:
```
                if (atomic_dec_and_test(&msc->online_refs)) {                                                           
                        struct mpam_msc_ris *ris;                                                                       
                                                                                                                        
                        mutex_lock(&msc->cfg_lock);                                                                     
  +                     mpam_mon_sel_outer_lock(msc);                                                                   
                        list_for_each_entry_srcu(ris, &msc->ris, msc_list,                                       
                                                 srcu_read_lock_held(&mpam_srcu)) {                                   
                                mpam_touch_msc(msc, &mpam_reset_ris, ris);                                            
                                                                                                                        
                                /*                                                                                      
                                 * The reset state for non-zero partid may be                                     
                                 * lost while the CPUs are offline.                                                     
                                 */                                                                                     
                                ris->in_reset_state = false;

                                if (mpam_is_enabled())
                                        mpam_touch_msc(msc, &mpam_save_mbwu_state, ris);
                        }                                    
  +                     mpam_mon_sel_outer_unlock(msc);
                        mutex_unlock(&msc->cfg_lock);
                } 
```

drivers/acpi/arm64/mpam.c
```diff
@@ -120,6 +154,57 @@ static int acpi_mpam_parse_resource(struct mpam_msc *msc,
...
+
+	case ACPI_MPAM_LOCATION_TYPE_INTERCONNECT:
+		/* Find the descriptor table, and check it lands in the parent msc */
+		offset = res->locator.interconnect_ifc_locator.inter_connect_desc_tbl_off;
+		if (offset >= tbl_msc->length) {
+			pr_err_once("Bad offset (%lu) for interconnect descriptor on msc %u\n",
+				    offset, tbl_msc->identifier);
+			return -EINVAL;
+		}
+		tbl_int_tbl = ACPI_ADD_PTR(struct acpi_mpam_interconnect_descriptor_table,
```
jamieNguyenNVIDIA, line 180: This exists in the [original commit](https://git.kernel.org/pub/scm/linux/kernel/git/morse/linux.git/commit/?id=98b622c413ee64b8e05f93f0ff5f8cf85776afba) as well.

The bounds check here only rejects offsets at or past tbl_msc->length, but the next statements dereference tbl_int_tbl->type_uuid and tbl_int_tbl->num_descriptors.

Suggested fix:
```
        case ACPI_MPAM_LOCATION_TYPE_INTERCONNECT:                                                                      
                /* Find the descriptor table, and check it lands in the parent msc */                                   
                offset = res->locator.interconnect_ifc_locator.inter_connect_desc_tbl_off;                              
  -             if (offset >= tbl_msc->length) {                                                                        
  +             if (offset >= tbl_msc->length ||                                                                        
  +                 tbl_msc->length - offset < sizeof(*tbl_int_tbl)) {                                                  
                        pr_err_once("Bad offset (%lu) for interconnect descriptor on msc %u\n",
                                    offset, tbl_msc->identifier);
                        return -EINVAL;
                }
                tbl_int_tbl = ACPI_ADD_PTR(struct acpi_mpam_interconnect_descriptor_table,
                                           tbl_msc, offset);
```

drivers/acpi/arm64/mpam.c
```diff
@@ -120,6 +154,57 @@ static int acpi_mpam_parse_resource(struct mpam_msc *msc,
...
+					   tbl_msc, offset);
+		guid_parse(UUID_MPAM_INTERCONNECT_TABLE, &spec_uuid);
+		import_guid(&int_tbl_uuid, tbl_int_tbl->type_uuid);
+		if (guid_equal(&spec_uuid, &int_tbl_uuid)) {
```
jamieNguyenNVIDIA, line 184: This exists in the [original commit](https://git.kernel.org/pub/scm/linux/kernel/git/morse/linux.git/commit/?id=98b622c413ee64b8e05f93f0ff5f8cf85776afba) as well.

This check looks inverted (`guid_equal()` returns true when the two GUIDs match).

Suggested fix:
```
  -             if (guid_equal(&spec_uuid, &int_tbl_uuid)) {                                                            
  +             if (!guid_equal(&spec_uuid, &int_tbl_uuid)) {                                                           
                        pr_err_once("Bad UUID for interconnect descriptor on msc %u\n",                                 
                                    tbl_msc->identifier);                                                               
                        return -EINVAL;                                                                                 
                }    
```

drivers/acpi/arm64/mpam.c
```diff
@@ -120,6 +154,57 @@ static int acpi_mpam_parse_resource(struct mpam_msc *msc,
...
+				    tbl_int_tbl->num_descriptors, tbl_msc->identifier);
+			return -EINVAL;
+		}
+
+		tbl_int = ACPI_ADD_PTR(struct acpi_mpam_interconnect_descriptor,
+				       tbl_int_tbl, sizeof(*tbl_int_tbl));
+		cache_id = tbl_int->source_id;
+
+		/* Unknown link type? */
+		if (tbl_int->link_type != 0 && tbl_int->link_type == 1)
```
jamieNguyenNVIDIA, line 203: This exists in the [original commit](https://git.kernel.org/pub/scm/linux/kernel/git/morse/linux.git/commit/?id=98b622c413ee64b8e05f93f0ff5f8cf85776afba) as well.

Kind of nit-picky, but this check can be simplified:
```
  -               if (tbl_int->link_type != 0 && tbl_int->link_type == 1)
  +               if (tbl_int->link_type == 1)                                                                          
                          return 0;     
```

fs/resctrl/rdtgroup.c
```diff
@@ -791,7 +850,10 @@ static ssize_t rdtgroup_tasks_write(struct kernfs_open_file *of,
 	while (buf && buf[0] != '\0' && buf[0] != '\n') {
 		pid_str = strim(strsep(&buf, ","));
 
-		if (kstrtoint(pid_str, 0, &pid)) {
+		is_iommu = string_is_iommu_group(pid_str, &iommu_group_id);
+		if (is_iommu)
+			ret = rdtgroup_move_iommu(iommu_group_id, rdtgrp, of);
```
jamieNguyenNVIDIA, line 855: This exists in the [original commit](https://git.kernel.org/pub/scm/linux/kernel/git/morse/linux.git/commit/?id=98b622c413ee64b8e05f93f0ff5f8cf85776afba) as well.

If rdtgroup_move_iommu() returns an error here, this looks like it should break the loop.

And in the success case, it will fall through.  In this case, pid is never assigned, but we'll call `rdtgroup_move_task()` anyway.  That seems incorrect to me.

Suggested fix:
```
      if (string_is_iommu_group(pid_str, &iommu_group_id)) {
          ret = rdtgroup_move_iommu(iommu_group_id, rdtgrp, of);
          if (ret)
              break;
          continue;
      }
```

This breaks in the failure case and avoids the fall-through in the success case.

fs/resctrl/rdtgroup.c
```diff
@@ -816,6 +878,42 @@ static ssize_t rdtgroup_tasks_write(struct kernfs_open_file *of,
...
+	    !static_branch_unlikely(&resctrl_abi_playground))
+		return;
+
+	iommu_groups = iommu_get_group_kset();
+
+	while ((group_kobj = kset_get_next_obj(iommu_groups, group_kobj))) {
+		/* iommu_group_get_from_kobj() wants to drop a reference */
+		kobject_get(group_kobj);
+
+		group = iommu_group_get_from_kobj(group_kobj);
```
jamieNguyenNVIDIA, line 906: This exists in the [original commit](https://git.kernel.org/pub/scm/linux/kernel/git/morse/linux.git/commit/?id=98b622c413ee64b8e05f93f0ff5f8cf85776afba) as well.

Looks like there's a reference count leak with `group->devices_kobj` here:
```
        struct iommu_group *iommu_group_get_from_kobj(struct kobject *group_kobj)
        {
                ...
                kobject_get(group->devices_kobj);
                kobject_put(&group->kobj);
                return group;
        }
```
The loop calls `iommu_group_get_from_kobj()`, but never `iommu_group_put()`.  Suggested fix:
```
          group = iommu_group_get_from_kobj(group_kobj);                                                                
          if (!group)                                                                                                   
              continue;                                                                                                 
                                                                                                                        
          if (iommu_matches_rdtgroup(group, r))                                                                         
              seq_printf(s, "iommu_group:%s\n", group_kobj->name);                                                      
  +                                                                                                                     
  +       iommu_group_put(group);                                                                                       
      }  
```

nirmoy:

The GitHub watcher skips automatic boro reviews for PRs with more than 50 commits. This PR currently has 100 commits.

```text
review https://github.com/NVIDIA/NV-Kernels/pull/419
```

Head: `475f355d0408`

_This comment is maintained by nv-pr-bot. It is updated when the GitHub watcher sees a newer PR head._

drivers/resctrl/mpam_devices.c
```diff
@@ -1487,37 +1865,25 @@ static int mpam_save_mbwu_state(void *arg)
...
 		mbwu_state->correction += val;
 		mbwu_state->enabled = FIELD_GET(MSMON_CFG_x_CTL_EN, cur_ctl);
-		mpam_mon_sel_unlock(msc);
+		mpam_mon_sel_inner_unlock(msc);
 	}
 
 	return 0;
 }
 
-static void mpam_init_reset_cfg(struct mpam_config *reset_cfg)
```
jamieNguyenNVIDIA, line 1496: Is `mpam_init_reset_cfg()` no longer needed in 7.0? After this PR removes the CMAX/CMIN else-branches, it looks like `mpam_reset_ris()` no longer programs CMAX or CMIN on resctrl unmount or CPU online.

24.04's [e0b6de09b2a7](https://github.com/NVIDIA/NV-Kernels/commit/e0b6de09b2a78f7aa12400ee756e5e6564118578) re-introduced a refreshed `mpam_init_reset_cfg()` helper to handle this -- was it intentionally dropped from the forward-port?

nvmochs:
@fyu1 Couple of comments...

This PR is targeting the wrong branch. It should target 26.04_linux-nvidia-bos. You should submit another PR with ONLY the Grace content for 26.04_linux-nvidia.

<hr>

475f355d0408 NVIDIA: SAUCE: Fix mbm_L3_assign and mon_local_bytes

Is this fix needed in 6.17-HWE as well?

<hr>

Codex found this issue when reviewing the NVIDIA:VR:SAUCE patches

  - High: 3e04efc960a7 was not resolved correctly. The source patch adds an explicit reset config initializer; the target instead still passes an empty reset_cfg into mpam_reprogram_ris_partid() at drivers/resctrl/
    mpam_devices.c:1881. With an empty cfg, MBW_PBM falls into the else path and writes cfg->mbw_pbm as 0 at drivers/resctrl/mpam_devices.c:1726, while CMAX/CMIN are skipped unless feature bits are set at drivers/resctrl/
    mpam_devices.c:1763. That does not match the source patch intent of applying explicit default values. The annotation about reset_cpbm / reset_mbw_pbm is incomplete and the pick needs code correction.

<hr>

bc1ce80fc011 NVIDIA: VR: SAUCE: fs/resctrl: Avoid a race with dom_data_exit() and closid_num_dirty_rmid[]

Nit: It looks like this one was a clean pick and doesn’t require an annotation note under the pick tag.

<hr>

75d4de9e87cf NVIDIA: SAUCE: Update annotations to set CONFIG_RESCTRL_FS

Do we not need CONFIG_ARM64_MPAM_RESCTRL_FS = y ?

<hr> 

No issues with the patches picked from v7.1 - I verified they match upstream and picked clean.

fyu1:
No need to explicitly define ARM64_MPAM_RESCTRL_FS. The dependency chain can enable it autmatically: ONFIG_ARM64_MPAM is y debinia.master -> ARCH_HAS_CPU_RESCTRL = y -> RESCTRL_FS = y -> ARM64_MPAM_RESCTR_FS = y

fyu1:
@nvmochs "You should submit another PR with ONLY the Grace content for 26.04_linux-nvidia." Do you want the MPAM patches "only in upstream" or "ONLY the Grace"? The last patch 115 actually needs to fix code for Grace but it's not in upstream/LKML yet.

nvmochs:
> @nvmochs "You should submit another PR with ONLY the Grace content for 26.04_linux-nvidia." Do you want the MPAM patches "only in upstream" or "ONLY the Grace"? The last patch 115 actually needs to fix code for Grace but it's not in upstream/LKML yet.

In 26.04_linux-nvidia we need all MPAM patches that support Grace. My understanding is that would be the backported content from v7.1 and your single SAUCE patch (patch 115) from this PR.

fyu1:
https://github.com/NVIDIA/NV-Kernels/commit/475f355d0408698d9f632ba3b9a5dfb1f91e928b NVIDIA: SAUCE: Fix mbm_L3_assign and mon_local_bytes

Is this fix needed in 6.17-HWE as well?
6.17 hwe doesnt' have this issue. The issue is from 7.1 glue code.

drivers/resctrl/mpam_devices.c
fyu1: Will fix this in new PR.

nvmochs:
Closing PR as this PR is being replaced by https://github.com/NVIDIA/NV-Kernels/pull/429 (changes target branch to 7.0-bos)

drivers/acpi/arm64/mpam.c
fyu1: Hi, Jamie,
INTERCONNECT type is not supported on Grace and Vera. To make it work, a whole patch set needs to be created on future processors. The code here is not accurate but I will keep it so that keep close to original patch. Is that OK?

drivers/acpi/arm64/mpam.c
fyu1: Same as above comments. INTERNCONNECT type is not supported on Grace/Vera. I will keep this code as is.

fs/resctrl/rdtgroup.c
fyu1: Fixed.

fs/resctrl/rdtgroup.c
fyu1: Fixed.

drivers/acpi/arm64/mpam.c
fyu1: This code is in INTERCONNECT type. It's never executed on Grace/Vera. So I will leave it as is.

drivers/resctrl/mpam_devices.c
fyu1: mpam_init_reset_cfg() in 6.17 is to initialize reset_cfg during boot time. But upstream MPAM driver removes this function and combines the reset_cfg initialization and run time configuration together in mpam_reprogram_ris_partid() function. The upstream code is simpler and better.
So yes, mpam_init_reset_cfg() is dropped and replaced by mpam_reprogram_ris_partid().

drivers/acpi/arm64/mpam.c
fyu1: Talked to Jamie in a meeting. He agrees to keep the code as to reduce code changes.

fyu1:
> @fyu1 Couple of comments...
> 
> This PR is targeting the wrong branch. It should target 26.04_linux-nvidia-bos. You should submit another PR with ONLY the Grace content for 26.04_linux-nvidia.

Fixed.
> 
> [475f355](https://github.com/NVIDIA/NV-Kernels/commit/475f355d0408698d9f632ba3b9a5dfb1f91e928b) NVIDIA: SAUCE: Fix mbm_L3_assign and mon_local_bytes
> 
> Is this fix needed in 6.17-HWE as well?

No need. 6.17-HWE already has the fix. This fix is on top of upstream patches.

> 
> Codex found this issue when reviewing the NVIDIA:VR:SAUCE patches
> 
> * High: [3e04efc](https://github.com/NVIDIA/NV-Kernels/commit/3e04efc960a7104e8411cec17d8e9ab1b6b61606) was not resolved correctly. The source patch adds an explicit reset config initializer; the target instead still passes an empty reset_cfg into mpam_reprogram_ris_partid() at drivers/resctrl/
>   mpam_devices.c:1881. With an empty cfg, MBW_PBM falls into the else path and writes cfg->mbw_pbm as 0 at drivers/resctrl/mpam_devices.c:1726, while CMAX/CMIN are skipped unless feature bits are set at drivers/resctrl/
>   mpam_devices.c:1763. That does not match the source patch intent of applying explicit default values. The annotation about reset_cpbm / reset_mbw_pbm is incomplete and the pick needs code correction.

Fixed.

> 
> [bc1ce80](https://github.com/NVIDIA/NV-Kernels/commit/bc1ce80fc011ab3029395e63d75d6c4d67e79dc7) NVIDIA: VR: SAUCE: fs/resctrl: Avoid a race with dom_data_exit() and closid_num_dirty_rmid[]
> 
> Nit: It looks like this one was a clean pick and doesn’t require an annotation note under the pick tag.
Fixed.
> 
> [75d4de9](https://github.com/NVIDIA/NV-Kernels/commit/75d4de9e87cf9da328c121cd6afb13da716a81f2) NVIDIA: SAUCE: Update annotations to set CONFIG_RESCTRL_FS
> 
> Do we not need CONFIG_ARM64_MPAM_RESCTRL_FS = y ?
Not needed.
> 
> No issues with the patches picked from v7.1 - I verified they match upstream and picked clean.

PR #428, opened 2026-05-18
[26.04_linux-nvidia] MPAM backports and Grace fixes

fyu1:
This MPAM PR has 4 parts:

1-47: backported from upstream
48: enable RESCTRL_FS
49-52: forward ported from 6.17 hwe
53: fix issues on Grace
Please review and merge to 7.0 hwe.


nirmoy:

The GitHub watcher skips automatic boro reviews for PRs with more than 50 commits. This PR currently has 81 commits.

```text
review https://github.com/NVIDIA/NV-Kernels/pull/428
```

Head: `7bb83e678c6d`

_This comment is maintained by nv-pr-bot. It is updated when the GitHub watcher sees a newer PR head._

nvmochs:
@fyu1

cac60bd6aeb8 NVIDIA: SAUCE: arm_mpam: Include all associated
951532871c83 NVIDIA: SAUCE: arm_mpam: resctrl: Pre-allocate assignable monitors
02743bd38a8e NVIDIA: SAUCE: arm_mpam: resctrl: Pre-allocate free running monitors
43f25ff87ed9 NVIDIA: SAUCE: untested: arm_mpam: resctrl: pick classes for use as mbm counters

What function are these patches providing? It is a Grace feature?

What are the upstream plans for these patches? (It looks like there were part of the MPAM Part 2 series at one point but were dropped?)

<hr>

I verified the 47 patches from upstream were clean picks. No issues with those or with the annotations patch.

<hr>

cac60bd6aeb8 NVIDIA: SAUCE: arm_mpam: Include all associated

Codex found 2 issues with this patch...

1.

cac60bd6aeb8 drops a source hunk from ac1e5be5e8fd in drivers/resctrl/mpam_devices.c.

  The source patch changes mpam_ris_get_affinity() so memory-class components with empty affinity, or memory classes above level 3, are associated with cpu_possible_mask. The target commit does not carry that over. Current code still just
  does:

  drivers/resctrl/mpam_devices.c:505

  case MPAM_CLASS_MEMORY:
        get_cpumask_from_node_id(comp->comp_id, affinity);
        /* affinity may be empty for CPU-less memory nodes */
        break;

  The source has:

  if (cpumask_empty(affinity)) {
        dev_warn_once(..., "CPU-less numa node");
        cpumask_copy(affinity, cpu_possible_mask);
  } else if (class->level > 3)
        cpumask_copy(affinity, cpu_possible_mask);

  That matters because cac60bd6aeb8 changes CPU online/offline handling to iterate all components whose affinity contains the CPU. Without the affinity hunk, CPU-less memory nodes stay empty, and level >3 memory components stay tied to their
  NUMA node mask instead of all CPUs. That undermines the “include all associated” behavior for those components.

  The [fenghuay:] note should be improved. It currently does not mention omitting the mpam_ris_get_affinity() hunk. I think this is not just an annotation problem; the hunk should likely be added unless there is a deliberate branch-specific
  reason to omit it.

<hr>

2.

cac60bd6aeb8 duplicates for_each_mpam_resctrl_control(). The exact same macro is defined twice in drivers/resctrl/mpam_resctrl.c. That is a cleanup/build-hygiene issue, and the commit note should not say “adds” that macro if it was already present.

drivers/resctrl/mpam_devices.c
jamieNguyenNVIDIA, line 506: In pr#419, this portion of the code had:
```
		if (cpumask_empty(affinity)) {
			dev_warn_once(&msc->pdev->dev, "CPU-less numa node");
			cpumask_copy(affinity, cpu_possible_mask);
		} else if (class->level > 3)
			cpumask_copy(affinity, cpu_possible_mask);
```

Do you still need to keep the "else" case here?

nirmoy:
## BaseOS Kernel Review

Latest watcher review: [open review](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/428/latest.html)

Head: `f3404d4f7d2d`

jamieNguyenNVIDIA:
Hi @fyu1, can you please rebase this PR?

jamieNguyenNVIDIA:
LP: https://bugs.launchpad.net/ubuntu/+source/linux-nvidia-bos-7.0/+bug/2154527

drivers/resctrl/mpam_resctrl.c
```diff
@@ -0,0 +1,2192 @@
...
+ * expected total, due to rounding of the individual converted
+ * percentages) and over-commit (where the total of the converted
+ * allocations is greater than expected).
+ */
+static u16 percent_to_fract16(u8 pc, u8 wd)
+{
+	return pc ? (((pc * 65536) / 100) - 1) : 0;
+}
+
+static u32 mbw_max_to_percent(u16 mbw_max, struct mpam_props *cprops)
```
jamieNguyenNVIDIA, line 795: Alongside the RIS-reset change, this commit also rewrites the two fixed-point conversion helpers to drop the `wd` (field-width) argument:

```c
static u32 fract16_to_percent(u16 fract, u8 wd)
{
	return DIV_ROUND_CLOSEST((fract + 1) * 100, 65536);
}

static u16 percent_to_fract16(u8 pc, u8 wd)
{
	return pc ? (((pc * 65536) / 100) - 1) : 0;
}
```

A couple of things stand out:

1. The commit message only describes the RIS-reset work and doesn't mention the conversion change. (It looks like it rode along from the forward-port of `e0b6de09b2`.)
2. The comment kept above `percent_to_fract16()` still describes the value as split into `(1 << bwa_wd)` bands with the low bits ignored, which the new width-independent form no longer matches.
3. The in-tree kunit `test_percent_to_mbw_max()` (from `b4dbd4e5bc35 "arm_mpam: resctrl: Add kunit test for control format conversions"`) appears to fail. It computes `res.value = percent_to_mbw_max(pc) = percent_to_fract16(pc, width)` and asserts:

   ```c
   KUNIT_EXPECT_LE(test, res.value, res.max_value << res.shift);
   ...
   if (param->pc == 100)
       KUNIT_EXPECT_EQ(test, res.value, res.max_value << res.shift);
   ```

   For `pc == 100` the new `percent_to_fract16()` returns `0xFFFF` regardless of width, while `max_value << shift` is `((1 << width) - 1) << (16 - width)` — e.g. `0xFF00` (65280) at width 8, `0xC000` (49152) at width 2. So `0xFFFF` exceeds the upper bound and fails both the `LE` assertion and the exact-match for every implemented width below 16.

   There's also a read/write asymmetry on narrow fields: since `percent_to_fract16()` no longer left-justifies into the top `wd` bits, a value written and then read back through `fract16_to_percent()` can shift — at `bwa_wd = 2`, 50% writes `0x7FFF`, the hardware keeps `0x4000`, and the read-back reports 25%. (For `bwa_wd >= 8` the old and new forms agree, so the divergence is limited to narrow fields.)

Was dropping `wd` here intentional? If so, should the kunit and the retained comment be updated to match — or should the width handling be preserved for narrow fields?

fs/resctrl/rdtgroup.c
```diff
@@ -1688,6 +1687,30 @@ static int mbm_local_bytes_config_show(struct kernfs_open_file *of,
...
+		seq_puts(seq, "percentage\n");
+		break;
+	case RESCTRL_SCHEMA_MBPS:
+		seq_puts(seq, "mbps\n");
+		break;
+	/* The way these schema behave isn't discoverable from resctrl */
+	case RESCTRL_SCHEMA__AMD_MBA:
+		seq_puts(seq, "platform\n");
+		break;
+	}
```
jamieNguyenNVIDIA, line 1712: This switch has no `RESCTRL_SCHEMA_RANGE` case and no `default`, but RANGE is reachable for a schema that gets this file: arm_mpam assigns it to MB (`mpam_resctrl.c`, `case RDT_RESOURCE_MBA: r->schema_fmt = RESCTRL_SCHEMA_RANGE`), and `schemata_list_add()` selects it for `is_mba_sc()`. For those, none of the cases match and the function returns 0 without writing anything, so `info/MB/schema_format` reads back empty instead of a format string. The missing enumerator with no `default` will also draw `-Wswitch`. Was a "range" case intended here?

fyu1:
> duplicate for_each_mpam_resctrl_control

Duplicate for_each_mpam_resctrl_control() is fixed.

drivers/resctrl/mpam_devices.c
fyu1: @jamieNguyenNVIDIA The else is based on CPU-less feature which is not supported on Grace. On Grace, only MPAM_CLASS_CACHE is supported. This code is in MPAM_CLASS_MEMORY type and won't be triggered. When this kernel runs on Vera which supports CPU-less feature, the feature will not enabled on this kernel, which is Grace features only.

fyu1:
> @fyu1
> 
> [cac60bd](https://github.com/NVIDIA/NV-Kernels/commit/cac60bd6aeb892e09c96b10887d9c32c8561a725) NVIDIA: SAUCE: arm_mpam: Include all associated [9515328](https://github.com/NVIDIA/NV-Kernels/commit/951532871c83823c79df61fb5c744f2af210e1c3) NVIDIA: SAUCE: arm_mpam: resctrl: Pre-allocate assignable monitors [02743bd](https://github.com/NVIDIA/NV-Kernels/commit/02743bd38a8e215a7e866b0c469be422a869adc8) NVIDIA: SAUCE: arm_mpam: resctrl: Pre-allocate free running monitors [43f25ff](https://github.com/NVIDIA/NV-Kernels/commit/43f25ff87ed9532d5dbce3b18bff806c31c5ca27) NVIDIA: SAUCE: untested: arm_mpam: resctrl: pick classes for use as mbm counters
> 
> What function are these patches providing? It is a Grace feature?
> 
These are for assignment mode for mba and mbm. Without these, mba/mbm won't work. They are not Grace specific.

> What are the upstream plans for these patches? (It looks like there were part of the MPAM Part 2 series at one point but were dropped?)
They are being reviewed on LKML: https://lore.kernel.org/lkml/20260520212458.1797221-4-ben.horgan@arm.com/

> 
> I verified the 47 patches from upstream were clean picks. No issues with those or with the annotations patch.
> 
> [cac60bd](https://github.com/NVIDIA/NV-Kernels/commit/cac60bd6aeb892e09c96b10887d9c32c8561a725) NVIDIA: SAUCE: arm_mpam: Include all associated
> 
> Codex found 2 issues with this patch...
> 
> 
> [cac60bd](https://github.com/NVIDIA/NV-Kernels/commit/cac60bd6aeb892e09c96b10887d9c32c8561a725) drops a source hunk from [ac1e5be](https://github.com/NVIDIA/NV-Kernels/commit/ac1e5be5e8fddc807e9c5bbc10da3797a601bc95) in drivers/resctrl/mpam_devices.c.
> 
> The source patch changes mpam_ris_get_affinity() so memory-class components with empty affinity, or memory classes above level 3, are associated with cpu_possible_mask. The target commit does not carry that over. Current code still just does:
> 
> drivers/resctrl/mpam_devices.c:505
> 
> case MPAM_CLASS_MEMORY: get_cpumask_from_node_id(comp->comp_id, affinity); /* affinity may be empty for CPU-less memory nodes */ break;
> 
> The source has:
> 
> if (cpumask_empty(affinity)) { dev_warn_once(..., "CPU-less numa node"); cpumask_copy(affinity, cpu_possible_mask); } else if (class->level > 3) cpumask_copy(affinity, cpu_possible_mask);
> 
> That matters because [cac60bd](https://github.com/NVIDIA/NV-Kernels/commit/cac60bd6aeb892e09c96b10887d9c32c8561a725) changes CPU online/offline handling to iterate all components whose affinity contains the CPU. Without the affinity hunk, CPU-less memory nodes stay empty, and level >3 memory components stay tied to their NUMA node mask instead of all CPUs. That undermines the “include all associated” behavior for those components.

CPU-less feature is not supported on Grace. The code is for MPAM_CLASS_MEMORY type which is not supported on Grace and thus the code won't be executed on Grace. Without the code, CPU-less feature is ignore on Vera, which is expected.

> 
> The [fenghuay:] note should be improved. It currently does not mention omitting the mpam_ris_get_affinity() hunk. I think this is not just an annotation problem; the hunk should likely be added unless there is a deliberate branch-specific reason to omit it.
> 
I have updated commit message for omitting this hunk.

> 
> [cac60bd](https://github.com/NVIDIA/NV-Kernels/commit/cac60bd6aeb892e09c96b10887d9c32c8561a725) duplicates for_each_mpam_resctrl_control(). The exact same macro is defined twice in drivers/resctrl/mpam_resctrl.c. That is a cleanup/build-hygiene issue, and the commit note should not say “adds” that macro if it was already present.

Fixed.

fs/resctrl/rdtgroup.c
fyu1: Fixed.

nvmochs:
@fyu1 

Thanks for addressing my previous issues.

<hr>

d7c3069b1188 - NVIDIA: SAUCE: resctrl/mpam: reset RIS by applying explicit default config

  One other question on the reset-RIS patch: the note says reset_cfg is initialized in mpam_reprogram_ris_partid(), but I don’t see it actually being initialized there. mpam_reset_ris() passes an empty:

      struct mpam_config reset_cfg = {};

  and mpam_reprogram_ris_partid() relies on missing cfg feature bits to take reset/default paths.

  Most controls look okay with that model, but MBW_PBM looks different:

      if (mpam_has_feature(mpam_feat_mbw_part, rprops)) {
              if (mpam_has_feature(mpam_feat_mbw_part, cfg))
                      mpam_reset_msc_bitmap(msc, MPAMCFG_MBW_PBM, rprops->mbw_pbm_bits);
              else
                      mpam_write_partsel_reg(msc, MBW_PBM, cfg->mbw_pbm);
      }

  With an empty reset_cfg, that appears to write cfg->mbw_pbm == 0 rather than a full reset mask. Is mpam_feat_mbw_part/MBW_PBM also unsupported on the target platforms? If so, can you update the backport note to say that explicitly?
  Otherwise I think this may need a code adjustment to preserve the source reset behavior.

drivers/resctrl/mpam_resctrl.c
fyu1: @jamieNguyenNVIDIA fract16_to_percent() and percent_to_fract16() has wrong calculations. They don't consider wd and are obviously wrong. The updated PR removes them.

fyu1:
Changes in the updated PR:
1. ABMC code uses latest patches in https://gitlab.arm.com/linux-arm/linux-bh.git  mpam_abmc_v4
2. cmin/cmax are still same as last time.
3. Removed the "Include all MSC components..." patch which is for numa node and is unnecessary for Grace.

jamieNguyenNVIDIA:
The rebase appears to have dropped the downstream commit

  3a720efce0c1 ("NVIDIA: SAUCE: arm_mpam: Include all associated")

which changed CPU-online domain setup to cover all MPAM components
associated with a CPU rather than the first match. mpam_resctrl_alloc_domain()
now picks only the first component again:

    list_for_each_entry_srcu(comp_iter, &class->components, class_list, ...) {
        if (cpumask_test_cpu(cpu, &comp_iter->affinity)) {
            ctrl_comp = comp_iter;
            break;
        }
    }

Was dropping that intentional with the move to the newer upstream
snapshot, or did it get lost in the rebase?

drivers/resctrl/mpam_resctrl.c
```diff
@@ -0,0 +1,2060 @@
...
+/*
+ * The classes we've picked to map to resctrl events.
+ * Resctrl believes all the worlds a Xeon, and these are all on the L3. This
+ * array lets us find the actual class backing the event counters. e.g.
+ * the only memory bandwidth counters may be on the memory controller, but to
+ * make use of them, we pretend they are on L3. Restrict the events considered
+ * to those supported by MPAM.
+ * Class pointer may be NULL.
+ */
+#define MPAM_MAX_EVENT QOS_L3_MBM_TOTAL_EVENT_ID
```
jamieNguyenNVIDIA, line 48: This adds resctrl_arch_mon_can_overflow(), which lets the resctrl core
skip the per-domain overflow handler when the counters are wide enough
not to wrap.  It reads the backing class out of mpam_resctrl_counters[]
for both MBM events.

```
+bool resctrl_arch_mon_can_overflow(void)
+{
+	if (__resctrl_arch_mon_can_overflow(QOS_L3_MBM_LOCAL_EVENT_ID))
+		return true;
```

mpam_resctrl_counters[] is sized MPAM_MAX_EVENT + 1, and MPAM_MAX_EVENT
is QOS_L3_MBM_TOTAL_EVENT_ID (0x02), so the array has three slots
(indices 0..2).  QOS_L3_MBM_LOCAL_EVENT_ID is 0x03, so this reads
mpam_resctrl_counters[3].class, one past the end, and dereferences it as
a struct mpam_class *.

The path is reachable from mbm_setup_overflow_handler() once resctrl is
mounted on a monitor-capable system.

An earlier version of the series sized the array to the local event,
which kept this read in bounds; that change was lost in the rebase.
The slot is static-zeroed, so could MPAM_MAX_EVENT be restored to the
local event, letting can_overflow() read a NULL class and fall through
to the total probe?

```
  -#define MPAM_MAX_EVENT QOS_L3_MBM_TOTAL_EVENT_ID
  +#define MPAM_MAX_EVENT QOS_L3_MBM_LOCAL_EVENT_ID
```

nvmochs:
> @fyu1
> 
> Thanks for addressing my previous issues.
> 
> [d7c3069](https://github.com/NVIDIA/NV-Kernels/commit/d7c3069b1188d456cb211a212cd861494e92d67a) - NVIDIA: SAUCE: resctrl/mpam: reset RIS by applying explicit default config
> 
> One other question on the reset-RIS patch: the note says reset_cfg is initialized in mpam_reprogram_ris_partid(), but I don’t see it actually being initialized there. mpam_reset_ris() passes an empty:
> 
> ```
>   struct mpam_config reset_cfg = {};
> ```
> 
> and mpam_reprogram_ris_partid() relies on missing cfg feature bits to take reset/default paths.
> 
> Most controls look okay with that model, but MBW_PBM looks different:
> 
> ```
>   if (mpam_has_feature(mpam_feat_mbw_part, rprops)) {
>           if (mpam_has_feature(mpam_feat_mbw_part, cfg))
>                   mpam_reset_msc_bitmap(msc, MPAMCFG_MBW_PBM, rprops->mbw_pbm_bits);
>           else
>                   mpam_write_partsel_reg(msc, MBW_PBM, cfg->mbw_pbm);
>   }
> ```
> 
> With an empty reset_cfg, that appears to write cfg->mbw_pbm == 0 rather than a full reset mask. Is mpam_feat_mbw_part/MBW_PBM also unsupported on the target platforms? If so, can you update the backport note to say that explicitly? Otherwise I think this may need a code adjustment to preserve the source reset behavior.

@fyu1 In the latest update (Jun 3 - range: a71d5d9cf0d2^..712640a78a02) this does not appear to be addressed.

The current replacement appears to be 712640a78a02 rather than d7c3069b1188, but the relevant code is still the same:

  - drivers/resctrl/mpam_devices.c:1721 still creates struct mpam_config reset_cfg = {};
  - drivers/resctrl/mpam_devices.c:1571 still has the suspicious MBW_PBM logic:

```
  if (mpam_has_feature(mpam_feat_mbw_part, rprops)) {
  	if (mpam_has_feature(mpam_feat_mbw_part, cfg))
  		mpam_reset_msc_bitmap(msc, MPAMCFG_MBW_PBM, rprops->mbw_pbm_bits);
  	else
  		mpam_write_partsel_reg(msc, MBW_PBM, cfg->mbw_pbm);
  }
```
  With the empty reset_cfg, this still takes the else path and writes cfg->mbw_pbm == 0.

  The annotation was changed since d7c3069b1188, but not in a way that resolves the above finding. Current 712640a78a02 still says reset_cfg is initialized in mpam_reprogram_ris_partid(&reset_cfg), and it still does not say
  mpam_feat_mbw_part / MBW_PBM is unsupported on the target platforms.

The current note either needs to explicitly justify why MBW_PBM cannot matter here, or the code needs to preserve the source reset behavior for MBW_PBM.

nvmochs:
@fyu1 

Same comment as Jamie about the missing "NVIDIA: SAUCE: arm_mpam: Include all associated” patch.

<hr>

712640a78a02 NVIDIA: SAUCE: resctrl/mpam: reset RIS by applying explicit default config

This patch is documented as a forward-port of e0b6de09b2a7, but it only carries a tiny subset of the source behavior.

  In source e0b6de09b2a7, the patch is large:

  - removes mpam_reset_msc_bitmap()
  - removes reset_cpbm / reset_mbw_pbm from struct mpam_config
  - changes mpam_reprogram_ris_partid() so reset is driven by an explicit default mpam_config
  - adds/uses mpam_init_reset_cfg(&reset_cfg, &ris->props)
  - changes mpam_reset_component_cfg() to initialize explicit defaults and feature bits
  - changes fract16_to_percent() / percent_to_fract16()

  But picked commit 712640a78a02 only adds:

  if (cprops->cmax_wd)
          comp->cfg[i].cmax = MPAMCFG_CMAX_CMAX;

  in drivers/resctrl/mpam_devices.c:2632.

  That means the annotation is misleading in several places:

  - It says reset_cpbm / reset_mbw_pbm do not need to be defined, but that is not really the meaningful forward-port explanation. The picked tree still uses the older reset helper path via drivers/resctrl/mpam_devices.c:1441.
  - It says mpam_init_reset_cfg() was removed upstream and reset_cfg is initialized in mpam_reprogram_ris_partid(&reset_cfg), but the current code still passes an empty reset_cfg from drivers/resctrl/mpam_devices.c:1721.
  - It says fract16_to_percent() and percent_to_fract16() were removed, but both still exist in drivers/resctrl/mpam_resctrl.c:764.

  So my concern is not just “patch-id mismatch”; it is that the annotation claims several source hunks were handled, while the actual picked commit only initializes cmax in component config. Either the source behavior was
  already present/dropped elsewhere and the annotation should say that clearly, or this forward-port missed most of the source patch.

fyu1:
> The rebase appears to have dropped the downstream commit
> 
> [3a720ef](https://github.com/NVIDIA/NV-Kernels/commit/3a720efce0c10b5e280ffda83887c6a613d6ea45) ("NVIDIA: SAUCE: arm_mpam: Include all associated")
> 
> which changed CPU-online domain setup to cover all MPAM components associated with a CPU rather than the first match. mpam_resctrl_alloc_domain() now picks only the first component again:
> 
> ```
> list_for_each_entry_srcu(comp_iter, &class->components, class_list, ...) {
>     if (cpumask_test_cpu(cpu, &comp_iter->affinity)) {
>         ctrl_comp = comp_iter;
>         break;
>     }
> }
> ```
> 
> Was dropping that intentional with the move to the newer upstream snapshot, or did it get lost in the rebase?

Yes, the patch is dropped because it's CPU-less patch. Grace doesn't support CPU-less feature. On Grace, each component (L3 or mem) is associated to only on CPU. With CPU-less node, a component on a CPU-less node is associated on any CPU. So need to add multiple components to a CPU. We don't need that on Grace.

drivers/resctrl/mpam_resctrl.c
fyu1: I may need to change the code in 38e1ae53ec23. Latest LKML code only support TOTAL_EVENT_ID. So I'm planning to change overflow function to this:
+bool resctrl_arch_mon_can_overflow(void)
+{
+       return __resctrl_arch_mon_can_overflow(QOS_L3_MBM_TOTAL_EVENT_ID);
+}

fyu1:
> @fyu1
> 
> Thanks for addressing my previous issues.
> 
> [d7c3069](https://github.com/NVIDIA/NV-Kernels/commit/d7c3069b1188d456cb211a212cd861494e92d67a) - NVIDIA: SAUCE: resctrl/mpam: reset RIS by applying explicit default config
> 
> One other question on the reset-RIS patch: the note says reset_cfg is initialized in mpam_reprogram_ris_partid(), but I don’t see it actually being initialized there. mpam_reset_ris() passes an empty:
> 
> ```
>   struct mpam_config reset_cfg = {};
> ```
> 

mpam_reset_ris() is called in init path and effectively initialize a few registers.

mpam_reprogram_ris_partid(*cfg) takes empty reset_cfg. Then if cfg is empty and therefore no feature bits, the function directly writes the registers with initial values. 
e.g.
                if (mpam_has_feature(mpam_feat_mbw_max, cfg))
                        mpam_write_partsel_reg(msc, MBW_MAX, cfg->mbw_max);
                else
                        mpam_write_partsel_reg(msc, MBW_MAX, MPAMCFG_MBW_MAX_MAX);
empty cfg will write max MBW value to the register.

> and mpam_reprogram_ris_partid() relies on missing cfg feature bits to take reset/default paths.
> 
> Most controls look okay with that model, but MBW_PBM looks different:
> 
> ```
>   if (mpam_has_feature(mpam_feat_mbw_part, rprops)) {
>           if (mpam_has_feature(mpam_feat_mbw_part, cfg))
>                   mpam_reset_msc_bitmap(msc, MPAMCFG_MBW_PBM, rprops->mbw_pbm_bits);
>           else
>                   mpam_write_partsel_reg(msc, MBW_PBM, cfg->mbw_pbm);
>   }
> ```
> 
> With an empty reset_cfg, that appears to write cfg->mbw_pbm == 0 rather than a full reset mask. Is mpam_feat_mbw_part/MBW_PBM also unsupported on the target platforms? If so, can you update the backport note to say that explicitly? Otherwise I think this may need a code adjustment to preserve the source reset behavior.

MBW_PBM is not supported on Grace/Vera. This code is upstream code. I'll fix the code and update commit message to note this. Will send the fix to lkml.

fyu1:
> @fyu1
> 
> Same comment as Jamie about the missing "NVIDIA: SAUCE: arm_mpam: Include all associated” patch.
> 
> [712640a](https://github.com/NVIDIA/NV-Kernels/commit/712640a78a022e4500bb899d0c6d2275a159bbd1) NVIDIA: SAUCE: resctrl/mpam: reset RIS by applying explicit default config
> 
> This patch is documented as a forward-port of [e0b6de0](https://github.com/NVIDIA/NV-Kernels/commit/e0b6de09b2a78f7aa12400ee756e5e6564118578), but it only carries a tiny subset of the source behavior.
> 
> In source [e0b6de0](https://github.com/NVIDIA/NV-Kernels/commit/e0b6de09b2a78f7aa12400ee756e5e6564118578), the patch is large:
> 
> * removes mpam_reset_msc_bitmap()
> * removes reset_cpbm / reset_mbw_pbm from struct mpam_config
> * changes mpam_reprogram_ris_partid() so reset is driven by an explicit default mpam_config
> * adds/uses mpam_init_reset_cfg(&reset_cfg, &ris->props)
> * changes mpam_reset_component_cfg() to initialize explicit defaults and feature bits
> * changes fract16_to_percent() / percent_to_fract16()
> 
> But picked commit [712640a](https://github.com/NVIDIA/NV-Kernels/commit/712640a78a022e4500bb899d0c6d2275a159bbd1) only adds:
> 
> if (cprops->cmax_wd) comp->cfg[i].cmax = MPAMCFG_CMAX_CMAX;
> 
> in drivers/resctrl/mpam_devices.c:2632.
> 
> That means the annotation is misleading in several places:
> 
> * It says reset_cpbm / reset_mbw_pbm do not need to be defined, but that is not really the meaningful forward-port explanation. The picked tree still uses the older reset helper path via drivers/resctrl/mpam_devices.c:1441.
> * It says mpam_init_reset_cfg() was removed upstream and reset_cfg is initialized in mpam_reprogram_ris_partid(&reset_cfg), but the current code still passes an empty reset_cfg from drivers/resctrl/mpam_devices.c:1721.
> * It says fract16_to_percent() and percent_to_fract16() were removed, but both still exist in drivers/resctrl/mpam_resctrl.c:764.
I changed the change log to say "changes of fract16_to_percent() and percent_to_fract16() were removed.." to make it clear that only changes in the previous commit were removed, not the entire functions.

> 
> So my concern is not just “patch-id mismatch”; it is that the annotation claims several source hunks were handled, while the actual picked commit only initializes cmax in component config. Either the source behavior was already present/dropped elsewhere and the annotation should say that clearly, or this forward-port missed most of the source patch.
As discussed on meeting, I will keep the commit message and add the changes in the change log to make tracking easier.

fyu1:
> > @fyu1
> > Thanks for addressing my previous issues.
> > [d7c3069](https://github.com/NVIDIA/NV-Kernels/commit/d7c3069b1188d456cb211a212cd861494e92d67a) - NVIDIA: SAUCE: resctrl/mpam: reset RIS by applying explicit default config
> > One other question on the reset-RIS patch: the note says reset_cfg is initialized in mpam_reprogram_ris_partid(), but I don’t see it actually being initialized there. mpam_reset_ris() passes an empty:
> > ```
> >   struct mpam_config reset_cfg = {};
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and mpam_reprogram_ris_partid() relies on missing cfg feature bits to take reset/default paths.
> > Most controls look okay with that model, but MBW_PBM looks different:
> > ```
> >   if (mpam_has_feature(mpam_feat_mbw_part, rprops)) {
> >           if (mpam_has_feature(mpam_feat_mbw_part, cfg))
> >                   mpam_reset_msc_bitmap(msc, MPAMCFG_MBW_PBM, rprops->mbw_pbm_bits);
> >           else
> >                   mpam_write_partsel_reg(msc, MBW_PBM, cfg->mbw_pbm);
> >   }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > With an empty reset_cfg, that appears to write cfg->mbw_pbm == 0 rather than a full reset mask. Is mpam_feat_mbw_part/MBW_PBM also unsupported on the target platforms? If so, can you update the backport note to say that explicitly? Otherwise I think this may need a code adjustment to preserve the source reset behavior.
> 
> @fyu1 In the latest update (Jun 3 - range: [a71d5d9](https://github.com/NVIDIA/NV-Kernels/commit/a71d5d9cf0d27d51370f04f51a59506356dd56fb)^..[712640a](https://github.com/NVIDIA/NV-Kernels/commit/712640a78a022e4500bb899d0c6d2275a159bbd1)) this does not appear to be addressed.
> 
> The current replacement appears to be [712640a](https://github.com/NVIDIA/NV-Kernels/commit/712640a78a022e4500bb899d0c6d2275a159bbd1) rather than [d7c3069](https://github.com/NVIDIA/NV-Kernels/commit/d7c3069b1188d456cb211a212cd861494e92d67a), but the relevant code is still the same:
> 
> * drivers/resctrl/mpam_devices.c:1721 still creates struct mpam_config reset_cfg = {};
> * drivers/resctrl/mpam_devices.c:1571 still has the suspicious MBW_PBM logic:
> 
> ```
>   if (mpam_has_feature(mpam_feat_mbw_part, rprops)) {
>   	if (mpam_has_feature(mpam_feat_mbw_part, cfg))
>   		mpam_reset_msc_bitmap(msc, MPAMCFG_MBW_PBM, rprops->mbw_pbm_bits);
>   	else
>   		mpam_write_partsel_reg(msc, MBW_PBM, cfg->mbw_pbm);
>   }
> ```
> 
> With the empty reset_cfg, this still takes the else path and writes cfg->mbw_pbm == 0.
> 
> The annotation was changed since [d7c3069](https://github.com/NVIDIA/NV-Kernels/commit/d7c3069b1188d456cb211a212cd861494e92d67a), but not in a way that resolves the above finding. Current [712640a](https://github.com/NVIDIA/NV-Kernels/commit/712640a78a022e4500bb899d0c6d2275a159bbd1) still says reset_cfg is initialized in mpam_reprogram_ris_partid(&reset_cfg), and it still does not say mpam_feat_mbw_part / MBW_PBM is unsupported on the target platforms.
> 
> The current note either needs to explicitly justify why MBW_PBM cannot matter here, or the code needs to preserve the source reset behavior for MBW_PBM.

fyu1:
The updated PR has the following changes:
1. Fix MBW_PBM setting in mpam_reprogram_ris_partid() (Matt)
2. Remove QOS_L3_MBM_LOCAL_EVENT_ID checking in resctrl_arch_mon_can_overflow() (Jamie)
3. Change change log in 91f12a38 to clarify a lot of changes from original 6.17 hwe commit (Matt and Jamie)

nvmochs:
@fyu1

I re-reviewed the latest and it looks much better, thanks!

One minor nit: the new 91f12a38b598 note has typos: irrevelant and messge.

Other than that, I will wait for you to replace the inline MBW_PBM fix with a standalone fix commit as discussed during the scrub meeting this AM. Once that is present I think we can get this merged.

fyu1:
> @fyu1
> 
> I re-reviewed the latest and it looks much better, thanks!
> 
> One minor nit: the new [91f12a3](https://github.com/NVIDIA/NV-Kernels/commit/91f12a38b598e0ca9e1e6802d334b4f414201b71) note has typos: irrevelant and messge.
> 
> Other than that, I will wait for you to replace the inline MBW_PBM fix with a standalone fix commit as discussed during the scrub meeting this AM. Once that is present I think we can get this merged.

Fixed in the updated PR.

nvmochs:
> > @fyu1
> > I re-reviewed the latest and it looks much better, thanks!
> > One minor nit: the new [91f12a3](https://github.com/NVIDIA/NV-Kernels/commit/91f12a38b598e0ca9e1e6802d334b4f414201b71) note has typos: irrevelant and messge.
> > Other than that, I will wait for you to replace the inline MBW_PBM fix with a standalone fix commit as discussed during the scrub meeting this AM. Once that is present I think we can get this merged.
> 
> Fixed in the updated PR.

Thanks, no further issues from me.

jamieNguyenNVIDIA:
Thanks @fyu1!

nvmochs:
Merged, closing PR.

```
8058d21f2169 (nresolute/main-next) NVIDIA: SAUCE: arm_mpam: Fix MPAMCFG_MBW_PBM register setting
03099786ac8a NVIDIA: SAUCE: resctrl/mpam: reset RIS by applying explicit default config
7285ebdec9ef NVIDIA: SAUCE: arm_mpam: resctrl: Add the glue code to convert to/from cmax
7a3f0c979054 NVIDIA: SAUCE: fs/resctrl: Add L2 and L3 'MAX' resource schema
c9ab72092b0a NVIDIA: SAUCE: fs/resctrl: Expose the schema format to user-space
48d5fa1976b9 NVIDIA: SAUCE: fs/resctrl: Add fflags_from_schema() for files based on schema format
26ab445f8bbc NVIDIA: SAUCE: fs/resctrl: Add additional files for percentage and bitmap controls
45dc9f3fcb9c NVIDIA: SAUCE: x86/resctrl: Move over to specifying MBA control formats
94f285443876 NVIDIA: SAUCE: fs/resctrl: Add specific schema types for 'range'
0837cb25efc4 NVIDIA: SAUCE: fs/resctrl: Add a schema format to the schema, allowing it to be different
f99a62d3d21d NVIDIA: SAUCE: fs/resctrl: Rename resctrl_get_default_ctrl() to include resource
80efae835022 NVIDIA: SAUCE: fs/resctrl: Move MBA supported check to parse_line() instead of parse_bw()
8d336944de32 NVIDIA: SAUCE: fs/resctrl: Abstract duplicate domain test to a helper
a9bb75b1736f NVIDIA: SAUCE: fs/resctrl: Group all the MBA specific properties in a separate struct
d6e777d59060 NVIDIA: SAUCE: arm_mpam: Rename mbw conversion to 'fract16' for code re-use
72cc0a884f17 NVIDIA: SAUCE: arm_mpam: Allow cmax/cmin to be configured
b45dc6b96bb5 NVIDIA: SAUCE: fs/restrl: Allow the overflow handler to be disabled
00cc043bba32 NVIDIA: SAUCE: arm_mpam: resctrl: Determine if any exposed counter can overflow
74d5bc48be9c NVIDIA: SAUCE: x86/resctrl: Add stub to allow other architecture to disable monitor overflow
16cbbde45f31 NVIDIA: SAUCE: fs/resctrl,x86/resctrl: Factor mba rounding to be per-arch
dcdcea6d2e04 NVIDIA: SAUCE: arm_mpam: Add resctrl_arch_round_bw()
58bb2cfa2d32 NVIDIA: SAUCE: arm64: mpam: Add memory bandwidth usage (MBWU) documentation
72c0ce56d304 NVIDIA: SAUCE: arm_mpam: resctrl: Add resctrl_arch_cntr_read() & resctrl_arch_reset_cntr()
570cfb7c3e57 NVIDIA: SAUCE: arm_mpam: resctrl: Add resctrl_arch_config_cntr() for ABMC use
b06156d53dda NVIDIA: SAUCE: arm_mpam: resctrl: Pre-allocate assignable monitors
b0ab6949d273 NVIDIA: SAUCE: arm_mpam: resctrl: Pick classes for use as MBM counters
c21a154e2b18 NVIDIA: SAUCE: fs/resctrl: Document tasks file behaviour for task id 0 and idle tasks
021ab1ff6a35 NVIDIA: SAUCE: fs/resctrl: Document that automatic counter assignment is best effort
ddd72c46336e NVIDIA: SAUCE: fs/resctrl: Continue counter allocation after failure
134a1fa649cc NVIDIA: SAUCE: fs/resctrl: Add monitor property 'mbm_cntr_assign_fixed'
87db39292513 NVIDIA: SAUCE: fs/resctrl: Disallow the software controller when MBM counters are assignable
befbdfc848ef NVIDIA: SAUCE: x86,fs/resctrl: Create 'event_filter' files read only if they're not configurable
0a99b06ee1d3 NVIDIA: SAUCE: fs/resctrl: Tidy up the error path in resctrl_mkdir_event_configs()
ca01c1e669da NVIDIA: SAUCE: Update annotations to set CONFIG_RESCTRL_FS
0cbeaa4d29e4 arm_mpam: resctrl: Make resctrl_mon_ctx_waiters static
3e785fde96b3 arm_mpam: resctrl: Fix the check for no monitor components found
b015233c3e8f arm_mpam: resctrl: Fix MBA CDP alloc_capable handling on unmount
5112d1427f85 fs/resctrl: Add missing return value descriptions
974b4d21bda3 MAINTAINERS: Update resctrl entry
b172e8052357 fs/resctrl: Add "*" shorthand to set io_alloc CBM for all domains
5644b673c6c4 fs/resctrl: Report invalid domain ID when parsing io_alloc_cbm
abfd18a0360d arm64: mpam: Add initial MPAM documentation
544708148c8e arm_mpam: Quirk CMN-650's CSU NRDY behaviour
2fb5171d408f arm_mpam: Add workaround for T241-MPAM-6
0b7be4293b87 arm_mpam: Add workaround for T241-MPAM-4
22182656fe56 arm_mpam: Add workaround for T241-MPAM-1
141328fbbc34 arm_mpam: Add quirk framework
8a9f594a7696 arm_mpam: resctrl: Call resctrl_init() on platforms that can support resctrl
88fa80690584 arm64: mpam: Select ARCH_HAS_CPU_RESCTRL
b1cb514cb259 ALSA: usb-audio: Replace hard-coded number with MAX_CHANNELS
d12a4f2c6c85 arm_mpam: resctrl: Add empty definitions for assorted resctrl functions
d0b0ca7d7599 arm_mpam: resctrl: Update the rmid reallocation limit
27f7ba734254 arm_mpam: resctrl: Add resctrl_arch_rmid_read()
eaf5fc03c842 arm_mpam: resctrl: Allow resctrl to allocate monitors
bd7b97bfb404 arm_mpam: resctrl: Add support for csu counters
1ec94405653e arm_mpam: resctrl: Add monitor initialisation and domain boilerplate
bdaa9732e677 arm_mpam: resctrl: Add kunit test for control format conversions
45e163da5ee3 arm_mpam: resctrl: Add support for 'MB' resource
86f630ed2ecf arm_mpam: resctrl: Wait for cacheinfo to be ready
1d7a40083ed7 arm_mpam: resctrl: Add rmid index helpers
e9aaa4cf03e6 arm_mpam: resctrl: Convert to/from MPAMs fixed-point formats
1f74bfbde43c arm_mpam: resctrl: Hide CDP emulation behind CONFIG_EXPERT
60d99560705c arm_mpam: resctrl: Add CDP emulation
af87ba985b51 arm_mpam: resctrl: Add plumbing against arm64 task and cpu hooks
d81850d3f58b arm_mpam: resctrl: Implement helpers to update configuration
a8648cc8dce2 arm_mpam: resctrl: Add resctrl_arch_get_config()
3f84f3db5d53 arm_mpam: resctrl: Implement resctrl_arch_reset_all_ctrls()
55687c5510e0 arm_mpam: resctrl: Pick the caches we will use as resctrl resources
f8e794b0e73d arm_mpam: resctrl: Add boilerplate cpuhp and domain allocation
b5fb7023d65e arm64: mpam: Add helpers to change a task or cpu's MPAM PARTID/PMG values
ac3902a1ffc0 arm64: mpam: Initialise and context switch the MPAMSM_EL1 register
9a7c92166bde arm64: mpam: Add cpu_pm notifier to restore MPAM sysregs
59911264c94d arm64: mpam: Advertise the CPUs MPAM limits to the driver
92020eb8e7be arm64: mpam: Drop the CONFIG_EXPERT restriction
edfdd3bb7522 arm64: mpam: Re-initialise MPAM regs when CPU comes online
33813629cfba arm64: mpam: Context switch the MPAM registers
a9dc00296970 KVM: arm64: Make MPAMSM_EL1 accesses UNDEF
714682ba2589 KVM: arm64: Preserve host MPAM configuration when changing traps
5f3540600db9 arm64/sysreg: Add MPAMSM_EL1 register
676aad9bb363 arm_mpam: Reset when feature configuration bit unset
2aeac8384422 arm_mpam: Ensure in_reset_state is false after applying configuration
```

PR #448, opened 2026-06-02
[26.04_linux-nvidia-bos] VFIO: Add CXL Type-2 passthrough support for BOS

kobak2026:
## Summary

Backport the VFIO CXL Type-2 passthrough stack onto `26.04_linux-nvidia-bos`.

This series adds VFIO CXL device detection, CXL UAPI exposure, DVSEC/HDM emulation, DPA and component register region plumbing, reset handling, selftest coverage, and enables `CONFIG_VFIO_CXL_CORE` for BOS `amd64` and `arm64`.

The BOS port also carries a conservative DPA mmap policy: DPA is exposed as read/write but not mmap-capable until CXL core can prove CPU-readable backing for the CXL HPA. This avoids a reproducible userspace `SIGBUS` in `cxl_type2.dpa_mmap_fault` while preserving the rest of VFIO CXL functionality.

## Validation

- Verified `CONFIG_VFIO_CXL_CORE=y` for:
  - `arm64-nvidia-bos-64k`
  - `amd64-nvidia-bos`
- Built BOS arm64 raw kernel:
  - `7.0.0-2008-nvidia-bos-64k`
- Installed and booted on:
  - `vr-nvl72-eba-l10-comp071`
- Confirmed CXL Type-2 endpoints are active:
  - `FBSta: Cache+ IO+ Mem+`
- Ran `vfio_cxl_type2_test 0002:81:00.0`.

Validated selftest cases:
- `cxl_type2.device_info_has_cxl`
- `cxl_type2.device_info_cap_cxl_payload`
- `cxl_type2.region_info_cxl_capability`
- `cxl_type2.component_bar_sparse_mmap`
- `cxl_type2.dpa_region_info`
- `cxl_type2.comp_regs_region_info`
- `cxl_type2.dpa_mmap_fault`
- `cxl_type2.comp_regs_no_mmap`
- `cxl_type2.comp_reg_mmap_blocked`
- `cxl_type2.hdm_cap_read`
- `cxl_type2.hdm_ctrl_commit_to_committed`
- `cxl_type2.dvsec_lock_semantics`

Final result after the DPA mmap guard:
- `pass: 10`
- `fail: 0`
- `skip: 1`
- `error: 0`

[...]

PR #486, opened 2026-07-09
[26.04_linux-nvidia-bos] vfio/cxl: Advertise DPA mmap and update its selftest

mmhonap:
https://bugs.launchpad.net/ubuntu/+source/linux-nvidia-bos/+bug/2160591

The DPA region needs VFIO_REGION_INFO_FLAG_MMAP so a VMM can mmap it and map it into stage-2 for the device to reach over ATS. Without the flag the VMM keeps DPA as a slow-path I/O region that the stage-2 listener skips, and the GPU's ATS access to DPA-resident memory never resolves.

The dpa_mmap_fault selftest is skipped when the flag is absent. With the flag set it runs, but its raw host CPU load/store of DPA does not model how a VMM uses the region: the VMM mmaps DPA and maps it into the IOAS, and the guest reaches it through stage-2. Rewrite the test to do the same (mmap plus iommu_map) and drop the host dereference.

### DIFF
--- a/drivers/vfio/pci/cxl/vfio_cxl_core.c
+++ b/drivers/vfio/pci/cxl/vfio_cxl_core.c
@@ -1034,6 +1034,7 @@ static void vfio_cxl_region_release(struct vfio_pci_core_device *vdev,
 				    cxl->region_size, true);
 
 	if (cxl->region_vaddr) {
+		unregister_pfn_address_space(&cxl->dpa_pfn_space);
 		memunmap(cxl->region_vaddr);
 		cxl->region_vaddr = NULL;
 	}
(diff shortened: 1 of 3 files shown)


nirmoy:
## BaseOS Kernel Review

### Summary

The vfio/cxl selftest has two unchecked-return NULL/error paths: iova_allocator_init() result used without a NULL check, and iommu_map() failure ignored before an unconditional iommu_unmap(), both of which can crash the test. Minor comment/wording issues also noted.

Findings: Critical: 0, High: 0, Medium: 2, Low: 2

Latest watcher review: [open review](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/486/latest.html)

Generated test plan: [open test plan](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/486/test-plan.html)

Kernel deb build: **successful** ([download debs](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/486/debs/), 4 files)

Head: `0c40c2c481af`

clsotog:
@mmhonap 
This patch is not upstream and Im guessing only to VR? if yes can you add to the commit title NVIDIA: VR: SAUCE: 
If not specific to VR then add to the commit tittle NVIDIA: SAUCE: 
You can check in the git log of the tree examples with this tag at the commit. 
Also to the title of the PR can you add [26.04_linux-nvidia-bos]. You can check other PR titles. 
After those changes we can start ack the PR to submit to Canonical. 
Thanks

nirmoy:
This is coming from https://github.com/NVIDIA/NV-Kernels/pull/448 

[NVIDIA: VR: SAUCE: vfio/cxl: Register regions with VFIO layer · NVIDIA/NV-Kernels@e658757](https://github.com/NVIDIA/NV-Kernels/commit/e658757fac34c0f7713d95ceb4f81030fcd5f22c#diff-c76f3ebfd8079c14e6e99048f142074db78600feb4004b0b5947e4c3bdb4d682R547-R717)

I thought it came from upstream but that is not true. Cc @kobak2026

kobak2026 (commented):
I validated this on 10.103.232.135 with the patch applied on
  top of 26.04_linux-nvidia-bos.
  Host kernel:
  7.0.0-2015-pr486-nvidia-bos-64k
  Result:
  vfio_cxl_type2_test passed 11/11 on 0002:81:00.0.
  The updated dpa_mmap_fault test passes with:
    DPA mmap + IOAS map ok: iova=0x0 size=0x200000
  I also rebuilt the old pre-PR selftest behavior, where
  dpa_mmap_fault mmap()s the DPA region and directly
  dereferences the host pointer. That still reproduces the
  original SIGBUS on the same PR kernel:
    pass:10 fail:1 skip:0 error:0
    dpa_mmap_fault: Test terminated unexpectedly by signal 7
  So the evidence matches the commit rationale: the old
  failure is from host CPU dereference of DPA, while the PR
  tests the VMM path of mmap + IOAS mapping.
  One review comment:
  Since this patch requires DPA MMAP for the VMM/ATS path, the
  selftest should probably fail if VFIO_REGION_INFO_FLAG_MMAP
  is missing instead of SKIPing. Otherwise a future
  regression that drops MMAP again would be hidden as a
  skipped test.
  Minor nit:
  The test name dpa_mmap_fault is now stale. It no longer
  verifies host CPU fault/dereference behavior; it verifies
  DPA mmap + IOAS mapping. A rename or comment update would
  make the intent clearer.

nirmoy (approved):

nirmoy:
@mmhonap I forgot please add "NVIDIA: VR: SAUCE:" tag for the patch as this is a internal patch

mmhonap:
Thank you for the comments Nirmoy/Koba.

1. I have updated the commit message to have the internal tags. The commit now also makes use of register_pfn_address_space as demonstrated in nvgrace code. This will add additional containment against host SErrors from user space accesses.

2. Added fixes by tag and acked by tags as suggested.

3. Added the check suggested in review.

4. Renamed the change to dpa_mmap_ioas_map

5. Ran the test on my environment and verified that all 11 tests are passing now. Also verified booting a 4 GPU passthrough VM and driver load along with CUDA/UVM sanity tests running.

OK  cxl_type2.dpa_mmap_ioas_map
PASSED: 11 / 11 tests passed.
Totals: pass:11 fail:0 xfail:0 xpass:0 skip:0 error:0

6. Verified that when MMAP flag is not present for DPA region, the test is now failing and not getting skipped
TAP version 13
1..1
Starting 1 tests from 1 test cases.
 RUN           cxl_type2.dpa_mmap_ioas_map ...
vfio_cxl_type2_test.c:645:dpa_mmap_ioas_map:Expected 0 (0) != reg.flags & VFIO_REGION_INFO_FLAG_MMAP (0)
dpa_mmap_ioas_map: Test terminated by assertion
FAIL  cxl_type2.dpa_mmap_ioas_map
not ok 1 cxl_type2.dpa_mmap_ioas_map
FAILED: 0 / 1 tests passed.
Totals: pass:0 fail:1 xfail:0 xpass:0 skip:0 error:0

clsotog (approved):

clsotog:
Applied to 26.04_linux-nvidia-bos.

jamieNguyenNVIDIA:
This has been applied:
```
78a40f24863a2 (HEAD -> nvidia-bos-next, canonical-resolute/nvidia-bos-next) NVIDIA: VR: SAUCE: vfio/cxl: advertise DPA mmap and update its selftest
```

Closing MR.

PR #497, opened 2026-07-23
[26.04_linux-nvidia] drtm: Enable ARM64 Secure Launch

ianm-nv:
BugLink: https://bugs.launchpad.net/bugs/2161563

[Impact]

Enable DRTM (Dynamic Root of Trust for Measurement) on the linux-nvidia kernel,
giving Grace-based platforms hardware-measured boot-integrity attestation. Adds:

- ARM64 Secure Launch (DRTM): on `drtm=on` the kernel (as DLME) requests a
  dynamic launch, measures ACPI/DTB/cmdline/initrd into the TPM (PCR 18),
  validates the EFI systab/memory map, asserts SMMU DMA protection, and disables
  EFI runtime services.
- FIRME attestation driver (`nvidia_firme`): TSM report + measurement-register
  backends (configfs/sysfs).

[Test Plan]

On DRTM-capable Grace/GB200 (TF-A/EL3 Secure Launch service + TPM):

1. Boot `drtm=on` and confirm the launch engaged: `dmesg | grep -i slaunch`
   shows `DRTM Secure Launch detected`, the ACPI/DTB/cmdline/initrd measurement
   chain, and `SMMU DMA protection verified`; `modprobe nvidia_firme` loads.
2. Boot without `drtm=on` → boots via the standard EFI stub (feature inert when
   not requested).

Validated on Grace/GB200/GB300

[Where problems could occur]

The Secure Launch code is gated behind the `drtm=on` command-line parameter, so
without it the kernel behaves as before. The one change that applies to every
arm64 build regardless is the kernel image: it ships uncompressed
(`EFI_ZBOOT=n`), adding about 48 MB to `/boot` and the package per flavour.

[Other Info]

The trusted-KASLR patches (`ARM64_SECURE_LAUNCH_KASLR`) are included in the
series but left disabled: enabling them requires `MEMORY_HOTPLUG=n`. There will 
be followup patches to separate this dependency.

References:
- Arm DRTM Architecture (DEN0113) — dynamic-launch/DLME interface this
  implements: https://developer.arm.com/documentation/den0113/latest/
- Arm TRNG Firmware Interface (DEN0098) — SMCCC `TRNG_RND64`, used by the
  trusted-KASLR patches: https://developer.arm.com/documentation/den0098/latest/
- Trusted Firmware-A DRTM service — firmware side that performs the launch:
  https://trustedfirmware-a.readthedocs.io/en/latest/design_documents/drtm_poc.html
- ARM64 Secure Launch upstreaming (LPC 2024, TrenchBoot):
  https://lpc.events/event/18/contributions/1814/attachments/1482/3132/arm_securelaunch_lpc_2024.pdf


nirmoy:
## BaseOS Kernel Review

### Summary

Findings: no problems found

Latest watcher review: [open review](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/497/latest.html)

Generated test plan: [open test plan](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/497/test-plan.html)

Kernel deb build: **successful** ([download debs](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/497/debs/), 4 files)

Head: `1dab17fda017`

sforshee:
Codex identified one issue which looks legit in `nvidia-firme`:

> High: drivers/virt/coco/nvidia-firme/nvidia-firme.c:129-179: multi-chunk token retrieval can read/write past the 16 KiB shared buffer. offset is passed to firmware and then used as buf + offset; once offset >= FIRME_BUF_SIZE, the copy is out of bounds. Bound res.a1 to the bounce buffer and copy from the actual chunk base.

There is a check that `token_size + res.a1 > FIRME_MAX_TOKEN_SIZE` (with `offset` and `token_size` having identical values), but `FIRME_MAX_TOKEN_SIZE` is 32K so this does not prevent overrunning the buffer. I'm honestly not clear on why this code even needs to keep advancing the offset into `buf`, but I haven't looked at the spec.

A few lower-prioirty, non-blocking observations from me:

- A number of `pr_info` prints in the `nvidia-firme` driver are probably more `pr_debug` territory -- i.e. diagnostic messages useful for debugging. Hopefully they don't get printed often, otherwise they may clutter dmesg.
- Some functions repeat the same cleanups in multiple error paths; might benefit from a unified cleanup path with `goto`.

jamieNguyenNVIDIA:
> Codex identified one issue which looks legit in `nvidia-firme`:
> 
> > High: drivers/virt/coco/nvidia-firme/nvidia-firme.c:129-179: multi-chunk token retrieval can read/write past the 16 KiB shared buffer. offset is passed to firmware and then used as buf + offset; once offset >= FIRME_BUF_SIZE, the copy is out of bounds. Bound res.a1 to the bounce buffer and copy from the actual chunk base.
> 
> There is a check that `token_size + res.a1 > FIRME_MAX_TOKEN_SIZE` (with `offset` and `token_size` having identical values), but `FIRME_MAX_TOKEN_SIZE` is 32K so this does not prevent overrunning the buffer. I'm honestly not clear on why this code even needs to keep advancing the offset into `buf`, but I haven't looked at the spec.
> 
> A few lower-prioirty, non-blocking observations from me:
> 
> * A number of `pr_info` prints in the `nvidia-firme` driver are probably more `pr_debug` territory -- i.e. diagnostic messages useful for debugging. Hopefully they don't get printed often, otherwise they may clutter dmesg.
> * Some functions repeat the same cleanups in multiple error paths; might benefit from a unified cleanup path with `goto`.

Claude flagged the same issue with `nvidia-firme`.

sforshee:
> Codex identified one issue which looks legit in `nvidia-firme`:
> 
> > High: drivers/virt/coco/nvidia-firme/nvidia-firme.c:129-179: multi-chunk token retrieval can read/write past the 16 KiB shared buffer. offset is passed to firmware and then used as buf + offset; once offset >= FIRME_BUF_SIZE, the copy is out of bounds. Bound res.a1 to the bounce buffer and copy from the actual chunk base.
> 
> There is a check that `token_size + res.a1 > FIRME_MAX_TOKEN_SIZE` (with `offset` and `token_size` having identical values), but `FIRME_MAX_TOKEN_SIZE` is 32K so this does not prevent overrunning the buffer. I'm honestly not clear on why this code even needs to keep advancing the offset into `buf`, but I haven't looked at the spec.
> 
> A few lower-prioirty, non-blocking observations from me:
> 
> * A number of `pr_info` prints in the `nvidia-firme` driver are probably more `pr_debug` territory -- i.e. diagnostic messages useful for debugging. Hopefully they don't get printed often, otherwise they may clutter dmesg.
> * Some functions repeat the same cleanups in multiple error paths; might benefit from a unified cleanup path with `goto`.

From the spec, `FIRME_ATTEST_PAT_GET` is a long-running operation using a shared buffer, essentially paginating the data if the supplied buffer isn't large enough for the passed buffer size. When the loop sets `challenge_sz` to `0` this indicates to subsequent calls that we are continuing the previous incomplete read instead of starting a new one. `FIRME_ATTEST_PAT_GET` will copy to `buf_phys + offset`. Since we also pass in a buffer size this may prevent from overwriting the end of the buffer, assuming the firmware is being careful, but then we would still fail to retrieve the full token. From the spec I'm also not entirely sure that the value passed for the buffer size is correct.

I think if the loop is going to copy out the written bytes after each iteration, it should pass `0` as the offset and always copy from the start of the buffer. I've only spent a little time with the spec though, so I may be wrong. But I do think the current code will fail one way or the other if the token size ends up being larger than the allocated buffer size.

jamieNguyenNVIDIA:
Three Claude findings -- 1 and 2 are confirmed from building.  3 is a minor doc inaccuracy.  IMO: 1 is the important one to fix.

**1. Build break with `CONFIG_ARM64_SECURE_LAUNCH=n` — c8995cdc6d20 (high, confirmed by build)**

`arch/arm64/kernel/setup.c:359` adds:

```c
if (sl_dlme_region_pa && acpi_disabled)
        panic("slaunch: DRTM launch requires ACPI mode, but ACPI is disabled\n");
```

This reference is not wrapped in any `#ifdef`, but `sl_dlme_region_pa` is
declared only under `#ifdef CONFIG_ARM64_SECURE_LAUNCH` in `asm/drtm.h` (and its
definition in `sl_stub.o` is Makefile-gated on the same symbol). Every other
slaunch touch point in `setup_arch()` resolves to a no-op stub when the feature
is off — only this raw variable is exposed. The config is reachable:
`ARM64_SECURE_LAUNCH` is a plain `bool` that `depends on ARM64 && EFI`, so any
`CONFIG_EFI=n` build (or an explicit `n`) fails to compile `setup.c`. The deb
build here passes only because the shipped config sets it `=y`.

Confirmed by compiling `arch/arm64/kernel/setup.o` with `ARM64_SECURE_LAUNCH=n`:

```
arch/arm64/kernel/setup.c:359:13: error: ‘sl_dlme_region_pa’ undeclared (first use in this function)
  359 |         if (sl_dlme_region_pa && acpi_disabled)
```

Fix: use the always-defined helper —

```c
if (slaunch_active() && acpi_disabled)
```

**2. `.efistub` rename drops the zboot decompressor's initialized `.data` — a24e02364225 (medium, confirmed by build)**

The change flips `STUBCOPY_FLAGS-$(CONFIG_ARM64)` from
`--prefix-alloc-sections=.init` to `--prefix-alloc-sections=.efistub` and updates
`arch/arm64/kernel/vmlinux.lds.S`, but not
`drivers/firmware/efi/libstub/zboot.lds`, which links the same `lib.a` and still
matches only `.init.*` / `.text*` / `.data*` / `.bss*`. The old `.init` prefix
lined up with those globs; the renamed `.efistub.*` inputs match nothing there
and fall through as orphan sections (the zboot link has no `--orphan-handling`).

Confirmed on a native arm64 `CONFIG_EFI_ZBOOT=y` build of the branch. In
`arch/arm64/boot/vmlinuz.efi.elf`:

```
[ 8] .data          PROGBITS  ...f23000  size 0x0     (empty)
[ 9] .efistub.data  PROGBITS  ...f23000  size 0x14    (orphan, at _edata)
       -> __efistub_primary_display_guid (16B) + __efistub_efi_loglevel (4B)
_data == _edata == 0xf23000     __data_rawsize = 0
```

Because `__data_rawsize = _edata - _data` collapses to 0, the PE `.data` section
loads zero raw bytes, so the decompressor's initialized statics are zero-filled
at load: `efi_loglevel` becomes 0 (near-silent decompressor) and
`primary_display_guid` becomes all-zero. Reverting just the one-line prefix change
restores the correct layout (`.efistub.data` folds back into `.data`,
`__data_rawsize = 0x200`), pinning this commit as the cause. This affects every
arm64 `CONFIG_EFI_ZBOOT=y` build of the series.

The shipped flavour sets `EFI_ZBOOT=n` (7a9e0f7a01a6), but 2513dabbef68 ("Allow
SECURE_LAUNCH with EFI_ZBOOT") re-enables that combination.

Fix: extend the `zboot.lds` patterns to also match `.efistub.*` (`.text`,
`.rodata`, `.data`, `.bss`), or add `--orphan-handling=error` to the zboot link so
this cannot regress silently.

**3. KASLR Kconfig help text names the wrong entropy source — ca2f8d5b468c (low)**

The `ARM64_SECURE_LAUNCH_KASLR` help text in `arch/arm64/Kconfig` says the offset
comes from "the CPU's FEAT_RNG (RNDR)" and that it panics "if the CPU lacks
FEAT_RNG (RNDR)," but the code seeds from the SMCCC firmware TRNG
(`ARM_SMCCC_TRNG_RND64`) and never touches RNDR. Documentation-only; reword to
reference the firmware TRNG.

---

drivers/virt/coco/nvidia-firme/nvidia-firme.c
```diff
@@ -0,0 +1,387 @@
...
+ * Layout: [48 identity_digest][48 mexchange_digest][1 gpu_id][3 reserved]
+ */
+#define FIRME_BMDR_SIZE			100
+
+/*
+ * Shared buffer sizing.  ATF computes: size = (page_count + 1) * 4KB.
+ * page_count=3 → 16KB buffer, sufficient for most tokens.
+ */
+#define FIRME_BUF_PAGE_COUNT		3
+#define FIRME_MIN_SHARED_BUF_SZ	SZ_4K
```
sforshee, line 55: Is this a constant because the value is known for the implementations this driver is targeting? Based on DEN0149 v1.0 §8.12, this value should be the minimum size of a shared buffer as per the `MIN_SH_BUF_SZ` field in FIRME feature register 1. If this is not the same value, the actual size of the buffer may not match the size calculated by the firmware.

drivers/virt/coco/nvidia-firme/nvidia-firme.c
```diff
@@ -0,0 +1,387 @@
...
+			return -ENOSPC;
+		}
+
+		memcpy(&token[token_size], buf + offset, res.a1);
+		token_size += res.a1;
+		offset += res.a1;
+
+		/* Clear challenge after first call */
+		challenge_sz = 0;
+	} while (res.a2 > 0);  /* res.a2 = remaining bytes */
```
sforshee, line 179: DEN0149 Table 6.1 expressly states that for these return values:

```
status = INCOMPLETE
written size = 0
remaining size = 0
```

`FIRME_ATTEST_PAT_GET` must be invoked again to continue retrieving the token. Here the loop terminates. Completion should be based on `status == SUCCESS`, probably with some retry limits to avoid hanging indefinitely if the firmware is misbehaving.

drivers/virt/coco/nvidia-firme/nvidia-firme.c
```diff
@@ -0,0 +1,387 @@
...
+		/* res.a1 = bytes written this call, starting at offset in buf */
+		if (token_size + res.a1 > FIRME_MAX_TOKEN_SIZE) {
+			pr_err("nvidia-firme: token exceeds max size\n");
+			free_pages((unsigned long)buf, get_order(FIRME_BUF_SIZE));
+			return -ENOSPC;
+		}
+
+		memcpy(&token[token_size], buf + offset, res.a1);
+		token_size += res.a1;
+		offset += res.a1;
```
sforshee, line 175: The psuedocode in DEN0149 has a nested loop, where the inner loop terminates when `offset >= buf_size` and the outer loop copies the data out of `buf` and resets `offset` to zero. As noted previously, here offset could potentially be incremented beyond the end of the buffer. Though I think that is only likely to happen with misbehaving firmware; the more likely outcome is that once `offset` reaches the buffer size, `FIRME_ATTEST_PAT_GET` will return `INVALID_PARAMETERS`. With this implementation copying data out after every `FIRME_ATTEST_PAT_GET` call the offset should always be 0.

drivers/virt/coco/nvidia-firme/nvidia-firme.c
```diff
@@ -0,0 +1,387 @@
...
+
+/*
+ * Shared buffer sizing.  ATF computes: size = (page_count + 1) * 4KB.
+ * page_count=3 → 16KB buffer, sufficient for most tokens.
+ */
+#define FIRME_BUF_PAGE_COUNT		3
+#define FIRME_MIN_SHARED_BUF_SZ	SZ_4K
+#define FIRME_BUF_SIZE			((FIRME_BUF_PAGE_COUNT + 1) * \
+					 FIRME_MIN_SHARED_BUF_SZ)
+#define FIRME_MAX_TOKEN_SIZE		SZ_32K
```
sforshee, line 58: What is the basis for this maximum token size? DEN0149 says it should be based on `MAX_PAT_PG_CNT` from feature register 1.

drivers/virt/coco/nvidia-firme/nvidia-firme.c
```diff
@@ -0,0 +1,387 @@
...
+		    (long)res.a0 != FIRME_INCOMPLETE) {
+			pr_err("nvidia-firme: SMC failed, status=%ld (NOT_SUPPORTED=%ld, INVALID_PARAMS=%ld, DENIED=%ld)\n",
+			       (long)res.a0, FIRME_NOT_SUPPORTED,
+			       FIRME_INVALID_PARAMETERS, (long)-5);
+			free_pages((unsigned long)buf, get_order(FIRME_BUF_SIZE));
+			return -EIO;
+		}
+
+		/* res.a1 = bytes written this call, starting at offset in buf */
+		if (token_size + res.a1 > FIRME_MAX_TOKEN_SIZE) {
```
sforshee, line 167: Though it's highly unlikely, this check is not overflow-safe. I think we can safely assume that `res.a1 < FIRME_MAX_TOKEN_SIZE`, in which case this would be better:

```
if (token_size > FIRME_MAX_TOKEN_SIZE  - res.a1) {
    ...
}
```

drivers/virt/coco/nvidia-firme/nvidia-firme.c
```diff
@@ -0,0 +1,387 @@
...
+#define FIRME_BUF_SIZE			((FIRME_BUF_PAGE_COUNT + 1) * \
+					 FIRME_MIN_SHARED_BUF_SZ)
+#define FIRME_MAX_TOKEN_SIZE		SZ_32K
+#define FIRME_MAX_RETRIES		5
+
+static bool nvidia_firme_available(void)
+{
+	struct arm_smccc_res res;
+
+	arm_smccc_1_1_invoke(FIRME_SERVICE_VERSION, 0, 0, 0, 0, &res);
```
sforshee, line 65: Should this be checking the attestation service instead of the base service (`x1 = 0x4`)?

sforshee (commented):
Adding some additional feedback on the nvidia-firme driver after spending more time with DEN0149.

ianm-nv:
Converting to a draft while we sort out the issues identified.

PR #507, opened 2026-07-27
[26.04_linux-nvidia]  i2c: mediatek: add ACPI/MT8901 support and gpiolib zero-debounce fix

kmaddaraki:
Two small SAUCE patches to bring up the MediaTek MT8901 I²C host controller under ACPI on 26.04 linux-nvidia.

**gpiolib**: acpi: route acpi_dev_gpio_irq_wake_get_by() debounce through the warn-only wrapper

Mainline commit e4a77f9c85a5 ("gpiolib: acpi: Make set debounce errors non fatal") added the warn-only acpi_gpio_set_debounce_timeout() wrapper and converted two of the three ACPI call sites that program a debounce timeout. The third, in acpi_dev_gpio_irq_wake_get_by(), still returns the raw error: on firmware whose _CRS declares a GpioInt with a DebounceTimeout on an edge-triggered pin, the MediaTek EINT driver rejects set_config(PIN_CONFIG_INPUT_DEBOUNCE) with -EINVAL, the error propagates through acpi_dev_gpio_irq_get(), and i2c_hid_acpi fails probe with HID over i2c has not been provided an Int IRQ — the internal keyboard never enumerates. Convert the remaining call site to the same wrapper so all three ACPI debounce paths behave consistently: the rejection is logged and the IRQ is still delivered. Core gpiolib.c semantics are untouched.

**i2c: mediatek**: add ACPI/MT8901 support and firmware-managed clocks
Adds ACPI binding to the DT-only i2c-mt65xx driver so the MT8901 I²C controllers enumerate under NVDA0200:
- New mt8901_compat (v3 register layout, default_parent_rate=124.8 MHz) and an acpi_match_table entry NVDA0200 → mt8901_compat.
- Switch probe to device_get_match_data() and fwnode-aware device_property_read_*() so DT and ACPI share the same code path.
- device_set_node(&adap->dev, dev_fwnode(&pdev->dev)) so i2c_acpi_register_devices() walks the child HID nodes and instantiates them.
- Gate devm_clk_get() for main/dma on has_acpi_companion() — use devm_clk_get_optional() for the ACPI path (firmware keeps those clocks running and does not expose them via the Linux clk framework), and fall back to dev_comp->default_parent_rate in mtk_i2c_set_speed() when no clock handle is available.

DT behaviour is unchanged: both device_get_match_data() and device_property_read_*() fall through to the existing OF helpers for DT-described nodes, and the DT clock path is preserved.

LP: https://bugs.launchpad.net/ubuntu/+source/linux-nvidia-7.0/+bug/2163407

### DIFF
--- a/drivers/i2c/busses/i2c-mt65xx.c
+++ b/drivers/i2c/busses/i2c-mt65xx.c
@@ -4,6 +4,7 @@
  * Author: Xudong Chen <xudong.chen@mediatek.com>
  */
 
+#include <linux/acpi.h>
 #include <linux/clk.h>
 #include <linux/completion.h>
 #include <linux/delay.h>
@@ -269,6 +270,14 @@ struct mtk_i2c_compatible {
 	unsigned char ltiming_adjust: 1;
 	unsigned char apdma_sync: 1;
 	unsigned char max_dma_support;
+	/*
+	 * Parent clock rate to use when the platform does not expose a
+	 * Linux clock provider for the I2C source clock (e.g. firmware-
+	 * managed clocks under ACPI). Left zero on SoCs whose firmware
+	 * exposes clocks through the clk framework; in that case the
+	 * driver continues to call clk_get_rate() on a mandatory clock.
+	 */
+	unsigned int default_parent_rate;
 };
 
 struct mtk_i2c_ac_timing {
@@ -522,6 +531,26 @@ static const struct mtk_i2c_compatible mt8192_compat = {
 	.max_dma_support = 36,
 };
 
+static const struct mtk_i2c_compatible mt8901_compat = {
+	.regs = mt_i2c_regs_v3,
+	.pmic_i2c = 0,
+	.dcm = 0,
+	.auto_restart = 1,
+	.aux_len_reg = 1,
+	.timing_adjust = 1,
+	.dma_sync = 0,
+	.ltiming_adjust = 1,
+	.apdma_sync = 1,
+	.max_dma_support = 40,
+	/*
+	 * MT8901 I2C on firmware-managed platforms: the IP is clocked off
+	 * a 124.8 MHz parent that is not exposed via clk_get_rate(). This
+	 * value mirrors the rate the SoC input clock is programmed to by
+	 * the boot loader.
+	 */
+	.default_parent_rate = 124800000,
+};
+
 static const struct of_device_id mtk_i2c_of_match[] = {
 	{ .compatible = "mediatek,mt2712-i2c", .data = &mt2712_compat },
 	{ .compatible = "mediatek,mt6577-i2c", .data = &mt6577_compat },
@@ -538,6 +567,14 @@ static const struct of_device_id mtk_i2c_of_match[] = {
 	{}
 };
 MODULE_DEVICE_TABLE(of, mtk_i2c_of_match);
+
+#ifdef CONFIG_ACPI
+static const struct acpi_device_id mtk_i2c_acpi_match[] = {
+	{ "NVDA0200", (kernel_ulong_t)&mt8901_compat },
+	{ }
+};
+MODULE_DEVICE_TABLE(acpi, mtk_i2c_acpi_match);
+#endif
 
 static u16 mtk_i2c_readw(struct mtk_i2c *i2c, enum I2C_REGS_OFFSET reg)
 {
@@ -1347,24 +1384,31 @@ static const struct i2c_algorithm mtk_i2c_algorithm = {
 	.functionality = mtk_i2c_functionality,
 };
 
-static int mtk_i2c_parse_dt(struct device_node *np, struct mtk_i2c *i2c)
+static int mtk_i2c_parse_fw(struct mtk_i2c *i2c)
 {
 	int ret;
 
-	ret = of_property_read_u32(np, "clock-frequency", &i2c->speed_hz);
+	ret = device_property_read_u32(i2c->dev, "clock-frequency",
+				       &i2c->speed_hz);
 	if (ret < 0)
 		i2c->speed_hz = I2C_MAX_STANDARD_MODE_FREQ;
 
-	ret = of_property_read_u32(np, "clock-div", &i2c->clk_src_div);
-	if (ret < 0)
-		return ret;
+	ret = device_property_read_u32(i2c->dev, "clock-div",
+				       &i2c->clk_src_div);
+	if (ret < 0) {
+		if (has_acpi_companion(i2c->dev))
+			i2c->clk_src_div = 1;
+		else
+			return ret;
+	}
 
 	if (i2c->clk_src_div == 0)
 		return -EINVAL;
 
-	i2c->have_pmic = of_property_read_bool(np, "mediatek,have-pmic");
-	i2c->use_push_pull =
-		of_property_read_bool(np, "mediatek,use-push-pull");
+	i2c->have_pmic = device_property_read_bool(i2c->dev,
+						   "mediatek,have-pmic");
+	i2c->use_push_pull = device_property_read_bool(i2c->dev,
+						       "mediatek,use-push-pull");
 
 	i2c_parse_fw_timings(i2c->dev, &i2c->timing_info, true);
 
@@ -1376,6 +1420,7 @@ static int mtk_i2c_probe(struct platform_device *pdev)
 	int ret = 0;
 	struct mtk_i2c *i2c;
 	int i, irq, speed_clk;
+	unsigned int parent_rate;
 
 	i2c = devm_kzalloc(&pdev->dev, sizeof(*i2c), GFP_KERNEL);
 	if (!i2c)
@@ -1395,10 +1440,12 @@ static int mtk_i2c_probe(struct platform_device *pdev)
 
 	init_completion(&i2c->msg_complete);
 
-	i2c->dev_comp = of_device_get_match_data(&pdev->dev);
-	i2c->adap.dev.of_node = pdev->dev.of_node;
+	i2c->dev_comp = device_get_match_data(&pdev->dev);
+	if (!i2c->dev_comp)
+		return -ENODEV;
 	i2c->dev = &pdev->dev;
 	i2c->adap.dev.parent = &pdev->dev;
+	device_set_node(&i2c->adap.dev, dev_fwnode(&pdev->dev));
 	i2c->adap.owner = THIS_MODULE;
 	i2c->adap.algo = &mtk_i2c_algorithm;
 	i2c->adap.quirks = i2c->dev_comp->quirks;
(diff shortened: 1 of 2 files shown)


nirmoy:
## BaseOS Kernel Review

### Summary

Findings: no problems found

Latest watcher review: [open review](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/507/latest.html)

Generated test plan: [open test plan](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/507/test-plan.html)

Kernel deb build: **successful** ([download debs](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/507/debs/), 4 files)

Head: `f2581685e05e`

jamieNguyenNVIDIA:
Are these patches that you intend to send to the mailing list? If so, I'd suggest doing that first and then referencing the mailing list URL in this PR:
```
(backported from <lore URL>)
```

Secondly, I suggest targeting this PR at the following branches instead of `24.04_linux-nvidia-7.0-next`:

- 26.04_linux-nvidia-bos
- 26.04_linux-nvidia

Canonical helps ensure that the 24.04 kernel source is kept in sync with the 26.04 versions, so no need to target PRs at the 24.04+7.0 branches.

drivers/gpio/gpiolib.c
```diff
@@ -2783,6 +2783,15 @@ int gpio_set_debounce_timeout(struct gpio_desc *desc, unsigned int debounce)
...
 	int ret;
 
+	/*
+	 * A debounce timeout of 0 means "no debounce" - there is nothing to
+	 * configure, and some GPIO controllers reject set_config(DEBOUNCE, 0)
+	 * with -EINVAL on edge-triggered interrupts. Return success without
+	 * touching the hardware.
+	 */
+	if (!debounce)
+		return 0;
```
nirmoy: **Medium:** This changes the core API from “zero disables debounce” to “zero is ignored.” If firmware or an earlier request enabled debounce, returning here leaves hardware filtering and `desc->debounce_period_us` stale, and skips the line-state notification. Please scope the ACPI workaround to the affected call path or make the MediaTek EINT implementation handle zero by disabling debounce.

drivers/i2c/busses/i2c-mt65xx.c
```diff
@@ -1347,24 +1378,27 @@ static const struct i2c_algorithm mtk_i2c_algorithm = {
...
+				       &i2c->speed_hz);
 	if (ret < 0)
 		i2c->speed_hz = I2C_MAX_STANDARD_MODE_FREQ;
 
-	ret = of_property_read_u32(np, "clock-div", &i2c->clk_src_div);
+	ret = device_property_read_u32(i2c->dev, "clock-div",
+				       &i2c->clk_src_div);
 	if (ret < 0)
-		return ret;
+		i2c->clk_src_div = 1;	/* ACPI doesn't supply this */
```
nirmoy: **Low:** This fallback also applies to DT and to malformed properties. `clock-div` is required by `i2c-mt65xx.yaml`, and the previous DT path returned the read error; silently using 1 changes DT behavior and may calculate the wrong SCL timing. Please default only for an absent ACPI property and preserve the DT/malformed-property error.

nirmoy (commented):
Codex found these two issues.

clsotog:
Got the same findings as Nirmoy.

jamieNguyenNVIDIA:
@kmaddaraki: Are these going to get sent upstream?

kmaddaraki:
> @kmaddaraki: Are these going to get sent upstream?

Mediatek has plan to upstream these changes and it will happen around Oct 2026

jamieNguyenNVIDIA:
# Review: PR 507 — additional notes

The two findings already posted stand (core debounce semantics on
`gpiolib.c`, the `clock-div` fallback on `i2c-mt65xx.c`). The following is
not covered by them.

## A four-line fix already exists for the gpiolib problem

`e4a77f9c85a5` ("gpiolib: acpi: Make set debounce errors non fatal") is
already in this branch. It added a warn-only wrapper and converted two of
the three `gpio_set_debounce_timeout()` call sites:

```c
/* drivers/gpio/gpiolib-acpi-core.c */
static void acpi_gpio_set_debounce_timeout(struct gpio_desc *desc,
                                           unsigned int acpi_debounce)
{
        ...
        ret = gpio_set_debounce_timeout(desc, acpi_debounce);
        if (ret)
                gpiod_warn(desc, "Failed to set debounce-timeout %u: %d\n",
                           acpi_debounce, ret);
}
```

Two consequences.

**The commit message is out of date.** It says the error "propagated
through `gpiod_get_index()` and `acpi_dev_gpio_irq_get()`". The
`gpiod_get_index()` path — `gpiod_find_and_request()` →
`gpiod_fwnode_lookup()` → `gpiod_find_by_fwnode()` → `acpi_find_gpio()` —
only warns now.

**One call site still fails**, `acpi_dev_gpio_irq_wake_get_by()`:

```c
ret = gpio_set_debounce_timeout(desc, info.debounce * 10);
if (ret)
        return ret;
```

Converting that single call to `acpi_gpio_set_debounce_timeout()` is the
"scope it to the affected call path" fix suggested above, needs no core
change, and finishes what `e4a77f9c85a5` started. That commit also records
the upstream position on this failure mode:

> But `gpio_set_debounce_timeout()` failing is a somewhat normal
> occurrence, since not all debounce values are supported on all
> GPIO/pinctrl chips.

## The zero-only guard is incomplete

`mtk_eint_set_debounce()` rejects the request without inspecting the
requested value at all:

```c
static unsigned int mtk_eint_can_en_debounce(struct mtk_eint *eint,
                                             unsigned int eint_num)
{
        ...
        if (eint->pins[eint_num].debounce && sens != MTK_EINT_EDGE_SENSITIVE)
                return 1;
        else
                return 0;
}
```

An MT8901 `_CRS` declaring any non-zero `DebounceTimeout` on an
edge-triggered EINT would still return `-EINVAL` and still fail
`acpi_dev_gpio_irq_get()`. Fixing the call site covers that case; the zero
guard does not.

## Nits

- `default_parent_rate = 124800000` has no stated source. With
  `clk_src_div` falling back to 1, it is the sole input to the ACPI timing
  calculation — a comment naming the source would make it reviewable.
- `i2c->adap.dev.of_node = pdev->dev.of_node;` in `mtk_i2c_probe()` is
  dead; `device_set_node()` sets `of_node` from the fwnode a few lines
  later.
- checkpatch: `CHECK: Alignment should match open parenthesis` on the
  `mtk_i2c_set_speed()` call. A local holding the parent rate would read
  better than the nested conditional.

## Two corrections to the earlier review

- `desc->debounce_period_us` is not left stale by this patch.
  `gpio_set_debounce_timeout()` never wrote that field on any path; it is
  written only by `gpiolib-cdev.c`.
- The skipped `gpiod_line_state_notify()` is arguably correct now, since no
  configuration changed. Previously it fired even when the call was a
  no-op.

---

Both changed objects build clean on arm64 defconfig with `CONFIG_ACPI=y`
and `=n`.

drivers/i2c/busses/i2c-mt65xx.c
kmaddaraki: Fixed in v2 — the fallback is now gated on has_acpi_companion():
  
   if (ret < 0) {
       if (has_acpi_companion(i2c->dev))
           i2c->clk_src_div = 1;
       else
           return ret;
   }
  
DT retains the previous behavior of returning the property-read error, so a malformed or missing clock-div still fails the probe as i2c-mt65xx.yaml requires.

drivers/gpio/gpiolib.c
kmaddaraki: Fixed in v2 by taking your first option — the workaround is now scoped to the affected ACPI call path. The gpiolib.c change is dropped entirely; instead the remaining hard-failing call site in acpi_dev_gpio_irq_wake_get_by() is converted to the warn-only acpi_gpio_set_debounce_timeout() wrapper that e4a77f9c85a5 already applied to the other two ACPI call sites. Core semantics ("zero disables debounce") are untouched for all non-ACPI callers.

kmaddaraki:
@jamieNguyenNVIDIA sorry for the delay. Testing took more time than anticipated.

v2 implements exactly this — thanks for the pointer to e4a77f9c85a5. The gpiolib commit is now a conversion of the third call site to the existing wrapper (net 4 lines in gpiolib-acpi-core.c, gpiolib.c untouched), and it also covers the nonzero-debounce failure case the previous zero-only guard missed.
The i2c-mt65xx nits are all folded in: dead adap.dev.of_node assignment removed, comment added for default_parent_rate = 124800000, and the mtk_i2c_set_speed() alignment fixed via a parent_rate local.

jamieNguyenNVIDIA:
> @jamieNguyenNVIDIA sorry for the delay. Testing took more time than anticipated.
> 
> v2 implements exactly this — thanks for the pointer to [e4a77f9](https://github.com/NVIDIA/NV-Kernels/commit/e4a77f9c85a528b3289c1d9570d6d73a7b5f847b). The gpiolib commit is now a conversion of the third call site to the existing wrapper (net 4 lines in gpiolib-acpi-core.c, gpiolib.c untouched), and it also covers the nonzero-debounce failure case the previous zero-only guard missed. The i2c-mt65xx nits are all folded in: dead adap.dev.of_node assignment removed, comment added for default_parent_rate = 124800000, and the mtk_i2c_set_speed() alignment fixed via a parent_rate local.

@kmaddaraki: No worries!  Does https://github.com/NVIDIA/NV-Kernels/pull/507/changes/c439da8ba5af052d09d1a5dcac1debc3a8296307 need to carry the following fixes tag?
```
Fixes: 8dcb7a15a585 ("gpiolib: acpi: Take into account debounce settings")
```

Other than that:

nvmochs:
+1 on the fixes tag.

I confirmed the issues reported here and in PR 518 have been resolved.

clsotog:
no further issues from me. 

kmaddaraki:
@jamieNguyenNVIDIA  Yes, agreed — added Fixes: 8dcb7a15a585 ("gpiolib: acpi: Take into account debounce settings")

jamieNguyenNVIDIA:
Hi @kmaddaraki, looks like this one needs a rebase.

jamieNguyenNVIDIA:

jamieNguyenNVIDIA:
Applied:
- https://git.launchpad.net/~canonical-kernel/ubuntu/+source/linux-nvidia/+git/resolute/commit/?h=main-next&id=1e2a75fb7d0f925eab94fd32e62fe6147d954ea1
- https://git.launchpad.net/~canonical-kernel/ubuntu/+source/linux-nvidia/+git/resolute/commit/?h=main-next&id=c8ca6b82aeb7c0bd94f1b76e9c1880255114faed

PR #508, opened 2026-07-27
[linux-nvidia-6.18-next] July patch refresh

ltrager:
# Overview

Apply the latest versions of the following LKML patch series:

* [MPAM-Fb firmware support](https://lore.kernel.org/r/20260804100627.3217149-1-andre.przywara@arm.com)
* [MPAM ABMC](https://lore.kernel.org/all/20260707162440.12132-1-ben.horgan@arm.com/)
* [PCI error recovery](https://lore.kernel.org/all/20260720192505.2957-1-alifm@linux.ibm.com/)
* [Olympus PMU](https://lore.kernel.org/all/20260608234135.1856911-1-bwicaksono@nvidia.com/)

Doing this now as I am working on testing and promoting `linux-nvidia-6.18-next` into `linux-nvidia-6.18`. For purposes of this review I reverted the old versions before applying the new versions. When landing I will drop the old version and replace it with the new version to keep the git history clean.

## Downstream patches

Several downstream patches depend on code changed by the updated series. They were temporarily reverted and replayed after the new versions were applied.

  ### Reapplied without source changes

  - 53d926b8faef4 → 9b15e70aec4ce — arm_mpam: Add debugfs entries to show the MSC/RIS the driver discovered
  - 7bb1614cd5149 → e43dfd84f7dd0 — arm_mpam: Add force-disable debugfs trigger
  - f4d2dbac44069 → ce5aa984e99ad — arm_mpam: Fix monitor capability and MBM assign for MB events

  ### Updated for the new series

  - 542141fa7c443 → 711729a7a991a — arm_mpam: Expose the number of NRDY retries in debugfs
      - Refreshed NRDY retry accounting for the MPAM-Fb v4 propagated-error read paths.
      - Preserves the existing debugfs nrdy_retry_count interface.

  - 6e399ddebcd90 → 53c9fb9ec1341 — arm_mpam: Initialize mbwu_state values to 0
      - Retargeted the initialization to the revised v4 __ris_msmon_read() path.
      - Initializes both the monitor-read argument and value before use.
      - Updates the Fixes: reference to the corresponding MPAM-Fb v4 commit.

  - 381e35c99ef32 → 1a9142d9cf655 — arm_mpam: Include all MSC components during domain setup
      - Refreshed multi-MSC domain handling for ABMC v5.
      - Preserves v5 per-counter assignments and targeted counter resets.
      - Derives num_mbm_cntrs from the MPAM monitoring class selected for each resource.
      - Retains downstream MBA-backed monitoring and MB_HLIM initialization.

  - b053900fcf811 → 050def495708a — Fix mbm_L3_assign and mon_local_bytes
      - Preserves the Grace local-bandwidth counter.
      - Updates ABMC v5 resource synchronization to use the presence of assigned MBM counters.

  - 097383162f4a6 → b7d169ac2c0c2 — arm_mpam: resctrl: Add kunit test for ABMC/CDP interactions
      - Rewritten for ABMC v5 class-based counter selection.
      - Tests CDP counter splitting.
      - Tests that mbm_event remains enabled when CDP leaves zero assignable counters.
      - Tests that resource synchronization is deferred until counters are assigned.

  ### Fixes integrated into MPAM-Fb v4

  The following follow-up fixes no longer require separate downstream patches:

  - 66f84b172a810 — arm_mpam: Increase props size to account for new property
      - The corrected ACPI property-array size is integrated into MPAM-Fb v4.

  - 6feb6eccd8048 — arm_mpam: Mask token to max width
      - The MPAM-Fb token masking is integrated into v4.

  No downstream patches were dropped.

  The hashes on the right identify commits in this review branch. They will change when the temporary revert history is removed for landing.
  
# Testing

Built and run on a Vera system. Successfully reran [mpam_tests](https://github.com/ltrager/NV-Kernels/tree/mpam_tests)


drivers/resctrl/mpam_internal.h
```diff
@@ -545,8 +555,8 @@ int mpam_fb_send_read_request(struct mpam_msc *msc, u16 reg, u32 *result);
 int mpam_fb_send_write_request(struct mpam_msc *msc, u16 reg, u32 value);
 int mpam_fb_get_protocol_version(struct mpam_msc *msc);
 
-#define PCC_TYPE3_MSG_PAYLOAD_OFS	0x10
-#define MPAM_FB_MAX_MSG_SIZE	(PCC_TYPE3_MSG_PAYLOAD_OFS + 4 * sizeof(u32))
+#define MPAM_FB_PROT_HEADER_LEN	sizeof(u32)
+#define MPAM_FB_MAX_MSG_SIZE	(MPAM_FB_PROT_HEADER_LEN + 4 * sizeof(u32))
```
nirmoy: [High] `MPAM_FB_MAX_MSG_SIZE` is 20 bytes (the 4-byte protocol header plus the 16-byte write payload), but `mpam_fb_build_write_message()` places that payload after `sizeof(struct acpi_pcct_ext_pcc_shared_memory)` and writes through byte 31 of the PCC shared-memory region. The probe check can therefore accept a 20–31 byte region that the write path overruns. Should the capacity check use the full PCC header plus the largest payload while leaving the protocol `length` field at 20?

This comes directly from MPAM-Fb v4 [8/10](https://lore.kernel.org/r/20260723155454.1760823-9-andre.przywara@arm.com) together with the validation in [10/10](https://lore.kernel.org/r/20260723155454.1760823-11-andre.przywara@arm.com); it was not introduced by the downstream pickup. I do not consider it a blocker for this refresh, but if confirmed it is worth replying on the mailing-list thread so it can be fixed in the next revision.

drivers/resctrl/mpam_devices.c
```diff
@@ -60,55 +61,59 @@ struct srcu_struct mpam_srcu;
...
 		kfree(cur);
-		mutex_unlock(&pcc_chan_list_lock);
 		return ERR_PTR(err);
 	}
 
-	mutex_init(&cur->pcc_chan_lock);
+	/* Timeout based on the "nominal latency" from the PCC ACPI table. */
+	cur->pcc_cl.tx_tout = cur->pcc_chan->latency * 5;
+
+	ret = devm_mutex_init(dev, &cur->pcc_chan_lock);
```
nirmoy: [High] `cur` and its PCC channel are refcounted and may be shared by MSCs owned by different devices, but this makes the shared mutex devres-managed by the first `dev`. If that device unbinds while another MSC still holds a reference, its devres teardown can destroy the mutex before the shared object is released, leaving the remaining user to lock a destroyed mutex. A plain `mutex_init()` tied to `cur`'s lifetime appears safer; the initialization failure path should also release the requested channel and allocation.

This code comes directly from MPAM-Fb v4 [10/10](https://lore.kernel.org/r/20260723155454.1760823-11-andre.przywara@arm.com), not from the downstream pickup. It is not a blocker for this refresh, but if the lifetime concern is valid it is worth replying on that mailing-list thread.

drivers/resctrl/mpam_devices.c
```diff
@@ -1983,61 +1981,54 @@ static int mpam_save_mbwu_state(void *arg)
...
-			cfg->partid = FIELD_GET(MSMON_CFG_x_FLT_PARTID, cur_flt);
-			mbwu_state->correction += val;
-			mbwu_state->enabled = FIELD_GET(MSMON_CFG_x_CTL_EN, cur_ctl);
-		}
-		mpam_mon_sel_unlock(msc);
+		cfg->mon = i;
+		cfg->pmg = FIELD_GET(MSMON_CFG_x_FLT_PMG, cur_flt);
+		cfg->match_pmg = FIELD_GET(MSMON_CFG_x_CTL_MATCH_PMG, cur_ctl);
+		cfg->partid = FIELD_GET(MSMON_CFG_x_FLT_PARTID, cur_flt);
+		mbwu_state->correction += val;
```
nirmoy, line 2117: [Medium] `mpam_msc_read_mbwu_l()` now returns success with `MSMON___L_NRDY` when it cannot obtain a stable long-counter value, but this path unconditionally adds `val` to `mbwu_state->correction`. The earlier guard that skipped state accounting for the NRDY sentinel is gone, so an unstable read can add the sentinel bit to the correction and corrupt subsequent accounting. Should this return an error or explicitly skip the save/reset when `val == MSMON___L_NRDY`?

The behavior comes directly from MPAM-Fb v4 [3/10](https://lore.kernel.org/r/20260723155454.1760823-4-andre.przywara@arm.com); it was not introduced by the downstream pickup. I do not consider it a blocker for this refresh, but if confirmed it is worth reporting on the mailing-list thread.

drivers/pci/slot.c
```diff
@@ -63,6 +63,23 @@ static ssize_t cur_speed_read_file(struct pci_slot *slot, char *buf)
...
+	if (slot->per_func_slot)
+		return dev->devfn == slot->number;
+
+	return slot->number == PCI_SLOT_ALL_DEVICES ||
+		PCI_SLOT(dev->devfn) == slot->number;
+}
+
+static bool pci_slot_enabled_per_func(void)
+{
+	if (IS_ENABLED(CONFIG_S390))
```
nirmoy, line 77: [Medium] This enables per-function matching for every slot created in a `CONFIG_S390` build. `pci_dev_matches_slot()` then treats `slot->number` as a full `devfn`, while callers that still pass a conventional `PCI_SLOT(devfn)` value provide only the device number. Such callers would no longer match functions 1–7. Can this mode be selected explicitly by the s390 caller that supplies a full `devfn`, rather than globally for every slot on the architecture?

This is directly from PCI v22 [1/4](https://lore.kernel.org/r/20260720192505.2957-2-alifm@linux.ibm.com), not a downstream pickup change. It is not a blocker for this refresh, but if the concern applies to any s390 slot creator it is worth replying on the mailing-list thread.

nirmoy (approved):
Codex verified that the High/Medium findings called out inline are present in the corresponding mailing-list patches and were not introduced by the downstream pickup. I do not consider them blockers for this downstream refresh; if confirmed, they should be followed up on the linked mailing-list threads so upstream can address them.

sforshee:
A couple of nits on the provenance. Backport notes are typically placed below the `(backported from ...)`, line, not above it. And I noticed at least one commit which had backport notes but still said `cherry picked from`.

I got similar code review findings to Nirmoy. It seems like the fixes for the size check and to check the `mpam_msc_read_mbwu_l()` return value would be pretty trivial, so I think it would be good to add those. As noted on the upstream thread, he mutex lifetime one only really becomes an issue when `CONFIG_DEBUG_MUTEXES` is on as `__devm_mutex_init()` is a nop otherwise, but it still looks pretty straightforward to fix.

nirmoy:
## BaseOS Kernel Review

### Summary

No problems found.

Findings: no problems found

Latest watcher review: [open review](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/508/latest.html)

Generated test plan: [open test plan](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/508/test-plan.html)

Kernel deb build: **successful** ([download debs](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/508/debs/), 4 files)

Head: `d37ce02f1cd4`

drivers/pci/slot.c
ltrager: Sashiko bot replied with the same thing. Since this only effects S390X I'm leaving it as is.

drivers/resctrl/mpam_devices.c
ltrager: Fixed in V5 patchset

drivers/resctrl/mpam_internal.h
ltrager: Fixed in separate commit and [reported upstream](https://lore.kernel.org/all/ec478b24-0a30-4b63-b191-08d05f606bad@trager.us/#t).

drivers/resctrl/mpam_devices.c
ltrager: Fixed as a separate commit and [reported upstream](https://lore.kernel.org/all/2228a2be-96b1-4c8a-9786-6ba07fd302c0@trager.us/).

ltrager:
Thanks for the reviews. V5 of the MPAM-Fb PPC patchset just came out which fixed a few of the issues identified which is now included. I reported and fixed in seperate commits the NRDY check and shared memory size check, also fixed where the backport notes are.

Rebased with linux-nvidia-6.18-next and retested on Vera

drivers/resctrl/mpam_devices.c
```diff
@@ -2427,7 +2439,10 @@ static struct mpam_msc *do_mpam_msc_drv_probe(struct platform_device *pdev)
 	if (err)
 		return ERR_PTR(err);
 
-	mpam_mon_sel_lock_init(msc);
+	err = mpam_mon_sel_lock_init(dev, msc);
```
sforshee: •   High — uninitialized PCC mutex: This initializes the interface-specific lock before setting msc->iface. PCC devices initialize only the MMIO spinlock, then later use an uninitialized mon_sel_mutex, risking corruption/crashes under contention.

drivers/resctrl/mpam_devices.c
```diff
@@ -60,55 +61,57 @@ struct srcu_struct mpam_srcu;
...
 	if (IS_ERR(cur->pcc_chan)) {
 		long err = PTR_ERR(cur->pcc_chan);
 
 		kfree(cur);
-		mutex_unlock(&pcc_chan_list_lock);
 		return ERR_PTR(err);
 	}
 
+	/* Timeout based on the "nominal latency" from the PCC ACPI table. */
+	cur->pcc_cl.tx_tout = cur->pcc_chan->latency * 5;
```
sforshee: Medium — timeout unit mismatch: This assigns ACPI’s microsecond PCC latency directly to tx_tout, which expects milliseconds. Failure paths can wait roughly 1000× longer than intended.

Reference: https://uefi.org/sites/default/files/resources/ACPI_Spec_6.6.pdf §14.1.3

In the spec this is nominal latency, so it makes sense that it would be larger. But 5000x nominal seems extreme, and given that there's no comment and a scale factor of 5 it seems more like a mistake. I don't see this as a blocker here, but could be good feedback for the upstream patch.

nirmoy:
Thanks @ltrager My Ack remains

drivers/resctrl/mpam_devices.c
```diff
@@ -1523,63 +1544,63 @@ static void __ris_msmon_read(void *arg)
...
 	case mpam_feat_msmon_mbwu_44counter:
 	case mpam_feat_msmon_mbwu_63counter:
 		if (m->type != mpam_feat_msmon_mbwu_31counter) {
-			ret = mpam_msc_read_mbwu_l(msc, &now);
-			if (ret)
-				goto out_unlock;
+			m->err = mpam_msc_read_mbwu_l(msc, &now);
+			if (m->err)
+				return;
+			nrdy = now & MSMON___L_NRDY;
```
sforshee, line 1600: Low — retry counter undercounts: This doesn’t increment nrdy_retry_count for 44/63-bit MBWU counters.

Not a blocker for this PR, but might be good feedback upstream.

sforshee (commented):
Codex found a few more legitimate-looking issues this time. A couple are minor, but one looks potentially more significant.

In addition I noticed that the bodies in some of the commit messages you authored are not wrapped as is customary for kernel commit messages.

drivers/resctrl/mpam_devices.c
ltrager: Applied patch and [reported on the list.](https://lore.kernel.org/all/b9d57d9f-61c7-4e62-a120-7631bdb7069b@trager.us/)

drivers/resctrl/mpam_devices.c
ltrager: Applied patch and [reported upstream.](https://lore.kernel.org/all/c3d43f56-b875-4bbd-b249-1a5f33cd54ed@trager.us/)

drivers/resctrl/mpam_devices.c
ltrager: This is actually from https://git.kernel.org/pub/scm/linux/kernel/git/morse/linux.git and not on the list.  Since its low I'm leaving it as is. I may have to reach out to James directly.

ltrager:
Thanks for the review. v6 just came out so updated the patchset and fixed the timeout and setting mon_sel lock before init.

ltrager:
[V7 of the MPAM-Fb firmware support patch series](https://lore.kernel.org/all/20260731170324.2927758-1-andre.przywara@arm.com/) was released earlier today which contains fixes requested here. Updated and dropped my patches. Additionally added [arm_mpam: Apply T241-MPAM-6 to 63-bit counters](https://lore.kernel.org/r/20260727191326.2202616-1-sdonthineni@nvidia.com) which @fyu1 just posted PRs for the 7.0 kernels.

drivers/resctrl/mpam_devices.c
```diff
@@ -1983,61 +2050,57 @@ static int mpam_save_mbwu_state(void *arg)
...
-			cfg->match_pmg = FIELD_GET(MSMON_CFG_x_CTL_MATCH_PMG, cur_ctl);
-			cfg->partid = FIELD_GET(MSMON_CFG_x_FLT_PARTID, cur_flt);
-			mbwu_state->correction += val;
-			mbwu_state->enabled = FIELD_GET(MSMON_CFG_x_CTL_EN, cur_ctl);
-		}
-		mpam_mon_sel_unlock(msc);
-	}
-	return 0;
+		if (val & MSMON___L_NRDY)
+			continue;
```
sforshee, line 2108:
```
#define MSMON___L_NRDY          BIT(63)
```

On the 31-bit path, val came from a u32, and NRDY is MSMON___NRDY at bit 31. Bit 63 can never be set, so the invalid reading—including the bit-31 flag—is added to correction. Select the NRDY mask according to whether the long counter was used.

drivers/resctrl/mpam_devices.c
```diff
@@ -1983,61 +2050,57 @@ static int mpam_save_mbwu_state(void *arg)
...
 
-out_unlock:
-	mpam_mon_sel_unlock(msc);
+		cfg->mon = i;
+		cfg->pmg = FIELD_GET(MSMON_CFG_x_FLT_PMG, cur_flt);
+		cfg->match_pmg = FIELD_GET(MSMON_CFG_x_CTL_MATCH_PMG, cur_ctl);
+		cfg->partid = FIELD_GET(MSMON_CFG_x_FLT_PARTID, cur_flt);
+		mbwu_state->correction += val;
```
sforshee, line 2117: Shouldn't the same scaling that this series adds for `__ris_msmon_read()` for `mpam_has_quirk(T241_MBW_COUNTER_SCALE_64, msc)` be done here?

sforshee (changes requested):
A couple of additional codex findings notated inline.

Also, `arm_mpam: Apply T241-MPAM-6 to 63-bit counters` should be tagged as SAUCE.

drivers/resctrl/mpam_devices.c
ltrager: Resolved in 052e0e4ac895c55bf7eea6e822f6fe56c75b0b37.

drivers/resctrl/mpam_devices.c
ltrager: Resolved in 8f0340e53f83b6f21ca87750d15e4609f4b9fb3f

ltrager:
@sforshee thanks for the review. Resolved both issues with a SAUCE patch. [Ben Horgan actually has a series up to fix these issues along with a few others.](https://lore.kernel.org/all/20260710115546.29644-1-ben.horgan@arm.com/) That conflicts with [MPAM-Fb firmware support](https://lore.kernel.org/all/20260731170324.2927758-1-andre.przywara@arm.com/) which is also up for review. Once both land and upstream resolves conflicts I will add Ben's series.

sforshee (approved):
Looks good overall.

```
```

There is one thing a little confusing about the provenance still that I wanted to ask about. Some patches have something like this:

```
(cherry picked from commit https://github.com/NVIDIA/NV-Kernels/commit/568443a0b4e72cbc2f159629ea31ab9420cb3fae https://github.com/NVIDIA/NV-Kernels/tree/24.04_linux-nvidia-6.17-next)
Signed-off-by: Lee Trager <ltrager@nvidia.com>
Acked-by: Jamie Nguyen <jamien@nvidia.com>
Acked-by: Fenghua Yu <fenghuay@nvidia.com>
```

This makes it look like that the specific commit in this PR for 6.18 has already been acked by Jamie, Fenghua, and me, even though it is up for review in this PR which we have not previously acked. I think for this particular example at least you had reverted the changes and then cherry picked the same patch again -- is this right? Is it the same for the other commits where I see the same thing?

ltrager:
Updated to v8 of the [arm_mpam: Add MPAM-Fb firmware support](https://lore.kernel.org/all/20260804100627.3217149-1-andre.przywara@arm.com/#r) series.

@sforshee - In order to update the patch series some patches applied on top had to be reverted then reapplied. This allowed the old series revert, new series apply to all happen cleanly. The provenance remained because the patch didn't change.

My plan on when applying these patches is to apply the new series in place which should clean up the git history a bit.

ltrager:
@sforshee updated to add fix discussed over chat.

sforshee:
@ltrager Codex finds a race and a leak in the new fix:

  1. High — the PCC lifetime fix does not guarantee an SRCU grace period.
     mpam_msc_drv_remove() (https://github.com/NVIDIA/NV-Kernels/blob/5eb98db5288f7985985964e0b1af49367b6a04ea/drivers/resctrl/mpam_devices.c#L2457-L2469) now releases the channel after mpam_free_garbage(),
     but mpam_free_garbage() (https://github.com/NVIDIA/NV-Kernels/blob/5eb98db5288f7985985964e0b1af49367b6a04ea/drivers/resctrl/mpam_devices.c#L255-L270) returns without synchronizing when its global garbage
     list is empty.

     Another caller can drain the newly queued MSC and still be waiting in synchronize_srcu() while this remove call finds an empty list and releases the channel. An existing reader can then access the freed
     channel. Make channel release part of the deferred garbage cleanup, or perform an unconditional grace period before releasing it.

     This would also address the related leak in mpam_disable() (https://github.com/NVIDIA/NV-Kernels/blob/5eb98db5288f7985985964e0b1af49367b6a04ea/drivers/resctrl/mpam_devices.c#L3517-L3522), which destroys
     PCC-backed MSCs without dropping their channel references. The incomplete fix is 5eb98db5288f (https://github.com/NVIDIA/NV-Kernels/commit/5eb98db5288f7985985964e0b1af49367b6a04ea).

sforshee:
Looks good!

```
```

nirmoy:

PR #514, opened 2026-07-29
[26.04_linux-nvidia-bos] CXL: Refresh CXL Reset Series to v9 and rebase VFIO CXL series

JiandiAnNVIDIA:
## Summary
This branch refreshes the CXL reset support on `7.0-bos` from the older
NV-SAUCE **v1 save/restore + v5 cxl-reset** stack to Srirangan's **combined v9
"CXL Reset for Type-2 devices"** series, re-bases Manish's **VFIO CXL Type-2
passthrough v2** series on top of it, and applies Manish's **internal 6-patch
VFIO CXL reset** series (rewired to the v9 reset API). The whole stack is
rebased onto the current `NVIDIA/NV-Kernels` `26.04_linux-nvidia-bos` tip.
Because the previous CXL/VFIO commits were already pushed, history is **not**
rewritten within the series: the stale work is removed with `git revert`, and
the replacement series is applied fresh on top. All re-applied NVIDIA SAUCE
commits keep their original `Signed-off-by` / `(backported from …)`
provenance, with `[jan: …]` notes documenting each v9 adaptation.

LP: https://bugs.launchpad.net/ubuntu/+source/linux-nvidia-bos/+bug/2163076

## Motivation
- **v9 supersedes v1/v5.** Srirangan's v9 is a single, self-contained series
  that replaces the split v1 (save/restore) + v5 (cxl-reset) work and is aligned
  to the upstream direction (`reset_method`-based ABI, HDM decoder caching,
  BAR-relative HDM restore).
- **VFIO CXL must ride on v9.** Manish's VFIO CXL Type-2 series depends on CXL
  core helpers whose API changed under v9; it is re-based and adapted to the v9
  conventions rather than the reverted v1/v5 layout.
- **Function-scoped VFIO reset on v9.** Manish's internal 6-patch reset series
  provides the guest-facing `vfio_cxl_reset()` path (DPA reset handling,
  post-reset BAR access, HDM shadow preservation, DVSEC STATUS2 virtualization),
  rewired from the removed v5 `cxl_dev_reset`/`cxl_do_reset` API to v9's
  `cxl_reset_function`/`cxl_reset_execute`.
- **Upstream-faithful header layout.** Public CXL surface is consolidated in
  `include/cxl/cxl.h` (Alejandro's convention, which Srirangan's v9 explicitly
  follows: *"move shared decoder settings to include/cxl/cxl.h"*), rather than
  the `include/cxl/pci.h` placement used by the earlier port.
## Rebase onto latest NV-Kernels
The full stack is rebased onto the current `NV-Kernels/26.04_linux-nvidia-bos`
tip (the fork base had diverged). To keep the rebase conflict-free, the
in-base commit `NVIDIA: VR: SAUCE: vfio/cxl: advertise DPA mmap and update its
selftest` (`386a76e8a826`) is reverted on the base before replay and
re-applied on top afterwards (see below). The only base-divergence conflict was
in `cxl_reset_done()` and was resolved to keep NV-Kernels' upstream fix
`1eaa33ea` (`cxl/pci: Check memdev driver binding status`).
## What's in this PR (bottom → top)
1. **Revert the stale NV-SAUCE CXL/ATS stack** (reverts only — no history
   rewrite): old VFIO CXL series + internal reset patches, v5 cxl-reset,
   v1 CXL DVSEC save/restore, and the NV-SAUCE ATS "always on" series.
2. **Re-apply the upstream ATS series** (cherry-picked with `-x`):
   - `PCI: Add pci_ats_required() for CXL.cache capable devices`
   - `PCI: Allow ATS to be always on for pre-CXL devices`
   - `iommu/arm-smmu-v3: Allow ATS to be always on`
   This restores the proper upstream owner of `PCI_DVSEC_CXL_CACHE_CAPABLE`
   and `pci_ats_required()`, which v9 depends on.
3. **Srirangan's v9 CXL Reset series (11 patches)** + one out-of-tree fixup
   (`cxl: Prepare IOMMU reset block after CXL cache WB+I`).
   Backported from: https://lore.kernel.org/linux-cxl/20260709010304.680422-1-smadhavan@nvidia.com/
4. **Retain** `PCI/CXL: Hide SBR from reset_methods if masked by CXL` (Vidya Sagar).
5. **Manish's VFIO CXL Type-2 v2 series (20 patches)**, adapted to v9.
   Backported from: https://lore.kernel.org/linux-cxl/20260401143917.108413-1-mhonap@nvidia.com/
   (v2 is used intentionally — v3 requires the not-yet-posted QEMU vfio-cxl v2.)
6. **Enable + guard for Type-2 passthrough:**
   - `config: Enable CONFIG_VFIO_CXL_CORE for CXL Type-2 passthrough`
   - `cxl: Guard unlinked memdev endpoints` (Nirmoy Das)
7. **Re-apply `vfio/cxl: advertise DPA mmap and update its selftest`**
   (`386a76e8a826`, Manish Honap), adapted onto v9 + VFIO v2. This is
   **functionally required**: without the `VFIO_REGION_INFO_FLAG_MMAP` flag the
   DPA region stays a slow-path I/O region the stage-2 listener skips, the GPU's
   ATS access to DPA never resolves, and the **guest GPU driver fails to load**
   when a CXL-mode GPU is passed through. It also registers the DPA range with
   `memory_failure()` (`register_pfn_address_space()`) for host SError
   containment.
8. **Manish's internal VFIO CXL reset series (6 patches)**, rewired to v9's
   `cxl_reset_function` / `cxl_reset_execute` (applied with `git am -s`):
   - `cxl: add function-scoped CXL reset entry for vfio-pci`
   - `vfio/pci: wire CXL DPA reset handling`
   - `vfio/cxl: borrow PCI Memory Space for post-reset BAR access`
   - `vfio/cxl: preserve guest HDM decoder shadow across reset`
   - `vfio/cxl: virtualize DVSEC STATUS2 in vconfig shadow`
   - `vfio/cxl: implement vfio_cxl_reset()`
## Key adaptations (v9 re-base)
- **Header placement → `include/cxl/cxl.h`.** `cxl_find_regblock`,
  `cxl_probe_component_regs`, `cxl_await_range_active`, and
  `cxl_regblock_get_bar_info` are declared in the public `include/cxl/cxl.h`
  under `CONFIG_CXL_BUS` (with `!CONFIG` stubs), matching upstream v2 and v9's
  convention, instead of the reverted `include/cxl/pci.h` layout. `pci.h` is
  kept minimal; redundant private `drivers/cxl/cxl.h` prototypes were dropped
  (including a follow-up fix to patch 1's `cxl_get_hdm_info`).
- **Register defines → `include/uapi/cxl/cxl_regs.h`.** Component/HDM register
  defines are provided via the new uapi header; the duplicate copies were
  removed from `drivers/cxl/cxl.h` to avoid `-Wmissing`/redefinition breakage.
- **Decoder-programming refactor respected.** `cxld_set_interleave` /
  `cxld_set_type` now live in `drivers/cxl/core/reset.c` (v9's reusable-helper
  split); VFIO patch context that re-introduced the old `hdm.c` copies was
  dropped to avoid unused-function build breaks.
- **`cxl_reset_done()` guard union.** During both the rebase and the internal
  reset series, `cxl_reset_done()` was resolved to keep NV-Kernels'
  `1eaa33ea` driver-binding guard (`if (!cxlmd->dev.driver) return;`) *and*
  Manish's tightened `IS_ERR_OR_NULL(cxlmd->endpoint)` endpoint check, since
  both harden the same invalid-endpoint (`-ENXIO` default) path.
- **`386a76` revert-then-reapply.** The advertise-DPA-mmap commit is reverted
  on the rebase base (to avoid modify/delete conflicts against the CXL/VFIO
  reverts) and re-applied on top, cleanly adapted to the evolved
  `vfio_cxl_register_cxl_region()` (v9/v2 `region_active` / `precommitted`
  logic) with no code changes to Manish's original hunks.
- **Functional/security fixes preserved** from the original port (AI-review +
  manual), e.g. BIR range validation (`reject BIR >= PCI_STD_NUM_BARS`) and
  `bar_index` bounds checks in `cxl_regblock_get_bar_info()`.
## Provenance
- All re-applied NVIDIA SAUCE commits retain original authorship,
  `Signed-off-by` chains, and `(backported from …)` trailers. The `386a76`
  re-apply carries a `(cherry picked from commit 386a76e8a826…)` trailer.
- The internal 6-patch reset series is applied with `git am -s` (adds a
  `Signed-off-by: Jiandi An`); patch 1 required a manual `cxl_reset_done()`
  merge (see Key adaptations).
- v9-specific conflict resolutions are documented inline with `[jan: …]` notes.


nirmoy:

The PR watcher skips automatic boro reviews for changes with more than 50 commits. This change currently has 90 commits.

```text
review https://github.com/NVIDIA/NV-Kernels/pull/514
```

Head: `b625d9a3e54e`

_This comment is maintained by nv-pr-bot. It is updated when the watcher sees a newer head._

sforshee:
I realize this is still a draft, but my review with claude turned up some issues that I thought I'd share. I reviewed these and tried to eliminate anything that didn't look legitimate, but some I'm still not sure about.

## High

### 1. `cxl_reset_done()` could dereference an error pointer

**File:** `drivers/cxl/pci.c:977`

```c
guard(device)(&cxlmd->dev);
if (cxlmd->endpoint &&
    cxl_endpoint_decoder_reset_detected(cxlmd->endpoint)) {
```

`cxlmd->endpoint` is initialised to `ERR_PTR(-ENXIO)` in `cxl_memdev_alloc()` (`drivers/cxl/core/memdev.c:776`) and only becomes a real `struct cxl_port *` when the endpoint port is registered (`drivers/cxl/core/port.c:859`). `ERR_PTR(-ENXIO)` is non-NULL, so the `&&` short-circuit does not protect  cxl_endpoint_decoder_reset_detected()`, which dereferences it.

It looks at least theoretically possible for this to happen. The memdev is created via `cxl_pci_probe()` and `cxlmd->endpoint` is set via a different device probe, `cxl_mem_probe()`.

### 2. DVSEC extended-capability header is no longer virtualized for CXL devices

**File:** `drivers/vfio/pci/cxl/vfio_cxl_config.c:216,229`

Both fallback paths in `vfio_cxl_dvsec_readfn()` changed from `vfio_direct_config_read()` to vfio_raw_config_read()`.

`vfio_direct_config_read()` exists specifically to overlay the first 4 bytes of an extended capability from `vconfig` (`drivers/vfio/pci/vfio_pci_config.c:258-260`) — that is how vfio rewrites the ecap next-pointer chain to hide capabilities from the guest. `vfio_raw_config_read()` skips all virtualisation (`vfio_pci_config.c:286-297`).

**Failure scenario:** a guest reading the DVSEC ecap header (`dvsec_off` 0..3) on a CXL device gets the raw hardware next-pointer, so its ecap walk can be steered into capabilities vfio mapped as `PCI_CAP_ID_INVALID`. This affects every DVSEC on the device, not just the CXL one (offsets outside the CXL DVSEC also take the changed fallback).

The dispatcher's own comment at `vfio_pci_config.c:1094` still documents the `direct` behaviour as intended:

> *"otherwise fall back to vfio_direct_config_read so non-CXL DVSEC devices keep the
> extended-cap header mangling that the default ecap readfn applies."*

### 3. CXL DPA mapping lifecycle regressions in vfio

**Source:** `46b4838278b8 Revert "NVIDIA: VR: SAUCE: vfio/pci: Wire CXL DPA reset handling"` — reverted, not re-applied.

Four independent gates were dropped:

| Gate | Location | Consequence |
|---|---|---|
| DPA zap in `vfio_pci_zap_and_down_write_memory_lock()` | `drivers/vfio/pci/vfio_pci_core.c:1652` | `vfio_pci_zap_bars()` only covers BAR offsets; the DPA region lives in the device-region offset range, so user mappings of CXL.mem survive runtime-PM entry, D3 transitions and Memory-Space disable |
| `pm_runtime_engaged \|\| !__vfio_pci_memory_enabled()` in the fault handler | `drivers/vfio/pci/cxl/vfio_cxl_core.c:645` | faults re-insert PFNs into a suspended / memory-disabled device |
| same check in `vfio_cxl_region_rw()` | `drivers/vfio/pci/cxl/vfio_cxl_core.c:~815` | same, for fd I/O |
| prepare/finish passes in `vfio_pci_dev_set_hot_reset()` and in `vfio_exp_config_write()` / `vfio_af_config_write()` | `vfio_pci_core.c:2550`, `vfio_pci_config.c:912,997` | after a bus reset or a guest-initiated FLR through PCIe Device Control, `region_active` stays true and `comp_reg_virt[]` is never re-read, so the guest keeps mapping DPA whose HDM decoders the reset just cleared |

Only `vfio_pci_ioctl_reset()` (`vfio_pci_core.c:1250,1273`) still handles this.

If dropping guest-driven `INITIATE_CXL_RESET` was the intent — `vfio_cxl_config.c:100-104` now just warns *"CXL reset requested but not yet supported"* — these four items are separable from that change and worth keeping. `vfio_cxl_comp_regs_rw()` also lost its `memory_lock` read-side serialisation against `comp_reg_virt[]` rebuild
(`drivers/vfio/pci/cxl/vfio_cxl_emu.c:283`).

---

## Medium

### 4. Sub-dword DVSEC writes corrupt the vconfig shadow

**File:** `drivers/vfio/pci/cxl/vfio_cxl_config.c:205-232` (read), `:292-312` (write)

**Reachability: confirmed, unconditional.** `vfio_config_do_rw()` (`drivers/vfio/pci/vfio_pci_config.c:1943-1949`) is the only sizing gate on the path and explicitly emits 1-byte chunks:

```c
count = min(count, vfio_pci_cap_remaining_dword(vdev, *ppos));
if (count >= 4 && !(*ppos % 4))      count = 4;
else if (count >= 2 && !(*ppos % 2)) count = 2;
else                                 count = 1;
```

`perm` resolves to `ecap_perms[PCI_EXT_CAP_ID_DVSEC]` -> `.writefn = vfio_pci_dvsec_dispatch_write` (`vfio_pci_config.c:1166`) -> `vdev->dvsec_writefn` = `vfio_cxl_dvsec_writefn`, called with `count == 1` and `__le32 val = 0` holding only that byte. Nothing in between rejects sub-dword access.

Register offsets (`include/uapi/cxl/cxl_regs.h:73-77`): CONTROL `0x0c`, STATUS `0x0e`, CONTROL2 `0x10`, STATUS2 `0x12`, LOCK `0x14`. PCI extended caps are dword-aligned, so `0x0c` / `0x10` / `0x14` are dword-aligned and `0x0e` / `0x12` are word- but not dword-aligned.

**Failure scenarios, most severe first:**

1. **A guest polling `STATUS2.CACHE_INVALID` with a 32-bit read can spin forever.** `vfio_cxl_dvsec_readfn` `case CXL_DVSEC_CONTROL2_OFFSET` does `memcpy(val, vdev->vconfig + pos, count)`; with `count == 4` that returns CONTROL2 **and STATUS2** out of vconfig. But `cxl_dvsec_status2_write()` never mirrors STATUS2 into vconfig — its own comment says *"STATUS2 is not mirrored in vconfig - reads go to hardware"* — and nothing updates it after `vfio_config_init()` seeds it from hardware. Meanwhile CONTROL2 bit 1 (`INITIATE_CACHE_WBI`) *is* forwarded to hardware (`vfio_cxl_config.c:135-141`). So a guest that triggers WBI and then polls with a dword read at `0x10` reads the stale seed value indefinitely. A 16-bit poll at `0x12` works. The removed `status2_hw_shadow_merge()` existed exactly to handle the dword case.

2. **1-byte write at CONTROL (`0x0c`) zeroes bits 15:8 of the shadow.** `wval16 = (u16)le32_to_cpu(val)` = the single byte -> `cxl_dvsec_control_write()` -> `dvsec_virt_write16(CONTROL, new_val)` writes all 16 bits. Lost bits: Cache_SF_Granularity (8:6), Cache_Clean_Eviction (9), P2P_Mem_Enable (12), Viral_Enable (14). Same shape for LOCK (`0x14`).

3. **Odd offsets are write-discarded but read from hardware.** `dvsec_off` `0x0d` / `0x0f` / `0x11` / `0x13` / `0x15` match no case in either switch: writes hit `default:` and are dropped, reads hit `default:` and go to `vfio_raw_config_read()`. So byte 1 of every virtualised register reads from hardware while byte 0 reads from shadow. Consequence: `VIRAL_STATUS` (STATUS bit 14, high byte) can never be cleared by any byte-granular write.

4. **4-byte write at `0x0c` or `0x10` silently drops the STATUS / STATUS2 half.** The reverted code forwarded it (`cxl_dvsec_status_write(vdev, val >> 16)`). A read of the same dword returns both halves, so read and write are asymmetric.

**Severity bound — does not reach hardware.** Every `pci_write_config_word()` in these handlers is gated on a low-byte bit: `INITIATE_CACHE_WBI` (bit 1) and `DISABLE_CACHING` (bit 0) in `cxl_dvsec_control2_write()`, `VOLATILE_HDM_PRES_ERROR` (bit 3) in `cxl_dvsec_status2_write()`. A 1-byte write carries those intact, so truncation cannot inject a wrong value into the device. `cxl_dvsec_control_write()` / `_status_write()` / `_lock_write()` only call `dvsec_virt_write16()`, which touches `vdev->vconfig` alone. This is guest-visible emulation incorrectness, not host or device state corruption — which is why this is Medium and finding 2 is not.

The reverted version handled all of this with `reg_start = dvsec_off & ~1u`, per-register shadow merging, and dword-spanning forwarding. Nothing replaced it.

### 5. Non-ARI Non-CXL Function Map bit index looks wrong

**File:** `drivers/cxl/core/reset.c:772-782`

```c
static int cxl_reset_func_map_bit(struct pci_dev *sibling, bool ari)
{
        if (ari)
                return sibling->devfn;
        /*
         * Without ARI, the Function Map is organized as 32 device slots per
         * conventional 3-bit function number.
         */
        return PCI_FUNC(sibling->devfn) * 32 + PCI_SLOT(sibling->devfn);
}
```

CXL r3.x section 8.1.4 defines the Non-CXL Function Map as one bit per *Function* of the device hosting the DVSEC (register 0 covers functions 0-31, register 1 functions 32-63, ...). The non-ARI index should be `PCI_FUNC(devfn)`.

PCIe endpoints are always device 0, so the expression evaluates to `PCI_FUNC * 32`: function 1 tests bit 32 instead of bit 1.

**Consequence:** fails safe — a sibling the device declared non-CXL is treated as CXL, so `cxl_reset_validate_function_scoped()` returns `-ENOTTY` and the method is never advertised on multi-function devices. The feature is silently disabled where it should work.

Introduced by `c2d27b26dd21`, backported from <https://lore.kernel.org/linux-cxl/20260709010304.680422-1-smadhavan@nvidia.com/>, so it likely needs fixing upstream too.

### 6. Missing CPU cache invalidation downgraded to a warning

**File:** `drivers/cxl/core/reset.c:1085-1089`

```c
if (!cpu_cache_has_invalidate_memregion()) {
        pci_warn(pdev,
                 "CPU cache synchronization unavailable; continuing without cache invalidation\n");
        return 0;
}
```

The rest of the CXL core treats this as fatal (`-ENXIO` / `-EPERM`) precisely because stale dirty lines over an HPA range whose decoders are about to be torn down are a data-corruption hazard. `cxl_hdm_ranges_request()` claiming the ranges via `request_mem_region()` limits exposure, but this is a deliberate weakening of an upstream invariant — justify it in the commit message or return an error.

### 7. `#ifdef CONFIG_CXL_BUS` silently disables the feature at `CXL_BUS=m`

**File:** `include/cxl/cxl.h:409`

`CXL_BUS` is tristate. With `CXL_BUS=m`, `CONFIG_CXL_BUS` is undefined (`CONFIG_CXL_BUS_MODULE` is defined instead), so `reset.c` — built under `CXL_HDM=y` — picks up the stub:

```c
static inline int
cxl_find_regblock(struct pci_dev *pdev, enum cxl_regloc_type type,
                  struct cxl_register_map *map)
{ return -EOPNOTSUPP; }
```

`__pci_cxl_hdm_init()` then always fails, `pdev->hdm` is never populated and `cxl_reset` never works — with no build error and only a `pci_dbg`. This config ships `CXL_BUS=y` so it does not bite today. IS_ENABLED(CONFIG_CXL_BUS)` would make it robust.

### 8. `probe` and do-reset disagree for `cxl_reset`

**File:** `drivers/cxl/core/reset.c:1338-1346`

```c
cxl_reset_context_init(&ctx, pdev);
if (probe)
        return cxl_reset_validate_function_scoped(&ctx);

cxl_hdm_range_context_init(&range_ctx);

if (!cxl_reset_scope_hdm_available(&ctx)) {
        rc = -ENOTTY;
        goto out;
}
```

The probe path never checks HDM availability, so `reset_method` advertises `cxl_reset` for devices where the actual reset always returns `-ENOTTY`. It falls through to the next method so nothing breaks, but the sysfs attribute lies, and the ABI doc (`Documentation/ABI/testing/sysfs-bus-pci:156-166`) only documents the scope condition.

---

## Low / cleanup

All in `drivers/cxl/core/reset.c` unless noted.

- **`__pci_cxl_hdm_init()` dead code and error-path oddities:** `info` is provably   NULL after the guard at `:401-405`, so `if (!info)` (`:425`) is always true, `allocated_info` is always true, and `info->decoder_count &&` at `:464` is always false. The `goto out_unmap` at `:495` leaves `rc == 0`, so the `pdev->hdm` race path frees `info` and returns success. `PCI_COMMAND` is restored twice on that path (`:487` then `:515`). The helper `cxl_hdm_enable_mem()` at `:306` does exactly what `:411-423` open-codes.
- **`cxl_pci_hdm_read_decoder()` `:362-365`:** `if (!committed) return 0; return 0;`
- **`cxl_reset_wait_done()` `:1259-1260`** computes `deadline` *before* the mandatory 100 ms RRS sleep. With `timeout_ms == CXL_RESET_RRS_WAIT_MS` (the two shortest spec encodings) the deadline has expired by the first poll, so the device gets exactly one status read.

nvmochs:
@JiandiAnNVIDIA - I really appreciate the detailed PR summary, it really helped with understanding the logic behind this refresh.

First, some comments / questions about provenance and patch trailers...

```
130f0ced9589 (jiandi/cxl-reset-v9) NVIDIA: VR: SAUCE: cxl: Guard unlinked memdev endpoints
21e9b8250690 NVIDIA: VR: SAUCE: config: Enable CONFIG_VFIO_CXL_CORE for CXL Type-2 passthrough
2b9035d80e7f selftests/vfio: Add CXL Type-2 VFIO assignment test
928ee9f765fa NVIDIA: VR: SAUCE: docs: vfio-pci: Document CXL Type-2 device passthrough
e880819f87d2 NVIDIA: VR: SAUCE: vfio/cxl: Provide opt-out for CXL feature
751e365983bc NVIDIA: VR: SAUCE: vfio/pci: Advertise CXL cap and sparse component BAR to userspace
9584c626c59b NVIDIA: VR: SAUCE: vfio/cxl: Register regions with VFIO layer
6a10b36eaf3a NVIDIA: VR: SAUCE: vfio/cxl: Virtualize CXL DVSEC config writes
c03c90cc08a9 NVIDIA: VR: SAUCE: vfio/cxl: DPA VFIO region with demand fault mmap and reset zap
f9a3bcbf018f NVIDIA: VR: SAUCE: vfio/cxl: CXL region management support
b0bdcce79bbf NVIDIA: VR: SAUCE: vfio/cxl: Wait for HDM ranges and create memdev
fd6a64c1288f NVIDIA: VR: SAUCE: vfio/cxl: Introduce HDM decoder register emulation framework
fa387d570cdb NVIDIA: VR: SAUCE: vfio/pci: Export config access helpers
45290db4788a NVIDIA: VR: SAUCE: vfio/cxl: Detect CXL DVSEC and probe HDM block
a3327409ecde NVIDIA: VR: SAUCE: vfio/pci: Add CONFIG_VFIO_CXL_CORE and stub CXL hooks
5d6de6a74baf NVIDIA: VR: SAUCE: vfio/pci: Add CXL state to vfio_pci_core_device
a3475c0c2b3a NVIDIA: VR: SAUCE: vfio: UAPI for CXL-capable PCI device assignment
e9a741772357 NVIDIA: VR: SAUCE: cxl: Record BIR and BAR offset in cxl_register_map
716183d3e893 NVIDIA: VR: SAUCE: cxl: Split cxl_await_range_active() from media-ready wait
8c62a8bf986b NVIDIA: VR: SAUCE: cxl: Move component/HDM register defines to uapi/cxl/cxl_regs.h
580a3bd18536 NVIDIA: VR: SAUCE: cxl: Declare cxl_probe_component_regs in public header
f23d55c57d86 NVIDIA: VR: SAUCE: cxl: Add cxl_get_hdm_info() for HDM decoder metadata
```

I cannot explicitly tell where these patches came from but it looks like these were basically re-picked them from the existing branch.

To make this clear, I think these patches should include a pick tag with the linux-nvidia-bos-7.0 SHA, then have annotation notes (if any for the v9 compat fixups) and then your SOB.

That way the trailers would be ordered like this:

```
<existing trailers preserved from original 7.0 commit>
<pick tag with original 7.0 commit SHA>
[jan: … if needed ...]
<jan SOB>
```

Happy to discuss further over Slack or a call if further clarification is needed.

<hr>

```
52e2392628e7 NVIDIA: VR: SAUCE: cxl: Prepare IOMMU reset block after CXL cache WB+I
```

Per the PR notes, this patch is an OOT fixup, which I assume means it has not been posted to LKML. However the commit message includes a LKML link to Srirangan’s reset series, is that intentional?

<hr>

```
16f32d50861c Revert "PCI/CXL: Hide SBR from reset_methods if masked by CXL”
```

While I don’t really have an issue with this, why are is this being reverted and then adding it back? Was it just to make the backporting cleaner?

<hr>

Now, for actual bug findings, my review with codex found a few overlaps with Seth’s review, so I’m not including those here. Of the items that Seth found but Codex missed, I had Codex check those findings to see if it agreed. The only one it did not agree with was Seth’s Medium 5 (Non-ARI Non-CXL Function Map bit index looks wrong) - here are the comments from Codex:

 -  I think this one is not legitimate. The code’s non-ARI index PCI_FUNC(devfn) * 32 + PCI_SLOT(devfn) at
    drivers/cxl/core/reset.c:772 matches the CXL non-ARI map layout I found: register number corresponds to Function N, bit x corresponds to Device x. A public CXL 2.0 excerpt shows register 0 mapping to Device x Function 0 and register 1 to Device x Function 1; the CXL 3.1 errata confirms the DVSEC placement on Device 0 Function 0, but not the peer’s PCI_FUNC()-only interpretation. Sources: CXL 2.0 excerpt (https://pdfcoffee.com/cxl-20-pdf-free.html), CXL 3.1 errata excerpt (https://www.scribd.com/document/824283059/Errata-and-Clarifications-to-the-Compute-Express-Link-Specification-Rev3-1-2024March22-eval).

So that one will take a bit more investigation to determine if there is actually an issue or not.

<hr>

Lastly, my Codex review found two new items:

  1. VFIO CXL: committed-but-unlocked HDM decoders can leak host HPA through COMP_REGS

  Source: re-picked VFIO CXL series, fd6a64c1288f (vfio/cxl: Introduce HDM decoder register emulation framework)
  File: drivers/vfio/pci/cxl/vfio_cxl_emu.c:459

  vfio_cxl_reinit_comp_regs() snapshots the hardware HDM decoder registers into comp_reg_virt[], then tries to sanitize firmware-committed decoders before exposing the shadow through the VFIO COMP_REGS region. The comment says committed decoder BASE must be zeroed so the host HPA is not exposed and QEMU can later program the guest GPA.

  However, the code only zeroes BASE_LOW/HIGH inside:
```
  if (ctrl & CXL_HDM_DECODER0_CTRL_LOCK) {
          ...
          *hdm_reg_ptr(cxl, base_lo_off) = 0;
          *hdm_reg_ptr(cxl, base_hi_off) = 0;
  }
```
  So if hardware reports a decoder as COMMITTED but not LOCKed, the shadow still contains the host HPA copied from hardware. That violates the intended VFIO ABI and can disclose host physical address layout to userspace/the guest stack. The selftest also documents that BASE is expected to be zero for firmware-committed decoders.

  Suggested fix: for every COMMITTED decoder, always zero BASE_LOW/HIGH in comp_reg_virt[]; only make clearing COMMIT_LOCK conditional on LOCK being set.

  2. CXL reset: HDM restore is skipped after a completed reset if cache re-enable fails

  Source: v9 CXL reset series. The return-code pattern comes from 1c0510182695 (cxl: Add CXL Device Reset helper) and becomes problematic once f7d2b743a4da adds HDM restore after reset.
  File: drivers/cxl/core/reset.c:1315

  cxl_reset_execute() waits for reset completion, then re-enables CXL caching. If reset completed successfully but cxl_reset_enable_cache() fails, the helper overwrites rc with that late error:
```
  else if (rc2)
          rc = rc2;
```
  The caller only restores HDM decoders when cxl_reset_execute() returns success:
```
  if (!rc)
          rc = cxl_restore_hdm_decoders(&ctx);
```
  That means a post-reset cache re-enable failure suppresses HDM restore even though the device reset already happened and volatile HDM decoder state may have been cleared. cxl_pci_target_reset_done() still runs afterward, so the system can unblock reset recovery with HDM state unrestored.

  Suggested fix: track “reset was initiated/completed” separately from the final return code. If reset completed, attempt HDM restore even when cache re-enable fails, then return/report the cache-enable error after preserving any restore failure ordering.

JiandiAnNVIDIA:
@nvmochs 

- trailers fixed for commits being picked from the same branch.
- lore link removed from the out of tree patch on top of Srirangan's cxl reset v9 series
- 6f32d50861c Revert "PCI/CXL: Hide SBR from reset_methods if masked by CXL”
[Matt Ochs]: While I don’t really have an issue with this, why are is this being reverted and then adding it back? Was it just to make the backporting cleaner? 
[Jiandi An]: It touches drivers/pci/pci.c and collides with Srirangan's v1 save/restore and v5 cxl-reset series (pci.c reset code)

JiandiAnNVIDIA:
**Latest refresh** of this PR rebases onto the tip of `NV-Kernels/26.04_linux-nvidia-bos` and applies Manish's internal 6-patch VFIO CXL reset series on top.

## Matt's and Seth's consolidated AI review issues (previous round)

```
• Bare minimum, I'd require these before integration:

    VFIO DVSEC read fallback must use vfio_direct_config_read()
        This is a uAPI correctness issue and is hit whenever userspace/guest reads
        DVSEC extended-cap headers or non-emulated DVSEC bytes.
        Current raw fallback can expose the real ecap next-pointer chain and defeat
        VFIO capability hiding.
        Small fix, high confidence.

    VFIO DVSEC sub-dword / spanning access handling
        Also uAPI. PCI config space legally supports byte/word/dword accesses, and
        VFIO's dispatcher definitely emits 1-byte chunks.
        Current code gives inconsistent shadow vs hardware reads and can make 32-bit
        CONTROL2/STATUS2 polling stale.
        This is reference-quality important because partners may copy the ABI behavior.

    VFIO CXL DPA lifecycle gates around reset / memory-disable / D3
        This is less "always hit" than DVSEC, but it is the largest operational footgun.
        If partners exercise guest FLR, hot reset, D3, or memory-space disable, stale DPA
        access can survive when it should be blocked or revalidated.
        I would at least restore the memory/D3 checks in DPA fd I/O and the reset
        zap/reactivate path for config FLR/hot reset.

    COMP_REGS committed decoder BASE zeroing
        uAPI/security-adjacent. The exposed VFIO region should not leak host HPA.
        The fix is tiny: zero BASE for any COMMITTED decoder, not only LOCKed decoders.

I'd put the cxl_reset_done() ERR_PTR fix just below that line: it is a clear bug and a
one-liner, but not uAPI and not certainly hit. Given the tiny fix, I'd still take it if
possible.

I would defer the HDM-restore-after-cache-enable-failure, missing CPU cache invalidation
policy, CONFIG_CXL_BUS=m, reset-method advertisement mismatch, and cleanup/timing nits if
you truly need the bare minimum.
```

Manish addressed several high-importance issues from the list above and fixed a couple of issues from his local kernel review prompts. He provided the 6-patch VFIO CXL reset series applied in this refresh:

```
5118f45c1253 NVIDIA: VR: SAUCE: vfio/cxl: implement vfio_cxl_reset()
d79829dc3d0c NVIDIA: VR: SAUCE: vfio/cxl: virtualize DVSEC STATUS2 in vconfig shadow
261abca60268 NVIDIA: VR: SAUCE: vfio/cxl: preserve guest HDM decoder shadow across reset
f2e0d49df870 NVIDIA: VR: SAUCE: vfio/cxl: borrow PCI Memory Space for post-reset BAR access
aadfa7575cbc NVIDIA: VR: SAUCE: vfio/pci: wire CXL DPA reset handling
c3912b870f20 NVIDIA: VR: SAUCE: cxl: add function-scoped CXL reset entry for vfio-pci
```

This refresh does **not** include any fix for Srirangan's v9 CXL reset series.

## Srirangan's comments on the v9 CXL reset series

```
1. cxl_reset_done() could dereference an error pointer
   The high priority bug is real and it didn't get flagged / I didn't notice it.
   So I fixed it in v10 now.

5. Non-ARI Non-CXL Function Map bit index looks wrong
6. Missing CPU cache invalidation downgraded to a warning
7. #ifdef CONFIG_CXL_BUS silently disables the feature at CXL_BUS=m
8. probe and do-reset disagree for cxl_reset

In these medium bugs:
  5 will get resolved in v10 since it will reject multifunction devices in the first place.
  6 is intentional. I don't want the reset to fail because cache flush SMCCC is unavailable
    (vera doesn't have the support as of yet).
  7 is a real v9 issue, fixed in v10 by moving the dependency to the private CXL header /
    build split.
  8 is fixed in v10.

The medium bugs don't break anything.
  For 5, that code path isn't actually executed on vera systems, so it is an upstreaming issue.
  For 7, the kernel config for the BOS kernel is fixed, right? If so, then we can get away with
    that too (unless it is causing issues with our config itself).
  For 8, that is totally fine (in fact in v10 I am intentionally leaving it in) because the HDM
    cache availability may not happen when the reset-method probe is happening (the HDM cache
    does get generated later). So I can't fail exposing cxl_reset at that point when the cache
    is unavailable.

Did some digging into the "high" issue also. Pasting the analysis in thread — looks like we
can ignore that too (or, if you want to fix it, the fix is also very simple and just needs to
use a macro in the if condition).
```

sforshee:
I still get several finding from codex. 1 and 3 look the most concerning, and 1 sounds familiar even though I don't see it mentioned above. I don't know enough about CXL to evaluate 2.

  1. [High] Revalidate reset scope when the reset executes — 492af82702d8

     pci_device_add() probes reset methods before adding the device to bus->devices (https://github.com/NVIDIA/NV-Kernels/blob/5118f45c1253d6c76b5b3645807d1523acf00290/drivers/pci/probe.c#L2753-L2761). Function 0 is also scanned before later functions, so its probe-time sibling walk is necessarily incomplete. cxl_reset_function() only validates siblings during probing (https://github.com/NVIDIA/NV-Kernels/blob/5118f45c1253d6c76b5b3645807d1523acf00290/drivers/cxl/core/reset.c#L1351-L1365); execution uses the cached result without revalidation.

     A multi-function CXL device can therefore advertise and execute a supposedly function-scoped reset that also resets active sibling functions. Revalidate immediately before execution under suitable PCI/topology serialization.

  2. [High] Do not treat a missing sibling-local CXL DVSEC as proof that the sibling is non-CXL — 492af82702d8

     After finding an in-scope sibling whose Non-CXL Function Map bit is clear, the code searches that sibling for its own CXL Device DVSEC and accepts it when none exists (https://github.com/NVIDIA/NV-Kernels/blob/5118f45c1253d6c76b5b3645807d1523acf00290/drivers/cxl/core/reset.c#L835-L845).

     The Function Map defines whether an existing function participates in CXL.cache/CXL.mem; when the map is absent, the Function 0 CXL DVSEC governs all functions. A participating sibling therefore need not carry its own DVSEC. This false negative can allow a reset that affects another CXL function. For any existing in-scope sibling, a clear map bit should block function-scoped reset; absent/unreadable maps should be conservative. See the official CXL 3.1 clarification for the Non-CXL Function Map (https://computeexpresslink.org/wp-content/uploads/2024/08/Errata-and-Clarifications-to-the-Compute-Express-Link-Specification-Rev3.1_2024July10_eval.pdf).

  3. [High] Fail reset when CPU cache invalidation is unavailable — dcb3c8a6e217

     With enabled HDM ranges, the code warns and returns success when no cache-maintenance provider exists (https://github.com/NVIDIA/NV-Kernels/blob/5118f45c1253d6c76b5b3645807d1523acf00290/drivers/cxl/core/reset.c#L1082-L1089). Reset then proceeds while stale CPU cache lines may still refer to memory whose contents or decoding changes across reset.

     This should return -ENXIO, matching the existing CXL region invalidation behavior, rather than treating cache synchronization as optional.

  4. [High] Keep the device quiesced after a post-reset failure — 36785a63b807 / 157596d7f48a

     Once reset initiation succeeds, a status timeout/error or decoder-restore failure causes the following sequence:
      - CXL.cache is re-enabled even on error (https://github.com/NVIDIA/NV-Kernels/blob/5118f45c1253d6c76b5b3645807d1523acf00290/drivers/cxl/core/reset.c#L1317-L1335).
      - HDM restoration is skipped or returns an error, but the IOMMU reset block is always released (https://github.com/NVIDIA/NV-Kernels/blob/5118f45c1253d6c76b5b3645807d1523acf00290/drivers/cxl/core/reset.c#L1362-L1370).

      - The outer PCI reset wrapper always restores PCI state (https://github.com/NVIDIA/NV-Kernels/blob/5118f45c1253d6c76b5b3645807d1523acf00290/drivers/pci/pci.c#L5198-L5204), potentially restoring Bus Master.

     The function can resume DMA/cache traffic with cleared or partially restored HDM decoders. Post-initiation errors need a failure state that survives the outer PCI restore, similar to the VFIO path’s compensating pci_clear_master().

Beyond these, a few minor issues of tagging/provenance:

- A number of commits say `cherry picked from ...` with notes about changes made, in which case they should say `backported from ...`.
- `24e87dcdbca9 selftests/vfio: Add CXL Type-2 VFIO assignment test` is not tagged SAUCE, yet it looks like it was taken from a mailing list patch to another branch and cherry picked from there. Seems like it should be tagged as SAUCE and indicate the branch from which it was cherry picked.
- The reset series all have the same lore link. I think each one is meant to link to the original patch in the series rather than all linking to the cover letter.

nvmochs:
> * The reset series all have the same lore link. I think each one is meant to link to the original patch in the series rather than all linking to the cover letter.

FYI, we typically only link to the series anchor rather than each individual patch.

nvmochs:
@JiandiAnNVIDIA 

Confirmed these issues from the prior review were resolved. Thanks to Manish and you for resolving!

```
  - cxl_reset_done() ERR_PTR dereference: fixed in c3912b870f20 with IS_ERR_OR_NULL(cxlmd->endpoint).
  - CONFIG_CXL_BUS=m public-header stub issue: fixed in c3912b870f20 by changing the public helper declarations to IS_ENABLED(CONFIG_CXL_BUS).
  - VFIO DPA lifecycle gaps: mostly fixed in aadfa7575cbc by adding vfio_cxl_prepare_reset(), vfio_cxl_finish_reset(), vfio_cxl_zap_dpa(), PM/memory-enabled fault gates, fd I/O gates, and hooks into ioctl
    reset, FLR config writes, and hot reset.

  - COMP_REGS HPA leak for committed decoders: fixed in 261abca60268; vfio_cxl_reinit_comp_regs() now zeros BASE for every committed decoder, not only locked ones.
  - COMP_REGS serialization against reset shadow rebuild: fixed in 261abca60268; vfio_cxl_comp_regs_rw() now takes memory_lock read-side.
  - DVSEC extended-capability header virtualization regression: fixed in d79829dc3d0c; fallback reads now use vfio_direct_config_read().
  - DVSEC sub-dword / dword-spanning emulation bugs: substantially fixed in d79829dc3d0c; byte writes merge correctly, high-byte accesses route to the right virtual register, STATUS/STATUS2 halves are handled
    on 32-bit writes, and STATUS2 reads merge hardware/shadow state.

  - STATUS2 CACHE_INVALID stale dword read issue: fixed in d79829dc3d0c via status2_hw_shadow_merge().
  ```

<hr>

Nit, all of the SHAs picked from linux-nvidia-bos branch need to have that branch identified after the SHA (otherwise we assume the SHA is an upstream SHA)

e.g.
```
(cherry picked from commit 7ca44b882feabac5a40fe2328db15abc899f7355 26.04_linux-nvidia-bos)
```

<hr>

For these two, the [jan] annotation note you added for this PR needs to come after the pick tag.

```
4663ecfc2a78 NVIDIA: VR: SAUCE: cxl: Declare cxl_probe_component_regs in public header
9c9f8d6120c8 NVIDIA: VR: SAUCE: cxl: Move component/HDM register defines to uapi/cxl/cxl_regs.h
```

Example: 
```
...
(cherry picked from commit e89694c4bb8e2867582fb7b6811069da9074a546)
[jan: keep Srirangan cxl-reset v9's cxl_hdm_info/cxl_commit/CONFIG_CXL_HDM block in include/cxl/cxl.h and dropped the moved register defines from the PRIVATE drivers/cxl/cxl.h (they were reverted out of public cxl.h on this branch, unlike the original port); the new uapi/cxl/cxl_regs.h now provides them to drivers/cxl via <cxl/cxl.h>. CXL_DECODER_MIN_GRANULARITY/MAX_ENCODED_IG and cxl_hdm_decoder_count stay private (used only in drivers/cxl, no vfio need).]
Signed-off-by: Jiandi An <jan@nvidia.com>
```

<hr>

My review of the new patches had 2 findings, however they are not critical and can be resolved in a later follow-up series.

Here are the findings from Codex:

  1. D3/runtime-PM DPA reactivation without HDM revalidation

  VFIO CXL now zaps DPA mappings on transient revoke paths through vfio_cxl_zap_dpa(), but that helper intentionally leaves region_active unchanged. That is fine for ordinary Memory Space disable / runtime suspend cases where HDM state is preserved, but vfio_pci_set_power_state() already handles devices where D3 transitions can perform a soft reset. If such a transition clears HDM decoder state, DPA access can become allowed again without a matching vfio_cxl_finish_reset() re-read of the decoder state.

  I think this can be follow-up because it depends on reset-like behavior during D3/runtime-PM, not the main CXL reset path. The normal explicit reset flows now have prepare/finish handling, so this is an edge lifecycle case rather than a guaranteed hit for basic passthrough/reset validation.

  2. CXL_Reset_Mem_Clr_Enable is not gated by CXL_Reset_Mem_Clr_Capable

  The guest-visible CONTROL2 shadow allows CXL_Reset_Mem_Clr_Enable, and the reset path passes it to hardware without checking whether the device advertises CXL_Reset_Mem_Clr_Capable. VFIO should mask or ignore that bit when unsupported.

  I think this can be follow-up because the fix is spec/uAPI cleanup around an optional reset modifier, not core reset functionality. It is worth fixing for reference correctness, but it should not block integration unless partners are expected to exercise Mem Clear immediately.

**Comment by sforshee** [comment]
> > * The reset series all have the same lore link. I think each one is meant to link to the original patch in the series rather than all linking to the cover letter.
> 
> FYI, we typically only link to the series anchor rather than each individual patch.

Thanks for clarifying ... I guess I've been doing it wrong 😅

**Comment by sforshee** [comment]
Provenance looks better. As discussed in slack, the hightlighted issues are not significant concerns currently, so fixes can be added later.

```
```

**Comment by nvmochs** [comment]
As mentioned in Srirangan's "comments on the v9 CXL reset series" above and as discussed offline, the additional findings are not an issue at this time and will be addressed in later revisions of these series.

Thanks for addressing the provenance nits! No further issues from me.

**Comment by clsotog** [comment]
Question with this commit:
https://github.com/NVIDIA/NV-Kernels/pull/514/changes/89c838c40f51e2428720c8b5b3d02c8948d45d00 
it was reverted at the beginning of this PR so when we add it back do we keep the acks from me and others or we just make it look like a cherry-pick from upstream with just your Signoff? 

Otherwise looks ok to me.

**Comment by JiandiAnNVIDIA** [comment]
> Question with this commit: [89c838c](https://github.com/NVIDIA/NV-Kernels/commit/89c838c40f51e2428720c8b5b3d02c8948d45d00) it was reverted at the beginning of this PR so when we add it back do we keep the acks from me and others or we just make it look like a cherry-pick from upstream with just your Signoff?
> 
> Otherwise looks ok to me.

I added the following before my final sign-off-by

 (cherry picked from commit 8aa66a105cedfeb3c74198ecad8a168e14aa9b57 26.04_linux-nvidia-bos)

**Comment by nvmochs** [comment]
Merged, closing PR.

```
902e1a008f27 (origin/26.04_linux-nvidia-bos) NVIDIA: VR: SAUCE: vfio/cxl: implement vfio_cxl_reset()
890cbce95cc9 NVIDIA: VR: SAUCE: vfio/cxl: virtualize DVSEC STATUS2 in vconfig shadow
d6331b30004f NVIDIA: VR: SAUCE: vfio/cxl: preserve guest HDM decoder shadow across reset
d4f9bd3c6f6e NVIDIA: VR: SAUCE: vfio/cxl: borrow PCI Memory Space for post-reset BAR access
c149c310bd25 NVIDIA: VR: SAUCE: vfio/pci: wire CXL DPA reset handling
320463e91149 NVIDIA: VR: SAUCE: cxl: add function-scoped CXL reset entry for vfio-pci
0d082a6bb2c7 NVIDIA: VR: SAUCE: vfio/cxl: advertise DPA mmap and update its selftest
c46dee57f0ba NVIDIA: VR: SAUCE: cxl: Guard unlinked memdev endpoints
431be84d0f1d NVIDIA: VR: SAUCE: config: Enable CONFIG_VFIO_CXL_CORE for CXL Type-2 passthrough
95a102ce0a32 NVIDIA: VR: SAUCE: selftests/vfio: Add CXL Type-2 VFIO assignment test
7c2ab2440813 NVIDIA: VR: SAUCE: docs: vfio-pci: Document CXL Type-2 device passthrough
792e110b94b7 NVIDIA: VR: SAUCE: vfio/cxl: Provide opt-out for CXL feature
3bf08746a712 NVIDIA: VR: SAUCE: vfio/pci: Advertise CXL cap and sparse component BAR to userspace
a8fdf9b84802 NVIDIA: VR: SAUCE: vfio/cxl: Register regions with VFIO layer
3570209d539a NVIDIA: VR: SAUCE: vfio/cxl: Virtualize CXL DVSEC config writes
9c6990af0c45 NVIDIA: VR: SAUCE: vfio/cxl: DPA VFIO region with demand fault mmap and reset zap
0ac36c93f381 NVIDIA: VR: SAUCE: vfio/cxl: CXL region management support
c2736e9ec1b3 NVIDIA: VR: SAUCE: vfio/cxl: Wait for HDM ranges and create memdev
9c9da8deec2b NVIDIA: VR: SAUCE: vfio/cxl: Introduce HDM decoder register emulation framework
748b73185123 NVIDIA: VR: SAUCE: vfio/pci: Export config access helpers
62872029f01a NVIDIA: VR: SAUCE: vfio/cxl: Detect CXL DVSEC and probe HDM block
6c2ed74460e0 NVIDIA: VR: SAUCE: vfio/pci: Add CONFIG_VFIO_CXL_CORE and stub CXL hooks
2065a26d7953 NVIDIA: VR: SAUCE: vfio/pci: Add CXL state to vfio_pci_core_device
b7790ef4bab1 NVIDIA: VR: SAUCE: vfio: UAPI for CXL-capable PCI device assignment
2dc8c7fcf11e NVIDIA: VR: SAUCE: cxl: Record BIR and BAR offset in cxl_register_map
caf23b9d6917 NVIDIA: VR: SAUCE: cxl: Split cxl_await_range_active() from media-ready wait
eca3b040744d NVIDIA: VR: SAUCE: cxl: Move component/HDM register defines to uapi/cxl/cxl_regs.h
4ee8b298dce9 NVIDIA: VR: SAUCE: cxl: Declare cxl_probe_component_regs in public header
dba5ceaee522 NVIDIA: VR: SAUCE: cxl: Add cxl_get_hdm_info() for HDM decoder metadata
aa36b9fc7b24 PCI/CXL: Hide SBR from reset_methods if masked by CXL
9de77e3907f3 NVIDIA: VR: SAUCE: cxl: Prepare IOMMU reset block after CXL cache WB+I
e24475800440 NVIDIA: VR: SAUCE: PCI/CXL: Restore HDM state after CXL bus reset
f2f8af84d5d3 NVIDIA: VR: SAUCE: Documentation/ABI: Document CXL Reset PCI reset method
d7a6975e486b NVIDIA: VR: SAUCE: PCI/CXL: Expose CXL Reset as a PCI reset method
9481c4c41f83 NVIDIA: VR: SAUCE: cxl: Restore CXL HDM state after PCI reset
4ddd20feb726 NVIDIA: VR: SAUCE: PCI/CXL: Discover the CXL reset scope
84dfe6c155d7 NVIDIA: VR: SAUCE: cxl: Validate HDM ranges before CXL reset
136562ef436f NVIDIA: VR: SAUCE: cxl: Add CXL Device Reset helper
5ee580b9c972 NVIDIA: VR: SAUCE: cxl: Cache endpoint decoder settings during PCI enumeration
38e0d6d7d83f NVIDIA: VR: SAUCE: cxl: Share HDM decoder decode logic
fb4782653150 NVIDIA: VR: SAUCE: cxl: Cache decoder settings on PCI devices
0b55bdb4aff9 NVIDIA: VR: SAUCE: cxl: Split decoder programming into a reusable helper
84d798ca1e5e iommu/arm-smmu-v3: Allow ATS to be always on
cc075804852a PCI: Allow ATS to be always on for pre-CXL devices
827608526c5e PCI: Add pci_ats_required() for CXL.cache capable devices
bebd7f0e80d7 Revert "NVIDIA: VR: SAUCE: PCI: Allow ATS to be always on for CXL.cache capable devices"
c81db6b741aa Revert "NVIDIA: VR: SAUCE: PCI: Allow ATS to be always on for pre-CXL devices"
cd8d2bab42eb Revert "NVIDIA: VR: SAUCE: iommu/arm-smmu-v3: Allow ATS to be always on"
f779a19db45e Revert "NVIDIA: VR: SAUCE: PCI: Add CXL DVSEC control, lock, and range register definitions"
3b2700a905c3 Revert "NVIDIA: VR: SAUCE: cxl: Move HDM decoder and register map definitions to include/cxl/cxl.h"
c2207e0f0050 Revert "NVIDIA: VR: SAUCE: PCI: Add virtual extended cap save buffer for CXL state"
07e6720223b5 Revert "NVIDIA: VR: SAUCE: PCI: Add cxl DVSEC state save/restore across resets"
d89f8264729b Revert "NVIDIA: VR: SAUCE: PCI: Add HDM decoder state save/restore"
1c9036a9a9a0 Revert "NVIDIA: VR: SAUCE: PCI: Add CXL DVSEC reset and capability register definitions"
3c3c970bf1dc Revert "NVIDIA: VR: SAUCE: PCI: Export pci_dev_save_and_disable() and pci_dev_restore()"
b762061350e3 Revert "NVIDIA: VR: SAUCE: cxl: Add memory offlining and cache flush helpers"
b8d82bf262c8 Revert "NVIDIA: VR: SAUCE: cxl: Add multi-function sibling coordination for CXL reset"
a2e112bde505 Revert "NVIDIA: VR: SAUCE: cxl: Add CXL DVSEC reset sequence and flow orchestration"
f186357261b4 Revert "NVIDIA: VR: SAUCE: cxl: Add cxl_reset sysfs interface for PCI devices"
3c9b046d50cb Revert "NVIDIA: VR: SAUCE: Documentation: ABI: Add CXL PCI cxl_reset sysfs attribute"
9b469cfd6d0a Revert "PCI/CXL: Hide SBR from reset_methods if masked by CXL"
aaf72b72644a Revert "NVIDIA: VR: SAUCE: cxl: Add cxl_get_hdm_info() for HDM decoder metadata"
361d850c74f0 Revert "NVIDIA: VR: SAUCE: cxl: Declare cxl_probe_component_regs in public header"
1175dc76b3ef Revert "NVIDIA: VR: SAUCE: cxl: Move component/HDM register defines to uapi/cxl/cxl_regs.h"
57c95d715863 Revert "NVIDIA: VR: SAUCE: cxl: Split cxl_await_range_active() from media-ready wait"
104532d18b04 Revert "NVIDIA: VR: SAUCE: cxl: Record BIR and BAR offset in cxl_register_map"
3ddd9e69933c Revert "NVIDIA: VR: SAUCE: vfio: UAPI for CXL-capable PCI device assignment"
7d990347ec78 Revert "NVIDIA: VR: SAUCE: vfio/pci: Add CXL state to vfio_pci_core_device"
18cdc3a614ec Revert "NVIDIA: VR: SAUCE: vfio/pci: Add CONFIG_VFIO_CXL_CORE and stub CXL hooks"
a8d541fc9bce Revert "NVIDIA: VR: SAUCE: vfio/cxl: Detect CXL DVSEC and probe HDM block"
20146f7d341f Revert "NVIDIA: VR: SAUCE: vfio/pci: Export config access helpers"
c3b7f7453c55 Revert "NVIDIA: VR: SAUCE: vfio/cxl: Introduce HDM decoder register emulation framework"
2ea7db86cdee Revert "NVIDIA: VR: SAUCE: vfio/cxl: Wait for HDM ranges and create memdev"
3b56fddea644 Revert "NVIDIA: VR: SAUCE: vfio/cxl: CXL region management support"
e9b986adfb3b Revert "NVIDIA: VR: SAUCE: vfio/cxl: DPA VFIO region with demand fault mmap and reset zap"
e96e84494169 Revert "NVIDIA: VR: SAUCE: vfio/cxl: Virtualize CXL DVSEC config writes"
5eaaaef2e2ad Revert "NVIDIA: VR: SAUCE: vfio/cxl: Register regions with VFIO layer"
f15a2d1b5a2a Revert "NVIDIA: VR: SAUCE: vfio/pci: Advertise CXL cap and sparse component BAR to userspace"
fbb7a8d15cf1 Revert "NVIDIA: VR: SAUCE: vfio/cxl: Provide opt-out for CXL feature"
d72f66053a50 Revert "NVIDIA: VR: SAUCE: docs: vfio-pci: Document CXL Type-2 device passthrough"
1be3a8eb2f54 Revert "selftests/vfio: Add CXL Type-2 VFIO assignment test"
d10d57ec6017 Revert "NVIDIA: VR: SAUCE: cxl: Export the CXL reset helpers for VFIO users"
f6e9a79af828 Revert "NVIDIA: VR: SAUCE: vfio/pci: Wire CXL DPA reset handling"
03d3b02b20d7 Revert "NVIDIA: VR: SAUCE: vfio/cxl: Ensure PCI Memory Space is enabled before post-reset BAR access"
b74c59bcb401 Revert "NVIDIA: VR: SAUCE: vfio/cxl: preserve HDM decoder base addresses across reset"
96175d773e9d Revert "NVIDIA: VR: SAUCE: vfio/cxl: virtualize DVSEC STATUS2 register in vconfig shadow"
d193056da348 Revert "NVIDIA: VR: SAUCE: vfio/cxl: Implement vfio_cxl_reset()"
15e1dda1827b Revert "NVIDIA: VR: SAUCE: config: Enable CONFIG_VFIO_CXL_CORE for CXL Type-2 passthrough"
aaf7cdab3faf Revert "NVIDIA: VR: SAUCE: cxl: Guard unlinked memdev endpoints"
8152f78c9675 Revert "NVIDIA: VR: SAUCE: vfio/cxl: advertise DPA mmap and update its selftest"
```

PR #539, opened 2026-08-10
[26.04_linux-nvidia-bos] Backport Vera I3C v7 support

nirmoy:
Backport the complete Vera I3C v7 series to the 26.04 BOS 7.0 kernel. The series adds the I3C core and DesignWare changes needed for Vera/Tegra410, ACPI matching for the controller, and SPD5118 hwmon support for DDR5 temperature monitoring.

This PR is only for `26.04_linux-nvidia-bos`; Vera does not require a companion non-BOS kernel PR.

The backport contains:

- all 12 patches from the upstream Vera I3C v7 series
- the complete five-commit mainline ACPICA LVR prerequisite chain
- one downstream BOS annotations commit that explicitly pins the inherited arm64 I3C and SPD5118 module policies
- linux-next commit `d2c743efd2d1ee64e94324664808f623dd865872`, which fixes the I3C unregister descriptor lifetime
- two NVIDIA: VR: SAUCE SETAASA follow-up fixes for firmware modalias generation and early device-add error propagation
- linux-next commit `038cf48b3170af26a70bf2dee4f8c3ac910f5176`, which fixes the empty-DAA UBSAN warning

Source series: https://lore.kernel.org/r/20260728065955.809445-1-akhilrajeev@nvidia.com

SETAASA follow-up series: https://lore.kernel.org/r/20260811103811.3380073-1-nirmoyd@nvidia.com

Backport adaptations for the Linux 7.0 BOS target are limited to:

- use `i3c_master_reattach_i3c_dev()` because this BOS branch predates the `_locked` rename
- resolve a context-only conflict in the ACPI child-enumeration patch because this branch predates surrounding I3C core changes; no code changes were made
- resolve a context-only conflict in the DesignWare clock-quirk patch because this branch predates the IBI data-capability detection block; no code changes were made

Validation:

- `git diff --check` passes against `26.04_linux-nvidia-bos`
- targeted builds pass for the affected ACPI, I3C, DesignWare I3C, and SPD5118 objects
- the upstream unregister fix applies cleanly and its stable patch ID matches linux-next
- annotation parsing confirms the explicit arm64 pins match the inherited `m` policies for both BOS flavours; amd64 continues to inherit its policies from `debian.master`
- the earlier PR head completed Vera boot and focused I2C/I3C runtime validation with an I3C-capable SBIOS; final runtime validation of the review-updated head is pending

LP: https://bugs.launchpad.net/bugs/2139154

### DIFF
--- a/drivers/i3c/master.c
+++ b/drivers/i3c/master.c
@@ -2285,8 +2285,12 @@ static int i3c_master_bus_init(struct i3c_master_controller *master)
 		 * addressable.
 		 */
 
-		if (i3cboardinfo->static_addr)
-			i3c_master_early_i3c_dev_add(master, i3cboardinfo);
+		if (i3cboardinfo->static_addr) {
+			ret = i3c_master_early_i3c_dev_add(master, i3cboardinfo);
+			if (ret && (i3cboardinfo->static_addr_method &
+				    I3C_ADDR_METHOD_SETAASA))
+				goto err_rstdaa;
+		}
 	}
 
 	/*
--- a/drivers/i3c/master.c
+++ b/drivers/i3c/master.c
@@ -5,6 +5,7 @@
  * Author: Boris Brezillon <boris.brezillon@bootlin.com>
  */
 
+#include <linux/acpi.h>
 #include <linux/atomic.h>
 #include <linux/bug.h>
 #include <linux/device.h>
@@ -2376,6 +2377,55 @@ EXPORT_SYMBOL_GPL(i3c_master_add_i3c_dev_locked);
 
 #define OF_I3C_REG1_IS_I2C_DEV			BIT(31)
 
+#ifdef CONFIG_ACPI
+static int i3c_acpi_get_i2c_resource(struct acpi_resource *ares, void *data)
+{
+	struct i2c_dev_boardinfo *boardinfo = data;
+	struct acpi_resource_i2c_serialbus *sb;
+
+	if (boardinfo->base.addr || !i2c_acpi_get_i2c_resource(ares, &sb))
+		return 1;
+
+	boardinfo->base.addr = sb->slave_address;
+	if (sb->access_mode == ACPI_I2C_10BIT_MODE)
+		boardinfo->base.flags |= I2C_CLIENT_TEN;
+
+	boardinfo->lvr = sb->lvr;
+
+	return 1;
+}
+
+static int i3c_acpi_add_i2c_boardinfo(struct i2c_dev_boardinfo *boardinfo,
+				      struct fwnode_handle *fwnode)
+{
+	struct acpi_device *adev = to_acpi_device_node(fwnode);
+	LIST_HEAD(resources);
+	int ret;
+
+	boardinfo->base.fwnode = acpi_fwnode_handle(adev);
+	acpi_set_modalias(adev, dev_name(&adev->dev), boardinfo->base.type,
+			  sizeof(boardinfo->base.type));
+
+	ret = acpi_dev_get_resources(adev, &resources,
+				     i3c_acpi_get_i2c_resource, boardinfo);
+	if (ret < 0)
+		return ret;
+
+	acpi_dev_free_resource_list(&resources);
+
+	if (!boardinfo->base.addr)
+		return -ENODEV;
+
+	return 0;
+}
+#else
+static inline int i3c_acpi_add_i2c_boardinfo(struct i2c_dev_boardinfo *boardinfo,
+					     struct fwnode_handle *fwnode)
+{
+	return -ENODEV;
+}
+#endif
+
 static int
 i3c_master_add_i2c_boardinfo(struct i3c_master_controller *master,
 			     struct fwnode_handle *fwnode, u32 *reg)
@@ -2392,6 +2442,15 @@ i3c_master_add_i2c_boardinfo(struct i3c_master_controller *master,
 		ret = of_i2c_get_board_info(dev, to_of_node(fwnode), &boardinfo->base);
 		if (ret)
 			return ret;
+
+		/* LVR is encoded in reg[2] for Device Tree. */
+		boardinfo->lvr = reg[2];
+	} else if (is_acpi_device_node(fwnode)) {
+		ret = i3c_acpi_add_i2c_boardinfo(boardinfo, fwnode);
+		if (ret) {
+			devm_kfree(dev, boardinfo);
+			return ret;
+		}
 	} else {
 		return -EINVAL;
 	}
@@ -2405,9 +2464,6 @@ i3c_master_add_i2c_boardinfo(struct i3c_master_controller *master,
 		dev_err(dev, "I2C device with 10 bit address not supported.");
 		return -EOPNOTSUPP;
 	}
-
-	/* LVR is encoded in reg[2]. */
-	boardinfo->lvr = reg[2];
 
 	list_add_tail(&boardinfo->node, &master->boardinfo.i2c);
 	fwnode_handle_get(fwnode);
@@ -2463,8 +2519,8 @@ i3c_master_add_i3c_boardinfo(struct i3c_master_controller *master,
 	return 0;
 }
 
-static int i3c_master_add_dev(struct i3c_master_controller *master,
-			      struct fwnode_handle *fwnode)
+static int i3c_master_add_of_dev(struct i3c_master_controller *master,
+				 struct fwnode_handle *fwnode)
 {
 	u32 reg[3];
 	int ret;
@@ -2488,6 +2544,74 @@ static int i3c_master_add_dev(struct i3c_master_controller *master,
 	return ret;
 }
 
+#ifdef CONFIG_ACPI
+static int i3c_master_add_acpi_dev(struct i3c_master_controller *master,
+				   struct fwnode_handle *fwnode)
+{
+	struct acpi_device *adev = to_acpi_device_node(fwnode);
+	acpi_bus_address adr;
+	u32 reg[3] = { 0 };
+	int ret;
+
+	/*
+	 * If the ACPI table entry has _ADR method, it's an I3C device.
+	 * Otherwise it may be an I2C device described by an I2cSerialBus
+	 * resource. If no I2cSerialBus resource is found, ignore the entry.
+	 */
+	if (!acpi_has_method(adev->handle, "_ADR")) {
+		ret = i3c_master_add_i2c_boardinfo(master, fwnode, reg);
+		if (ret == -ENODEV)
+			return 0;
+
+		return ret;
+	}
+
+	adr = acpi_device_adr(adev);
+
+	/* For I3C devices, _ADR will have the 48 bit PID of the device  */
+	reg[1] = upper_32_bits(adr);
+	reg[2] = lower_32_bits(adr);
+
+	fwnode_property_read_u32(fwnode, "mipi-i3c-static-address", &reg[0]);
+
+	return i3c_master_add_i3c_boardinfo(master, fwnode, reg);
+}
+
+static u8 i3c_acpi_i2c_get_lvr(struct i2c_client *client)
+{
+	struct acpi_device *adev = to_acpi_device_node(client->dev.fwnode);
+	struct i2c_dev_boardinfo boardinfo = {};
+	LIST_HEAD(resources);
+	int ret;
+	u8 lvr;
+
+	lvr = I3C_LVR_I2C_INDEX(2) | I3C_LVR_I2C_FM_MODE;
+
+	ret = acpi_dev_get_resources(adev, &resources,
+				     i3c_acpi_get_i2c_resource, &boardinfo);
+	if (ret < 0)
+		return lvr;
+
+	if (boardinfo.base.addr)
+		lvr = boardinfo.lvr;
+
+	acpi_dev_free_resource_list(&resources);
+
+	return lvr;
+}
+#else
+static inline int i3c_master_add_acpi_dev(struct i3c_master_controller *master,
+					  struct fwnode_handle *fwnode)
+{
+	return -ENODEV;
+}
+
+static inline u8 i3c_acpi_i2c_get_lvr(struct i2c_client *client)
+{
+	return I3C_LVR_I2C_INDEX(2) | I3C_LVR_I2C_FM_MODE;
+}
+#endif
+
 static int fwnode_populate_i3c_bus(struct i3c_master_controller *master)
 {
 	struct device *dev = &master->dev;
@@ -2499,7 +2623,13 @@ static int fwnode_populate_i3c_bus(struct i3c_master_controller *master)
 		return 0;
 
 	fwnode_for_each_available_child_node_scoped(fwnode, child) {
-		ret = i3c_master_add_dev(master, child);
+		if (is_of_node(child))
+			ret = i3c_master_add_of_dev(master, child);
+		else if (is_acpi_device_node(child))
+			ret = i3c_master_add_acpi_dev(master, child);
+		else
+			continue;
+
 		if (ret)
 			return ret;
 	}
@@ -2567,8 +2697,13 @@ static u8 i3c_master_i2c_get_lvr(struct i2c_client *client)
 	u8 lvr = I3C_LVR_I2C_INDEX(2) | I3C_LVR_I2C_FM_MODE;
 	u32 reg[3];
 
-	if (!fwnode_property_read_u32_array(client->dev.fwnode, "reg", reg, ARRAY_SIZE(reg)))
-		lvr = reg[2];
+	if (is_of_node(client->dev.fwnode)) {
+		if (!fwnode_property_read_u32_array(client->dev.fwnode, "reg",
+						    reg, ARRAY_SIZE(reg)))
+			lvr = reg[2];
+	} else if (is_acpi_device_node(client->dev.fwnode)) {
+		lvr = i3c_acpi_i2c_get_lvr(client);
+	}
 
 	return lvr;
 }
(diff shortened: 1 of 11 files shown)


clsotog:
I have this comments from Codex:

  - High: drivers/hwmon/spd5118.c:749, drivers/i3c/master.c:317, debian.nvidia-bos/config/annotations:273
    spd5118 is built as a module, but the new I3C driver only has an OF match table and no I3C modalias. The I3C device uevent still emits only MODALIAS=i3c:...; for PID-less SETAASA devices that becomes the all-zero I3C alias, so spd5118.ko will not autoload. It will bind if already loaded, but Tegra410 will likely miss the sensor by default. Add firmware modalias support to the I3C uevent path for these devices, or
    provide another working module alias/load mechanism.
  - Medium: drivers/i3c/master.c:2253
    i3c_master_early_i3c_dev_add() now has a SETAASA path that must succeed for PID-less devices, but its return value is still ignored. For SETDASA this was recoverable after ENTDAA; for SETAASA/PID-less devices it is not. If attach/reattach fails, bus init continues, SETAASA is sent, and no device is registered. Propagate failures at least when static_addr_method & I3C_ADDR_METHOD_SETAASA.

nvmochs:
@nirmoy 

a05cce2f8cec i3c: master: Add support for devices using SETAASA

Nit: Use backported from, no blank line between pick tag and context note

<hr>

One additional Codex finding...

• - High: debian.nvidia-bos/config/annotations:192 sets CONFIG_I3C=n on amd64 but leaves inherited I3C-dependent policies enabled. After olddefconfig, annotation validation reports 12 mismatches, including REGMAP_I3C, multiple I3C controllers/clients, I3C_OR_I2C, and SENSORS_SPD5118_DETECT. This can break amd64 configuration validation/package builds. Either retain I3C=m on amd64 or override all affected dependent policies.

jamieNguyenNVIDIA:
Reviewed all 17 commits at `f3c0e1f14840`. The backport itself is in very good
shape: 15 of the 16 cherry-picks are byte-identical to their cited upstream
commits, the one deviation is documented, and the affected objects build clean.
The points below are the delta on top of the annotations and Codex findings
already raised in this thread.

## Backport fidelity

Comparing the added and removed lines of each pick against its cited upstream
commit, with context and hunk headers excluded, exactly one source line in the
whole series differs from upstream — the one `a05cce2f8cec` documents:

```
 commit 9, a05cce2f8cec
-               ret = i3c_master_reattach_i3c_dev_locked(i3cdev, 0);
+               ret = i3c_master_reattach_i3c_dev(i3cdev, 0);
```

`09a82f00cd79`, `46f5daecde96` and `a6f892f3222e` show differences only in
context lines and `@@` header function names, from this tree lacking unrelated
upstream commits (no `<linux/bitmap.h>` in `master.c`; `dw-i3c-master.c` context
shows `master->maxdevs = ret >> 16` where upstream shows
`master->has_ibi_data = true`). Payloads are identical, so no adaptation was
needed and none was made.

Tree attribution is correct throughout and the two trailer styles in this PR are
not inconsistent: the five ACPICA picks cite bare SHAs because those commits are
in mainline `v7.2-rc1`, while the i3c and hwmon picks carry `linux-next` because
they are in no release tag.

Built at the series tip for arm64 (`defconfig` plus `CONFIG_ACPI=y`,
`CONFIG_I3C=m`, `CONFIG_DW_I3C_MASTER=m`, `CONFIG_SENSORS_SPD5118=m`):

```
  CC [M]  drivers/hwmon/spd5118.o
  CC [M]  drivers/i3c/master.o
  CC [M]  drivers/i3c/master/dw-i3c-master.o
  CC      drivers/acpi/acpica/rsserial.o
  LD [M]  drivers/i3c/i3c.o
```

No warnings, and every identifier introduced by the backported code resolves in
this tree at the commit that uses it.

## 1. `a05cce2f8cec`: the backport note names the wrong commit

The note reads:

```
[nirmoy: use the pre-8d8afa428318 static reattach helper name]
```

`8d8afa428318` ("i3c: master: Expose the APIs to support I3C hub") did not rename
anything. It exported a helper that was already called `*_locked`:

```
-static int i3c_master_reattach_i3c_dev_locked(struct i3c_dev_desc *dev,
+/**
+ * i3c_master_reattach_i3c_dev_locked() - reattach an I3C device with a new address
```

The rename was `79c41666b397` ("i3c: master: rename i3c_master_reattach_i3c_dev()
to *_locked"), the immediately preceding patch in the same NXP series, and it is
a pure rename of one definition and two call sites:

```
-static int i3c_master_reattach_i3c_dev(struct i3c_dev_desc *dev,
+static int i3c_master_reattach_i3c_dev_locked(struct i3c_dev_desc *dev,
```

As written the note is also self-contradictory: *pre*-`8d8afa428318` the helper
was already named `*_locked`, which is the opposite of what the backport uses.

The adaptation itself is correct — this tree has
`static int i3c_master_reattach_i3c_dev(...)` with a body byte-identical to
upstream's `*_locked` version, the rename is applied to all three sites, and
`git grep i3c_master_reattach_i3c_dev_locked a05cce2f8cec` returns nothing
tree-wide. Only the SHA in the note needs correcting to `79c41666b397`.

## 2. `f3c0e1f14840`: the arm64 half of the commit is a no-op

The subject is "Enable I3C and SPD5118 for Tegra410", but
`debian.nvidia-bos/config/annotations` opens with

```
include "../../debian.master/config/annotations"
```

and debian.master already carries all three symbols at `m` for arm64:

```
CONFIG_I3C              policy<{'amd64': 'm', 'arm64': 'm', ..., 's390x': 'n'}>
CONFIG_DW_I3C_MASTER    policy<{'amd64': 'm', 'arm64': 'm', ...}>
CONFIG_SENSORS_SPD5118  policy<{'amd64': 'm', 'arm64': 'm', ...}>
```

Exporting the flavour fragments before and after this commit confirms nothing
changes on the target:

```
arm64-nvidia-bos:     IDENTICAL before/after
arm64-nvidia-bos-64k: IDENTICAL before/after

amd64-nvidia-bos:
-CONFIG_DW_I3C_MASTER=m        +# CONFIG_DW_I3C_MASTER is not set
-CONFIG_I3C=m                  +# CONFIG_I3C is not set
-CONFIG_SENSORS_SPD5118=m      +# CONFIG_SENSORS_SPD5118 is not set
```

So the commit's only net effect is to *disable* three symbols on amd64. The
commit message should say that, or the arm64 entries should be dropped as
redundant.

## 3. `f3c0e1f14840`: disabling SPD5118 on amd64 removes an existing feature

`i2c_register_spd()` in `drivers/i2c/i2c-smbus.c` instantiates the `"spd5118"`
device for SMBIOS memory types `0x22` and `0x23`, which is the x86 DMI path named
in the `SENSORS_SPD5118_DETECT` help text. Setting `CONFIG_SENSORS_SPD5118=n` on
amd64 therefore removes DDR5 module temperature reporting from
amd64-nvidia-bos, which is a user-visible regression unrelated to Tegra410
enablement.

Since the arm64 side needs no change at all (point 2), leaving all three symbols
at their inherited `m` on both arches would satisfy the stated goal and avoid
both this and the annotations failure.

## Notes on the annotations finding already raised

The `CONFIG_I3C=n` breakage reported above in this thread reproduces. Running the
real tool at both the commit and its parent isolates what this commit causes from
pre-existing environment noise:

```
parent (f3c0e1f14840^):  28 mismatches
tip    (f3c0e1f14840):   41 mismatches
                         -- 13 added, all I3C related
```

```
CONFIG_AD4062              changed from m to -
CONFIG_ADI_I3C_MASTER      changed from m to -
CONFIG_CDNS_I3C_MASTER     changed from m to -
CONFIG_DW_I3C_MASTER       changed from n to -
CONFIG_I3C_OR_I2C          changed from m to y
CONFIG_IIO_ST_LSM6DSX_I3C  changed from m to -
CONFIG_INV_ICM45600_I3C    changed from m to -
CONFIG_MCTP_TRANSPORT_I3C  changed from m to -
CONFIG_MIPI_I3C_HCI        changed from m to -
CONFIG_MIPI_I3C_HCI_PCI    changed from m to -
CONFIG_REGMAP_I3C          changed from m to -
CONFIG_SENSORS_SPD5118_DETECT changed from n to -
CONFIG_SVC_I3C_MASTER      changed from m to -
```

Twelve of these are inherited from debian.master. The thirteenth,
`CONFIG_DW_I3C_MASTER`, is annotated `'n'` by this commit where the resolved
value is `-`; the in-file precedent for a symbol that stops being emitted is `'-'`
(see `CONFIG_GENERIC_CPU_CACHE_MAINTENANCE`).

`CONFIG_I3C_OR_I2C` flips rather than disappears because of its default:

```
config I3C_OR_I2C
	tristate
	default m if I3C=m
	default I2C
```

With `I3C=n` and `CONFIG_I2C=y` it resolves to `y` against an inherited `'m'`.

The check is build-gating: `debian/rules.d/2-binary-arch.mk` runs it inside
`stamp-prepare-%`, skipped only when `do_mainline_build` sets `do_skip_checks`.

The Codex autoload finding also reproduces — `spd5118.c` declares
`MODULE_DEVICE_TABLE` for `i2c` and `of` but not `i3c`, and `i3c_device_uevent()`
emits only `MODALIAS=i3c:dcr%02Xmanuf%04Xpart%04Xext%04X` with no firmware
modalias fallback of the kind `i2c_device_uevent()` has. For a PID-less SETAASA
device `pid` is 0, and the SETAASA path in `i3c_master_early_i3c_dev_add()`
returns before `i3c_master_retrieve_dev_info()`, so `dcr` is 0 as well and the
alias is all zeros.

nirmoy:
@clsotog @nvmochs @jamieNguyenNVIDIA Thanks for the detailed reviews. let me know how the curr version looks

drivers/i3c/master.c
```diff
@@ -282,9 +287,27 @@ static ssize_t modalias_show(struct device *dev,
 {
 	struct i3c_device *i3c = dev_to_i3cdev(dev);
 	struct i3c_device_info devinfo;
+	u8 static_addr_method = 0;
 	u16 manuf, part, ext;
+	int ret;
+
+	i3c_bus_normaluse_lock(i3c->bus);
+	if (i3c->desc->boardinfo)
```
jamieNguyenNVIDIA: Only one finding remaining from me: I think there's a potential NULL pointer dereference here:

`i3c->desc` can be NULL — `i3c_device_get_info()` guards it under the same lock,
and `i3c_device_uevent()` just below guards it too. `i3c_master_unregister_i3c_devs()`
sets `i3cdev->dev->desc = NULL` before calling `device_unregister()`, so the
attribute is still live in sysfs at that point and a read of
`/sys/bus/i3c/devices/*/modalias` racing a master unbind would dereference NULL.
The pre-patch version didn't touch `->desc` at all, it only called the NULL-safe
`i3c_device_get_info()`.

Adding the guard on its own would leave `devinfo` unwritten on that path, so the
`sprintf()` below would copy uninitialised stack into a world-readable file —
hence the `= {}` as well. Reading `info` inside the same critical section also
avoids taking the bus lock twice, and avoids `desc` being cleared between the two
sections and leaving `static_addr_method` and `devinfo` inconsistent.

Suggested fix:

```diff
-	struct i3c_device_info devinfo;
+	struct i3c_device_info devinfo = {};
 	u8 static_addr_method = 0;
 	u16 manuf, part, ext;
 	int ret;
 
 	i3c_bus_normaluse_lock(i3c->bus);
-	if (i3c->desc->boardinfo)
-		static_addr_method = i3c->desc->boardinfo->static_addr_method;
+	if (i3c->desc) {
+		devinfo = i3c->desc->info;
+		if (i3c->desc->boardinfo)
+			static_addr_method = i3c->desc->boardinfo->static_addr_method;
+	}
 	i3c_bus_normaluse_unlock(i3c->bus);
 
-	i3c_device_get_info(i3c, &devinfo);
-
```

nvmochs: > Only one finding remaining from me: I think there's a potential NULL pointer dereference here:
> 
> `i3c->desc` can be NULL — `i3c_device_get_info()` guards it under the same lock, and `i3c_device_uevent()` just below guards it too. `i3c_master_unregister_i3c_devs()` sets `i3cdev->dev->desc = NULL` before calling `device_unregister()`, so the attribute is still live in sysfs at that point and a read of `/sys/bus/i3c/devices/*/modalias` racing a master unbind would dereference NULL. The pre-patch version didn't touch `->desc` at all, it only called the NULL-safe `i3c_device_get_info()`.
> 
> Adding the guard on its own would leave `devinfo` unwritten on that path, so the `sprintf()` below would copy uninitialised stack into a world-readable file — hence the `= {}` as well. Reading `info` inside the same critical section also avoids taking the bus lock twice, and avoids `desc` being cleared between the two sections and leaving `static_addr_method` and `devinfo` inconsistent.
> 
> Suggested fix:
> 
> ```diff
> -	struct i3c_device_info devinfo;
> +	struct i3c_device_info devinfo = {};
>  	u8 static_addr_method = 0;
>  	u16 manuf, part, ext;
>  	int ret;
>  
>  	i3c_bus_normaluse_lock(i3c->bus);
> -	if (i3c->desc->boardinfo)
> -		static_addr_method = i3c->desc->boardinfo->static_addr_method;
> +	if (i3c->desc) {
> +		devinfo = i3c->desc->info;
> +		if (i3c->desc->boardinfo)
> +			static_addr_method = i3c->desc->boardinfo->static_addr_method;
> +	}
>  	i3c_bus_normaluse_unlock(i3c->bus);
>  
> -	i3c_device_get_info(i3c, &devinfo);
> -
> ```

Codex agrees this is a valid finding.

It also suggests this related upstream (in linux-next) fix:
d2c743efd2d1 i3c: master: Fix info leak and UAF in device unregister path

nvmochs:
@nirmoy Confirmed that the latest update addresses my prior findings.

```
bd24cdc77ffd i3c: dw-i3c-master: Add ACPI core clock frequency quirk
```

Was this a clean pick or did it require context fixup? (codex says replaying the branch produces a conflict for this patch)

<hr>

For any new patches / fixes, please be sure to add them to the tracking SS (new rows under the existing i3c row).

clsotog:
This is the P1 finding from my codex session:
NV-Kernels/drivers/i3c/master.c:2770: ACPI child devices are consumed by the I3C parent but never marked enumerated. i3c_master_add_acpi_dev() builds boardinfo from ACPI fwnodes, then the core registers I3C/I2C child devices, but there is no acpi_device_set_enumerated() / clear pairing like I2C and SPI use. That can let _ADR-only I3C children fall through to ACPI default
    platform-device creation, and leaves serial-bus children looking unhandled to ACPI remove/reconfiguration paths.

nirmoy:
## BaseOS Kernel Review

### Summary

Boro reported 8 findings.

Findings: Critical: 0, High: 0, Medium: 3, Low: 4, Info: 1

Latest watcher review: [open review](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/539/latest.html)

Generated test plan: [open test plan](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/539/test-plan.html)

Kernel deb build: **successful** ([download debs](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/539/debs/), 4 files)

Head: `01bde9aa65e4`

drivers/i3c/master.c
nirmoy: upstream code handles it better. For us we need d2c743efd2d1 and more guard. Updated

nirmoy:
> @nirmoy Confirmed that the latest update addresses my prior findings.
> 
> ```
> bd24cdc77ffd i3c: dw-i3c-master: Add ACPI core clock frequency quirk
> ```
> 
> Was this a clean pick or did it require context fixup? (codex says replaying the branch produces a conflict for this patch)
> 
It was not clean. context change needed, no code change ! Updated the patch to say `backported from ...`
> For any new patches / fixes, please be sure to add them to the tracking SS (new rows under the existing i3c row).

Will add those

nirmoy:
> This is the P1 finding from my codex session: NV-Kernels/drivers/i3c/master.c:2770: ACPI child devices are consumed by the I3C parent but never marked enumerated. i3c_master_add_acpi_dev() builds boardinfo from ACPI fwnodes, then the core registers I3C/I2C child devices, but there is no acpi_device_set_enumerated() / clear pairing like I2C and SPI use. That can let _ADR-only I3C children fall through to ACPI default platform-device creation, and leaves serial-bus children looking unhandled to ACPI remove/reconfiguration paths.

Codex says: 
```
The I2cSerialBus lifecycle concern is valid for dynamic ACPI reconfiguration, but I3C currently has no ACPI hotplug notifier.
```
So a upstream gap that is outscope for this PR.

jamieNguyenNVIDIA:

nvmochs:
> > @nirmoy Confirmed that the latest update addresses my prior findings.
> > ```
> > bd24cdc77ffd i3c: dw-i3c-master: Add ACPI core clock frequency quirk
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Was this a clean pick or did it require context fixup? (codex says replaying the branch produces a conflict for this patch)
> 
> It was not clean. context change needed, no code change ! Updated the patch to say `backported from ...`
> 
> > For any new patches / fixes, please be sure to add them to the tracking SS (new rows under the existing i3c row).
> 
> Will add those

Thanks for addressing this, no further issues from me!

clsotog:

jamieNguyenNVIDIA:
Applied: https://git.launchpad.net/~canonical-kernel/ubuntu/+source/linux-nvidia/+git/resolute/commit/?h=nvidia-bos-next&id=f8e341d3ffa90eabbcc15365adbeb4609ed846b1

Closing PR.

PR #543, opened 2026-08-12
[26.04_linux-nvidia-bos] MPAM update

ltrager:
# Overview

When backporting MPAM to the linux-nvidia-6.18 kernel I took the latest MPAM/resctrl patch series from LORE and Linus's tree. This left linux-nvidia-6.18-next with newer MPAM/resctrl support than 26.04_linux-nvidia-bos. This PR updates 26.04_linux-nvidia-bos MPAM/resctrl support to mirror what is in linux-nvidia-6.18-next, while retaining the required BOS-specific integration changes.

I have not yet created a bug on launchpad yet as I'd like to get feedback as to whether we want to proceed or not.

LP Bug: https://bugs.launchpad.net/ubuntu/+source/linux-nvidia/+bug/2167008

# Updated Patch Series

| Series | Changes |
| ------|-----------|
| [arm_mpam: Add MPAM-Fb firmware support v9](https://lore.kernel.org/r/20260807105451.3728284-1-andre.przywara@arm.com) | Updates the MPAM-Fb support to the complete 12-patch v9 series. Nine patches match LORE directly. Three were adapted to preserve BOS MAX/HARDLIM probing, MBW_PBM/CMAX/CMIN programming, and downstream programming-error handling in the threaded IRQ conversion. |
| [arm_mpam: resctrl: Counter Assignment (ABMC) v5](https://lore.kernel.org/r/20260707162440.12132-1-ben.horgan@arm.com) | Replaces the older Arm ABMC implementation with the five-patch v5 series and refreshes the associated ABMC/CDP KUnit coverage. The obsolete free-running-monitor and RMID-hook implementations are absorbed by the v5 design. |

# New Patch Series
New patch series applied to resolve conflicts with updated series

| Subject | Source |
|---------|--------|
| [x86,fs/resctrl: Pave the way for MPAM counter assignment v7](https://lore.kernel.org/r/20260506082855.3694761-1-ben.horgan@arm.com) | Adds all seven generic resctrl prerequisites using their final commits from Linus’s tree. |
| [fs/resctrl: IOMMU group fixes for tasks interface and cleanup v2](https://lore.kernel.org/r/20260414032610.1523958-1-zengheng4@huawei.com) | Adds all three v2 fixes for PID parsing, helper cleanup, and IOMMU-group migration when resctrl groups are removed. |
| KVM: arm64: Force guest EL1 to use user-space's partid configuration | [67faed4ccb4f](https://git.kernel.org/linus/67faed4ccb4f4b8055f50e9cc5670f1fc09ea287) |
| arm_mpam: Update architecture version check for MPAM MSC | [50a42e03cdbd](https://git.kernel.org/linus/50a42e03cdbd77d93366d301db3d367dce78eda6) |
| arm_mpam: Fix MPAMCFG_MBW_PBM register setting | [021118ce5ea9](https://git.kernel.org/linus/021118ce5ea954ec316d7e30bcf4506e12eb5222)  |

# Patches now sourced from Linus's tree

* [7625632fed43](https://git.kernel.org/linus/7625632fed431ddd655e839c302165536553f767) — fs/resctrl: Tidy up the error path in resctrl_mkdir_event_configs()
* [94a1206522d1](https://git.kernel.org/linus/94a1206522d11302ae7e7c28d3d494c8f0c9c58e) — x86,fs/resctrl: Create 'event_filter' files read only if they're not configurable
* [f52abe650241](https://git.kernel.org/linus/f52abe6502413450b8d0ecaad2555bbe4c6242eb) — fs/resctrl: Disallow the software controller when MBM counters are assignable
* [ee3d4c81d89c](https://git.kernel.org/linus/ee3d4c81d89c92fbeb65807971ac22b3dfa49220) — fs/resctrl: Add monitor property 'mbm_cntr_assign_fixed'
* [3aec86e4ea01](https://git.kernel.org/linus/3aec86e4ea013c084a232c83754d182c9aaf378e) — fs/resctrl: Continue counter allocation after failure
* [9a1646211f8c](https://git.kernel.org/linus/9a1646211f8c67c7c98f8109607e3b962aea13eb) — fs/resctrl: Document that automatic counter assignment is best effort
* [1cfa74c683ea](https://git.kernel.org/linus/1cfa74c683ea82d37156ccd7ab4f4659056dc701) — fs/resctrl: Document tasks file behaviour for task id 0 and idle tasks

# Reverted and Reapplied Patches

The older versions were reverted before replaying their replacements in dependency order. This avoids layering the newer series over overlapping BOS SAUCE implementations and makes the required conflict resolutions explicit.

 | Reverted functionality | Reverted commit | Reapplied commit |
  |---|---|---|
  | Apply T241-MPAM-6 to 63-bit counters | `0f0f62d1cd92` | `62b5c4d7ec21` |
  | Fix `mbm_L3_assign` and `mon_local_bytes` | `9d97d90c12bc` | `9343a4d7c69a` |
  | Add MBA `max_lim` sysfs support | `eb2a397ffd99` | `0375f48339eb` |
  | Probe MBW MAX_LIM and `hardlim_rw` | `3f89f038d12e` | `497927a4144f` |
  | Reset RIS using the explicit default configuration | `39c973a9670a` | `2c2b0824b283` |
  | Include associated MSC components during domain setup | `3ce46c56e790` | `40878d018a1f` |
  | Avoid MSC teardown for software programming errors | `7abd6df46f5e` | `68f7f93b2b89` |
  | Determine whether exposed counters can overflow | `280d19355685` | `a4bdf5756702` |
  | Assign IOMMU groups to resctrl groups | `69037f273891` | `37844a5cdb78` |
  | Add IOMMU QoS state helpers | `9f1da5623808` | `e87d5188d3bd` |
  | Add SMMUv3 MPAM query and set helpers | `f1c8da0cfad3` | `98377e75e771` |
  | Register SMMU capabilities with MPAM | `4bfc66310af1` | `2647763ba37e` |
  | Add the `this_is_not_abi` mount option | `1bf2d6c7b578` | `57dc56521ca0` |
  | Allow an MSC to be forced to have an unknown location | `5ddc3a894634` | `a94bee05d160` |
  | Allow the maximum PARTID to be overridden | `7a882f69a85b` | `0694f0e9b3af` |
  | Split locking around the `mon_sel` registers | `0d9b271e9466` | `e5a6b275c27c` |
  | Expose NRDY retry counts in debugfs | `154491fb0815` | `17c6383eb1fb` |
  | Add the MSC/RIS debugfs hierarchy | `56836799ed9a` | `b65066184f81` |
  | Avoid the `dom_data_exit()` and `rmid_ptrs[]` race | `4dee6626ad77` | `9176563c832a` |
  | Add ABMC counter read and reset hooks | `a639261dd308` | `bf6dfdc6ce01` |
  | Add `resctrl_arch_config_cntr()` for ABMC | `77c4e9f9d629` | `3dd88e6bf668` |
  | Add ABMC/CDP KUnit coverage | `a199cc64eb18` | `fe2c2b28b50f` |
  | Pre-allocate assignable monitors | `2d109004824f` | `6550788495fe` |
  | Pick MPAM classes for use as MBM counters | `eaff6c2bde68` | `a466792cdf0f` |
  
  # Testing
  
  Built and ran on a Vera system. Greenlit passes all MPAM tests
  
  ```
  nvidia@localhost:/home/nvidia/ltrager/greenlit$ sudo glt-test-suite --test-case arm64:test_mpam_resctrl_mount --test-case arm64:test_mpam_resctrl_schemata --test-case arm64:test_mpam_l3_occupancy --test-case arm64:test_mpam_mba_schemata --test-case arm64:test_mpam_resctrl_info
2026-08-11 19:09:32 - [INFO] glt-test-suite: ===== System Information =====
2026-08-11 19:09:32 - [INFO] glt-test-suite: Distribution: Ubuntu
2026-08-11 19:09:32 - [INFO] glt-test-suite: Release: 24.04
2026-08-11 19:09:32 - [INFO] glt-test-suite: Architecture: aarch64
2026-08-11 19:09:32 - [INFO] glt-test-suite: Kernel: 7.0.0-2017-nvidia-bos
2026-08-11 19:09:32 - [INFO] glt-test-suite: DMI:
2026-08-11 19:09:32 - [INFO] glt-test-suite: /sys/class/dmi/id/bios_date:Sat May 30 10:47:02 AM UTC 2026
2026-08-11 19:09:32 - [INFO] glt-test-suite: /sys/class/dmi/id/bios_vendor:NVIDIA
2026-08-11 19:09:32 - [INFO] glt-test-suite: /sys/class/dmi/id/bios_version:NV_SBIOS: 06.01.00.00, OEM_SBIOS: 06.01.00.00
2026-08-11 19:09:32 - [INFO] glt-test-suite: /sys/class/dmi/id/board_name:P3809-BMC
2026-08-11 19:09:32 - [INFO] glt-test-suite: /sys/class/dmi/id/board_serial:1331026110247
2026-08-11 19:09:32 - [INFO] glt-test-suite: /sys/class/dmi/id/board_vendor:NVIDIA
2026-08-11 19:09:32 - [INFO] glt-test-suite: /sys/class/dmi/id/board_version:C01
2026-08-11 19:09:32 - [INFO] glt-test-suite: /sys/class/dmi/id/chassis_serial:1942026270153
2026-08-11 19:09:32 - [INFO] glt-test-suite: /sys/class/dmi/id/chassis_type:28
2026-08-11 19:09:32 - [INFO] glt-test-suite: /sys/class/dmi/id/chassis_vendor:NVIDIA
2026-08-11 19:09:32 - [INFO] glt-test-suite: /sys/class/dmi/id/chassis_version:A.0
2026-08-11 19:09:32 - [INFO] glt-test-suite: /sys/class/dmi/id/ec_firmware_release:26.5
2026-08-11 19:09:32 - [INFO] glt-test-suite: /sys/class/dmi/id/modalias:dmi:bvnNVIDIA:bvrNV_SBIOS06.01.00.00,OEM_SBIOS06.01.00.00:bdSatMay30104702AMUTC2026:efr26.5:svnNVIDIA:pnVRNVL72:pvrA.0:rvnNVIDIA:rnP3809-BMC:rvrC01:cvnNVIDIA:ct28:cvrA.0:sku965-XXXXX-XXXX-XXX:pfaMGX:
2026-08-11 19:09:32 - [INFO] glt-test-suite: /sys/class/dmi/id/product_family:MGX
2026-08-11 19:09:32 - [INFO] glt-test-suite: /sys/class/dmi/id/product_name:VR NVL72
2026-08-11 19:09:32 - [INFO] glt-test-suite: /sys/class/dmi/id/product_serial:183XXXXXXXXXX
2026-08-11 19:09:32 - [INFO] glt-test-suite: /sys/class/dmi/id/product_sku:965-XXXXX-XXXX-XXX
2026-08-11 19:09:32 - [INFO] glt-test-suite: /sys/class/dmi/id/product_uuid:9385286f-53fa-9e88-475f-1a02676a0073
2026-08-11 19:09:32 - [INFO] glt-test-suite: /sys/class/dmi/id/product_version:A.0
2026-08-11 19:09:32 - [INFO] glt-test-suite: /sys/class/dmi/id/sys_vendor:NVIDIA
2026-08-11 19:09:32 - [INFO] glt-test-suite: /sys/class/dmi/id/uevent:MODALIAS=dmi:bvnNVIDIA:bvrNV_SBIOS06.01.00.00,OEM_SBIOS06.01.00.00:bdSatMay30104702AMUTC2026:efr26.5:svnNVIDIA:pnVRNVL72:pvrA.0:rvnNVIDIA:rnP3809-BMC:rvrC01:cvnNVIDIA:ct28:cvrA.0:sku965-XXXXX-XXXX-XXX:pfaMGX:
2026-08-11 19:09:32 - [INFO] glt-test-suite: ============================
2026-08-11 19:09:32 - [WARNING] glt-test-suite: Matching kernel perf binary is missing: /usr/lib/linux-tools/7.0.0-2017-nvidia-bos/perf; installed perf may not include events for the running kernel.
2026-08-11 19:09:32 - [WARNING] glt-test-suite: Linux source code is not in /usr/local/share/greenlit/linux.
2026-08-11 19:09:32 - [INFO] glt-test-suite: == Session /var/cache/greenlit/2026-08-11-19:09:32 started ==
2026-08-11 19:09:32 - [INFO] (arm64) test_mpam_resctrl_mount: resctrl already mounted at /sys/fs/resctrl; mount ability confirmed.
2026-08-11 19:14:02 - [INFO] glt-test-suite:
===========================================================

Summary:
	PASSED            : 5
	FAILED            : 0
	Allowed FAILURE   : 0
	SKIPPED           : 0
	WARNED            : 0

Details:
	[PASSED] arm64:test_mpam_resctrl_mount
	[PASSED] arm64:test_mpam_resctrl_schemata
	[PASSED] arm64:test_mpam_l3_occupancy
	[PASSED] arm64:test_mpam_mba_schemata
	[PASSED] arm64:test_mpam_resctrl_info

===========================================================
2026-08-11 19:14:02 - [INFO] glt-test-suite: == Session /var/cache/greenlit/2026-08-11-19:09:32 completed ==
```

### CODE REVIEW

**Comment by nirmoy** [comment]

The PR watcher skips automatic boro reviews for changes with more than 50 commits. This change currently has 91 commits.

```text
review https://github.com/NVIDIA/NV-Kernels/pull/543
```

Head: `c6f7a1cfe4cb`

fyu1:
@ltrager, Is it worth to revert a lot of patches which have been tested by various groups to keep pace to upstream?

1. [arm_mpam: Add MPAM-Fb firmware support v9](https://lore.kernel.org/r/20260807105451.3728284-1-andre.przywara@arm.com) is not supported on Nvidia platform.
2. PBM is not supported on Nvidia platform.
3. control assignment update is mainly cosmetic.
4. others are mainly cosmetic changes.

jamieNguyenNVIDIA:
Reviewed all 82 commits with Claude, one deep-dive pass per commit plus a
whole-series integration pass.

## Downstream-only — introduced by this PR

**1. `57dc56521ca0e` drops the mount error-path cleanup.**

The reverted `1bf2d6c7b5780` had two `disable_abi_playground()` calls; the
re-applied version has one. Missing from `rdt_get_tree()`:

```c
	if (ret && ctx->enable_abi_playground)
		disable_abi_playground();
```

The key is enabled before the first `goto out`, `static_key_enable()` is not
refcounted, and only `rdt_kill_sb()` disables it. So a failed `mount -o
this_is_not_abi` (`-EBUSY` against an existing mount) latches
`resctrl_abi_playground` on for the boot, and a mount that never asked for it
starts accepting `iommu_group:N` in `tasks`. Nothing later restores the hunk.
Deliberate?

**2. `2c2b0824b2838` moves the RIS reset off the width-aware bitmap helper.**

The new `mpam_init_reset_cfg()` sets the feature bit, not just the value:

```c
		mpam_set_feature(mpam_feat_mbw_part, reset_cfg);
		reset_cfg->mbw_pbm = GENMASK(props->mbw_pbm_bits - 1, 0);
```

so `mpam_reprogram_ris_partid()` now takes the single-`u32` arm instead of
`mpam_reset_msc_bitmap()`. Previously `mpam_reset_ris()` passed
`struct mpam_config reset_cfg = {}` with no feature bits, taking the width-aware
path.

`cpbm`/`mbw_pbm` are one `u32` each, but `cpbm_wd` is a 16-bit `MPAMF_CPOR_IDR`
field and `mbw_pbm_bits` comes from `GENMASK(28, 16)`; `cache_has_usable_cpor()`
tests `cpbm_wd <= 32` explicitly. Above 32 portions, are the upper words still
reset? At 65+, is `GENMASK(wd - 1, 0)` an out-of-range shift?

Two side notes on the same commit: the `fract16_to_percent`/`percent_to_fract16`
hunks are unrelated to the subject and unmentioned (they drop `r->membw.min_bw`
to 0, so `bw_validate()` accepts `MB:0=0`), and the changelog lists `mbw_prop`
among the defaulted controls with no `mpam_feat_mbw_prop` case present.

**3. `daa0489dc80b5` — the NRDY `continue` also skips the config capture.**

`cur_flt`/`cur_ctl` are not invalidated by NRDY but get skipped anyway.
`mbwu_state->enabled` is written nowhere else and starts false, so after an
NRDY-skipped first save it stays false with `MSMON_CFG_MBWU_CTL` already written
0 — does `mpam_restore_mbwu_state()` leave the monitor disabled? The counter is
also read *and zeroed* before the test, so the sample is discarded rather than
deferred and `correction` never advances. The base has no NRDY handling in
`mpam_save_mbwu_state()` at all, so this is all new.

**4. `8d82e5debf22b` — `rdt_move_group_iommus()` aborts on the first error.**

It breaks on the first per-group failure and all three callers discard the
return. In `rmdir_all_sub()` (`from == NULL`) the `iommu_matches_rdtgroup()`
filter is bypassed, so a group behind an SMMU without `ARM_SMMU_FEAT_MPAM`
returns `-EIO` and every later group keeps the stale PARTID/PMG — the state the
changelog says this removes. `rdt_move_group_tasks()`, which it says it mirrors,
has no abort path. It also lacks the `resctrl_abi_playground` gate the other two
iommu entry points carry.

## In the MPAM-Fb v9 series — not yet in linux-next

`drivers/resctrl/` in linux-next has no `mpam_fb.c`; this series exists only as
the lore posting, so these could go to Andre before v10 rather than being carried
as downstream deltas.

**5. `52d30f1bd138c` — nothing selects `MAILBOX`.**

`mpam_fb.o` is unconditional in `mpam-y` and calls `mbox_send_message()`, which
needs `CONFIG_MAILBOX` and has no stub; the chain is `ARM64_MPAM_DRIVER` →
`select ACPI_MPAM if ACPI`, a bare bool. arm64 `allnoconfig` +
`CONFIG_ARM64_MPAM=y` fails to link. Shipped config is fine (`MAILBOX=y`) —
`select MAILBOX` would close it.

**6. `4218063d71143` — the PCC channel is freed while the error IRQ is live.**

`mpam_unregister_irqs()` has one caller, `mpam_disable()`.
`mpam_msc_drv_remove()` calls `mpam_msc_destroy()` + `mpam_free_garbage()`, which
runs `pcc_mbox_free_channel()` and `kfree(pcc_chan)` while the devm threaded
error IRQ is still registered — devres frees it only after `.remove()` returns,
and no `suppress_bind_attrs` blocks sysfs unbind. Can
`__mpam_irq_handler()` → `mpam_msc_read_esr()` still reach `msc->pcc_chan` there?

The unbind gap itself is not new — `mpam_msc_drv_remove()` is byte-identical in
linux-next, so `struct mpam_msc` has the same exposure upstream. What v9 adds is
a second object freed in that window.

## In linux-next — mailing list rather than this PR

Both are reproduced faithfully, so they are not backport defects.

**7. `a466792cdf0f2` — `counter_update_class()` tests the candidate, not the incumbent.**

```c
	if (existing_class) {
		if (class->level == 3) {
			pr_debug("Existing class is L3 - L3 wins\n");
			return;
```

The first test reads `class` while the debug string and the `existing_class->level
< class->level` test below both mean the incumbent. An L3 candidate against a
level-255 memory incumbent keeps the memory class; reverse the order and L3 wins.
Which class backs `mbm_total_bytes` therefore follows firmware MPAM table order.
Identical in linux-next (`436d111d918ae`).

**8. `3dd88e6bf6681` — `mbwu_idx_to_mon[]` is not keyed by domain.**

Indexed by CDP-shifted closid and rmid only, ignoring `@d`, while resctrl
allocates counter ids per domain and exposes per-domain assignment via
`mbm_L3_assignments`. Unassigning in one domain reports "Unassigned" in all of
them, and a diverging counter id lets two groups share one hardware monitor.
Also in linux-next

jamieNguyenNVIDIA:
> @ltrager, Is it worth to revert a lot of patches which have been tested by various groups to keep pace to upstream?
> 
> 1. [arm_mpam: Add MPAM-Fb firmware support v9](https://lore.kernel.org/r/20260807105451.3728284-1-andre.przywara@arm.com) is not supported on Nvidia platform.
> 2. PBM is not supported on Nvidia platform.
> 3. control assignment update is mainly cosmetic.
> 4. others are mainly cosmetic changes.

Yes, this does seem a bit risky -- and we're trading SAUCE for a-little-bit-newer SAUCE in many cases.

ltrager:
Thanks for the review. The purpose of this PR is to bring the MPAM/resctrl patch stack in `26.04_linux-nvidia-bos` into alignment with `linux-nvidia-6.18-next`. I noticed some discrepancy when I started to update the [nvidia-kernel-patches-verification](https://github.com/NVIDIA/nvidia-kernel-patches-verification) tool for the patches I added to `linux-nvidia-6.18-next`. I audited the differences between [`linux-nvidia-6.18-next`, `26.04_linux-nvidia-bos`, and this PR](https://docs.google.com/spreadsheets/d/18vYJ5iBSsiqbnLtcmqnUyXP2fLmSNFoKxGQkaE3Yhjo/edit?gid=925578471#gid=925578471). Maintaining newer MPAM/resctrl support in the 6.18 reference kernel than in `26.04_linux-nvidia-bos`, our preferred kernel, creates an avoidable difference between the two patch stacks.

  Some apparently unrelated patches are source prerequisites for a conflict-free replay of the aligned stack:

* Omitting MPAM-Fb entirely causes `b65066184f81` ("arm_mpam: Add debugfs entries to show the MSC/RIS the driver discovered") to conflict because it expects the error-returning MSC accessor API introduced by the MPAM-Fb series. This source dependency exists even if MPAM-Fb is not used at runtime on NVIDIA hardware.
* The PBM fix is required in the source stack. Omitting it causes the following error-propagation patch to conflict in `mpam_reprogram_ris_partid()` because that patch expects the corrected PBM branch orientation. It is also already in Linus's tree as `021118ce5ea9` and has the same patch ID as the copy in linux-nvidia-6.18-next. The path remains feature-gated, so this does not enable PBM on hardware that does not advertise it.
* The counter-assignment update is also more than cosmetic: v5 changes the allocation model to per-MPAM-class counters and depends on the seven generic resctrl prerequisite commits now in Linus's tree. The remaining additions include functional KVM PARTID handling, architecture-version validation, IOMMU-group cleanup, reference-leak fixes, and MBWU state fixes.

I am fine dropping this PR but it will mean the `linux-nvidia-6.18` reference kernel and patching guide will differ from `26.04_linux-nvidia-bos`.

ltrager:
[I compared kernels built from 26.04_linux-nvidia-bos and this PR using the internal vera-perf-check tool.](https://docs.google.com/spreadsheets/d/1Dihi87f9h_Ci-s8yYGJGGjr5qMoJ45rZFG3RX-mlLWI/edit?gid=644212302#gid=644212302) I ran the qualification suite three times on each kernel and averaged the results to reduce run-to-run noise. The same three tests failed consistently on both branches, with no PR-specific failures. Overall performance is effectively unchanged; the MPAM benchmarks trend slightly better, while the remaining benchmarks show only minor differences.

clsotog:

jamieNguyenNVIDIA:
I re-reviewed the current 82-commit head (`01cb4e9840ad`) against the PR merge
base. The findings do not all have the same provenance: four are regressions
introduced by this refresh, two are incomplete or expanded fixes for
pre-existing behavior, and two were already present in the BOS branch.

## Regressions introduced by this refresh

**1. `3199f1b55b018` removes the failed-mount cleanup for
`this_is_not_abi`.**

The merge base calls `disable_abi_playground()` when `rdt_get_tree()` fails;
the re-applied commit drops that cleanup. `enable_abi_playground()` runs before
the `resctrl_mounted` check and `static_key_enable()` is not refcounted. A second
`mount -o this_is_not_abi` therefore enables the global key and then returns
`-EBUSY`, causing the existing mount to accept and display IOMMU-group entries
and report the option even though it was not mounted with it. Please restore
the error-path disable.

**2. `6a927ecb038eac` truncates RIS reset bitmaps wider than 32 bits.**

The base passes an empty `struct mpam_config` to
`mpam_reprogram_ris_partid()`, which selects the width-aware
`mpam_reset_msc_bitmap()` path. The new `mpam_init_reset_cfg()` sets the CPBM
and MBW_PBM feature bits and stores each bitmap in one `u32`, forcing the
single-register path. RIS widths are not clamped to 32 bits, and the retained
KUnit test models a 33-bit reset requiring two writes. Upper bitmap words can
therefore retain stale state. Please keep the width-aware reset path for these
controls.

**3. `13ddf9f68db5bf` leaves an MBWU monitor disabled when its saved value is
NRDY.**

`mpam_save_mbwu_state()` reads `cur_flt`/`cur_ctl`, disables the monitor, reads
and zeros its counter, and then `continue`s on NRDY before saving the
configuration, correction, or `enabled` state. On the first such save,
`enabled` remains false and `mpam_restore_mbwu_state()` skips the monitor; the
sample has also already been discarded. The base always saves the configuration
after disabling the monitor. Please preserve the configuration and enabled
state even when the counter sample is unusable.

**4. `78cd82f7759f57` adds an undeclared `CONFIG_MAILBOX` dependency.**

`mpam_fb.o` is linked unconditionally when `ARM64_MPAM_DRIVER=y` and calls
`mbox_send_message()`, which has no `!CONFIG_MAILBOX` stub. MPAM neither selects
nor depends on `MAILBOX`. I reproduced the resulting valid-config link failure
with arm64 `allnoconfig` plus `CONFIG_ARM64_MPAM=y`:

```text
drivers/resctrl/mpam_fb.o: undefined reference to `mbox_send_message'
```

Enabling `CONFIG_MAILBOX=y` made the same `vmlinux` build pass. The Kconfig
dependency needs to be expressed.

## Incomplete or expanded fixes for pre-existing behavior

**5. `e8f43c4fa58d6` only partially fixes stale IOMMU-group assignments.**

The base does not migrate IOMMU groups during resctrl group removal, so the
underlying stale-assignment problem is pre-existing. The new helper improves
that behavior, but stops on the first per-group error while all three callers
discard its return and continue deleting the resctrl group and freeing its
RMID/CLOSID. During unmount, an unsupported or non-MPAM IOMMU group can fail
before later assigned groups, leaving those later groups with stale PARTID/PMG
values. Either migration must continue past per-group failures, or deletion must
honor the error instead of silently proceeding.

**6. `01cb4e9840ad2` leaves the existing MSC/IRQ lifetime gap open and extends
it to the new PCC resource.**

The base already frees `struct mpam_msc` from `mpam_msc_drv_remove()` while its
devm error IRQ is released only after `.remove()` returns. The refresh adds a
PCC channel reachable from that same IRQ path, and the final commit releases the
channel through `mpam_free_garbage()` without unregistering the IRQ first.
`__mpam_irq_handler()` can therefore race ordinary device removal or sysfs
unbind and access the freed `msc` or PCC channel. `mpam_unregister_irqs()` needs
to run before MSC/PCC destruction on the platform remove path.

## Pre-existing in the BOS merge base

These are real issues, but they are not regressions caused by this refresh.
They should be fixed separately here or in their source series.

**7. `7a7587de5b0e1` preserves the existing candidate/incumbent typo in
`counter_update_class()`.**

The function logs "Existing class is L3" but tests `class->level == 3` instead
of `existing_class->level == 3`. An L3 candidate is discarded when a memory
class is already selected, making the backing class depend on firmware discovery
order. The same bug is present in the PR merge base and in the applied upstream
commit.

**8. `d6fe836e0374a` preserves the existing domainless MBWU monitor mapping.**

`mbwu_idx_to_mon[]` is keyed only by `(closid, rmid)`, while counter assignments
are per `rdt_l3_mon_domain`; `resctrl_arch_config_cntr()` receives `d` but does
not use it when updating the mapping. Assigning or unassigning one domain can
therefore overwrite the hardware-monitor selection used by another domain. The
same global mapping already exists in the PR merge base and in the applied
upstream series.

## Re-review coverage

All 82 current commits map to the previously reviewed series: 80 have identical
stable patch IDs. The other two are the revert/re-apply pair around
`__check_limbo()`; their changed lines are the same and only the newer base
context differs. A no-commit merge of the current target branch was clean,
`git diff --check` passed, and the merged arm64 `vmlinux` build passed with
`CONFIG_MAILBOX=y`.

ltrager:
@jamieNguyenNVIDIA 

**1. `3199f1b55b018` removes the failed-mount cleanup for
this_is_not_abi.**

Fixed in `16ce8427294d4`. The failed-mount path now calls disable_abi_playground() after enabling this_is_not_abi. The resulting patch is patch-identical to the earlier version that contained the cleanup.

**2. `6a927ecb038eac` truncates RIS reset bitmaps wider than 32 bits.**

This was a revert/reapply of `990f4e56511a`, but the reapplication had accumulated additional reset-helper and CMAX/CMIN changes. I rewrote it as `c9419e57dd0aa`; its stable patch ID now exactly matches `990f4e56511a`.

**3. `13ddf9f68db5bf` leaves an MBWU monitor disabled when its saved value is
NRDY.**

Agreed. I removed the downstream NRDY/scaling/initialization workarounds and replaced those commit slots with Ben Horgan’s posted MPAM fixes 5–7: `7f6262e68213f`, `0afd198267bea`, and `486c04d8345f`. They are adapted only for the current guard and error-return interfaces. val remains zero-initialized because __ris_msmon_read() accumulates into it.

**4. `78cd82f7759f57` adds an undeclared CONFIG_MAILBOX dependency.**

Confirmed and reported upstream against MPAM-Fb v9 patch 10/12. The missing dependency is present in the posted LKML patch and was not introduced by conflict resolution in this refresh. The Ubuntu kernel always enables `CONFIG_MAILBOX` so we are not affected.

**5. `e8f43c4fa58d6` only partially fixes stale IOMMU-group assignments.**

I verified that `26.04_linux-nvidia-bos` currently supports assigning IOMMU groups but does not migrate them at all when a resctrl group is removed. The imported migration patch, now `fd33760096a73`, improves that behavior, although its stop-on-first-error handling remains incomplete. Since this refresh is intended to track the latest LKML-posted series closely, I kept the posted patch unchanged. The continue-on-error behavior should be addressed separately in the source series.

**6. `01cb4e9840ad2` leaves the existing MSC/IRQ lifetime gap open and extends it to the new PCC resource.**

I verified that `26.04_linux-nvidia-bos` already destroys `struct mpam_msc` during platform removal before its devm-managed error IRQ is released. The MPAM-Fb PCC work extends that existing lifetime gap to the PCC channel. The reviewed `01cb4e9840ad2` downstream follow-up is no longer present in the current revision.

**7. `7a7587de5b0e1` preserves the existing candidate/incumbent typo in `counter_update_class()`.**

The carried ABMC v5 patch was not modified during this review; `c6f0a1478d952` is patch-identical to its pre-review version. It differs from the older `b072a22e6c35a` implementation in `26.04_linux-nvidia-bos`, although both contain the reported typo. The issue is now fixed separately by applying the exact posted follow-up as `813e9fd0e0d35`.

**8. `d6fe836e0374a` preserves the existing domainless MBWU monitor mapping.**

The carried patch was not modified during this review. The current `19fdb9ba629d5` is patch-identical to upstream commit `6e0e538a75ce`, and the same domainless mapping is present in `26.04_linux-nvidia-bos`. This should be addressed separately in the source series rather than through a downstream change in this refresh.

jamieNguyenNVIDIA:
Re-reviewed the current head (`813e9fd0e0d3`). Four of the eight items from the
previous round are fixed; four remain, and the reset commit has an additional
changelog/content problem. Their provenance differs: items 2 and 4 introduce
new failures into the target tree, item 3 is an incomplete but net-positive
source patch, and items 1 and 5 are already present in the merge base.

Two of the fixes are worth calling out as done properly rather than patched
over: `counter_update_class()` takes Ben Horgan's upstream patch rather than a
local edit, and the MBWU save path was rewritten to route through
`__ris_msmon_read()` instead of special-casing NRDY. I re-reviewed the rewrite
fresh: `__ris_msmon_read()` re-acquires `mon_sel_lock`, writes `CFG_MON_SEL` from
`ctx->mon`, and reads and clears `reset_on_next_read` while holding that lock.
The T241 scale quirk is also applied in the common read path, so dropping the
separate scale fix is consistent with this rewrite.

## Already present in the merge base

**1. `c9419e57dd0aa` no longer does what its changelog describes, and its
percentage conversions are not width-aware.**

This commit is patch-identical to `990f4e56511a` in the merge base, so neither
the conversion behavior nor the changelog mismatch is a net regression caused
by this PR. The conversion changes came from the older source patch; the stale
message and forward-port note are downstream adaptation problems that the
refresh replays unchanged.

Removing `mpam_init_reset_cfg()` fixed the bitmap-width regression from the
previous push. The commit message was not updated, however: its subject and
body still describe building an explicit reset configuration in
`mpam_init_reset_cfg()` and changing `mpam_reprogram_ris_partid()` to apply it.
That helper no longer exists. The actual reset-related change is one line that
adds a cached `cmax` default in `mpam_reset_component_cfg()`.

The rest of the commit replaces `fract16_to_percent()` and
`percent_to_fract16()` with formulas that never use their `wd` argument. This
contradicts the surrounding comments: only the upper `wd` bits are implemented,
the lower bits are ignored, and each value denotes one of `1 << wd` bands. It
also breaks the existing KUnit expectations. For example, with `wd == 1`, the
maximum implemented value is `0x8000`, but the new conversion reports 50 rather
than 100. With `wd == 8`, converting 33% returns non-canonical `0x5479`, outside
the test's expected `[0x5300, 0x5400]` range. The same helper is used for CMAX.

The conversion also makes `get_mba_min()` return zero for every width. For low
widths this changes the minimum from the upper bound of the first hardware band
(50% when `wd == 1`) to 0%, allowing `bw_validate()` to accept `MB:0=0` even
though the hardware value zero still represents that first non-zero band.

Almost all of the behavior named by the source patch's changelog was removed
during its earlier downstream adaptation while the message was retained. That
mismatch is worth correcting when the commit is refreshed, but it should not
be described as newly introduced relative to this PR's merge base.

The `[fenghuay: ...]` note is also inaccurate: `reset_cfg` is zero-initialized
in `mpam_reset_ris()`, not in `mpam_reprogram_ris_partid()`.

Can the conversion changes be reverted or split out and justified together
with updated tests, and can the reset commit's subject, body, and forward-port
note be rewritten to describe the remaining diff?

## Still outstanding

**2. `61549f38bb897` frees the PCC channel before SRCU readers and the error IRQ
are quiesced.**

This commit is patch-identical to the posted source patch, so the defect is
inherited rather than created while applying it downstream. PCC support is not
in the merge base, however, so merging this PR introduces the PCC-specific UAF
into the target tree. A separate downstream lifetime fix present in the
previous PR revision was dropped by this refresh.

The current PCC commit calls `mpam_pcc_chan_put()` as the first operation in
`mpam_msc_drv_remove()`:

```c
static void mpam_msc_drv_remove(struct platform_device *pdev)
{
	struct mpam_msc *msc = platform_get_drvdata(pdev);

	mpam_pcc_chan_put(msc->pcc_chan);

	mutex_lock(&mpam_list_lock);
	mpam_msc_destroy(msc);
	mutex_unlock(&mpam_list_lock);

	mpam_free_garbage();
}
```

If this drops the final reference, `mpam_pcc_chan_release()` frees the PCC
mailbox channel, mutex, and wrapper before the MSC is unlinked and before
`synchronize_srcu()`. An existing or new SRCU reader can still reach that MSC
through the global lists and enter `mpam_fb_send_request()` using the released
channel.

There is a second path through the error IRQ. The devm threaded IRQ remains
installed until devres cleanup after `.remove()` returns. A pending handler can
run `__mpam_irq_handler()` -> `mpam_msc_read_esr()` ->
`mpam_fb_send_request()` after the PCC channel has been released. The driver
does not set `suppress_bind_attrs`, so sysfs unbind can reach this teardown.

The `struct mpam_msc`/IRQ-dev-id part of this lifetime gap predates the PCC
series, but PCC introduces an earlier channel UAF window before the MSC is even
unlinked. Can the per-MSC IRQ be quiesced and the SRCU grace period completed
before releasing the PCC channel, or should runtime unbind be suppressed if it
is not supported?

**3. `fd33760096a73` aborts IOMMU-group migration on the first error.**

This commit is patch-identical to the posted source patch. The merge base does
not migrate IOMMU groups at all, so the underlying stale-assignment problem is
pre-existing and the commit is a net improvement rather than a regression. It
stops short in one place: `rdt_move_group_iommus()` breaks out of the walk on
the first per-group failure, and all three call sites discard its return value.

`arm_smmu_group_set_mpam()` returns `-EIO` when the SMMU lacks
`ARM_SMMU_FEAT_MPAM`:

```c
	if (!(master->smmu->features & ARM_SMMU_FEAT_MPAM))
		return -EIO;
```

In `rmdir_all_sub()` the `from == NULL` unmount path bypasses the
`iommu_matches_rdtgroup()` filter. An unsupported group can therefore fail
ahead of later assigned groups, leaving those later groups with stale
PARTID/PMG values. Teardown then releases their closid and rmid through
`closid_free()` and `free_rmid()`. `rdt_move_group_tasks()`, which the changelog
says this mirrors, has no equivalent abort path.

Should the walk continue after per-group failures, or should group deletion
honor and propagate the error?

**4. `04903184af7d6` adds an undeclared `CONFIG_MAILBOX` dependency.**

This commit is patch-identical to the posted source patch, including the
missing Kconfig dependency. MPAM-Fb is not present in the merge base, so this
is an inherited source defect that nevertheless becomes a new valid-config
build regression in the target tree when this PR is merged.

`mpam_fb.o` is unconditional in `mpam-y` and calls `mbox_send_message()`, which
has no `!CONFIG_MAILBOX` stub. `CONFIG_ARM64_MPAM=y` selects
`ARM64_MPAM_DRIVER`, but neither symbol selects `MAILBOX`.

I reproduced this at the current head with arm64 `allnoconfig` plus
`CONFIG_ARM64_MPAM=y`: the resulting configuration has
`CONFIG_ARM64_MPAM_DRIVER=y` and `CONFIG_MAILBOX=n`, and the vmlinux link fails:

```text
drivers/resctrl/mpam_fb.o: undefined reference to `mbox_send_message'
```

Enabling `CONFIG_MAILBOX=y` makes the same build pass. The target's shipped
arm64 annotations enable both `MAILBOX` and `PCC`, so product deployment is not
currently affected, but this remains a valid-config build failure. The gap is
also present in the posted source series. `select MAILBOX` from
`ARM64_MPAM_DRIVER` would express the unconditional dependency.

## For the mailing list rather than this PR

**5. `19fdb9ba629d5` uses a domain-independent `mbwu_idx_to_mon[]` mapping.**

The index is only `resctrl_arch_rmid_idx_encode(closid, rmid)`.
`resctrl_arch_config_cntr()` receives `d` but does not use it when updating the
mapping, while resctrl owns `d->cntr_cfg` per `rdt_l3_mon_domain` and exposes
per-domain assignments through `mbm_L3_assignments`.

Unassigning an event in one domain can set the global mapping to `-1` while
another domain remains assigned. Assigning different counter ids in different
domains similarly makes the last update select the hardware-monitor index for
all domains. The same domain-independent mapping already exists in the merge
base, and this commit is patch-identical to linux-next. It is therefore both a
pre-existing target-tree problem and a source issue, not a regression caused
by this refresh or a backport-fidelity defect.

Items 2 and 4 introduce new failures into the target tree and look like fixes
needed before merge, even though both defects came from their source patches.
Item 1 is already in the merge base but is worth cleaning up while its commit
is being refreshed. Item 3 needs an explicit decision on whether to carry the
source patch as-is or complete its cleanup behavior here. Item 5 is pre-existing
and belongs on the upstream mailing list.

jamieNguyenNVIDIA:
Re-reviewed the force-pushed head (`b7ad9aa2f1a6`) against the PR base and
the previously reviewed head (`813e9fd0e0d3`). The IOMMU continuation fix is
correct, and the ordinary MPAM-Fb PCC teardown path is now ordered correctly.
The following items still need resolution or clarification before ack. Their
provenance differs: item 1 is an incomplete fix for a pre-existing BOS bug,
item 2 is newly introduced by the downstream fix, item 3 is inherited from
the source series but becomes a new BOS build failure, and item 4 is specific
to this force-push.

## IRQ teardown

**1. `5f5d363b1098` still leaves a per-CPU error IRQ registered while freeing
its MSC.**

The PPI lifetime gap predates this PR: the BOS platform remove path does not
unregister error IRQs. The new commit attempts to fix it, but its PPI branch
calls `free_percpu_irq()` without first disabling the interrupt on every
relevant CPU:

```c
	if (irq_is_percpu(irq)) {
		msc->reenable_error_ppi = 0;
		free_percpu_irq(irq, msc->error_dev_id);
	}
```

The IRQ core requires `disable_percpu_irq()` on each CPU before
`free_percpu_irq()`. If a PPI is still enabled, `__free_percpu_irq()` warns and
returns without removing the action. The MPAM helper then clears
`error_irq_req` and frees the MSC, so a subsequent PPI can reach
`mpam_ppi_handler()` through an `error_dev_id` pointing at the freed object.

The old global teardown satisfied this prerequisite because removing the
CPUHP state invoked `mpam_cpu_offline()` on each online CPU. The new
per-device remove path does not. Can it disable the PPI on all enabled CPUs
while CPUs are locked before calling `free_percpu_irq()`?

**2. `5f5d363b1098` calls the MMIO access helper without holding MPAM SRCU.**

`mpam_unregister_msc_irq()` calls `mpam_touch_msc()` to disable the hardware
error interrupt. For an MMIO MSC, that helper explicitly requires both the
CPU hotplug lock and an MPAM SRCU read-side section:

```c
	lockdep_assert_cpus_held();
	WARN_ON_ONCE(!srcu_read_lock_held(&mpam_srcu));
```

The old `mpam_unregister_irqs()` caller holds both. The new
`mpam_msc_drv_remove()` caller holds only `cpus_read_lock()`, so removing an
enabled MMIO MSC with an error IRQ reaches the warning. This is a regression
introduced by the new downstream fix, not inherited from the source series.

## Inherited source defect that becomes a BOS regression

**3. `48b5dca9879b` still has an undeclared `CONFIG_MAILBOX` dependency.**

This commit is patch-identical to the posted MPAM-Fb source patch, including
the missing dependency, so this was not introduced by downstream conflict
resolution. MPAM-Fb is absent from the merge base, however, so the PR adds a
new valid-config failure to BOS.

I reproduced it at the new head with arm64 `allnoconfig` plus
`CONFIG_ARM64_MPAM=y`. That produces `CONFIG_ARM64_MPAM_DRIVER=y` with
`CONFIG_MAILBOX=n`, and the vmlinux link fails:

```text
drivers/resctrl/mpam_fb.o: undefined reference to `mbox_send_message'
```

Enabling `CONFIG_MAILBOX=y` made the same build pass. The shipped target
annotations enable `MAILBOX`, so the product configuration is not currently
affected, but `ARM64_MPAM_DRIVER` should select its unconditional mailbox
dependency.

## Force-push stack construction

**4. The push is based on an older target commit and includes an unrelated
PCI series.**

The PR base is `165a079e52d1`, but the current head's merge base is
`ab2723ae9159`. The head replays the target's final three commits under new
hashes. Their stable patch IDs and resulting trees match the target, so this
does not cause a content conflict, but it inflates the PR to 94 commits and
adds duplicate history. Can the branch be rebased onto `165a079e52d1`?

The push also adds a revert of the existing MSI-X SAUCE patch followed by five
upstream-derived PCI commits. The final MSI-X behavior was already in BOS, but
the series additionally introduces per-function s390 slot-reset behavior and
generic config-save/FLR accessibility checks. I found no concrete defect in
those patches, but they are unrelated to the MPAM/resctrl refresh, absent from
the PR description, and have no PCI or s390 testing reported here. Can they be
split into a separate PR, or can their inclusion and testing be documented?

## Fixed by this push

- `4a7d534f166e` releases the PCC channel only after the MSC is unlinked and
  the SRCU grace period completes.
- PCC MSCs cannot use PPIs, so `5f5d363b1098` correctly synchronizes their
  hard and threaded IRQ handlers with `devm_free_irq()` before destruction.
- `b7ad9aa2f1a6` continues IOMMU-group migration after errors while preserving
  the first error and balancing iterator references. The previous IOMMU
  early-abort finding can be closed.

One broader lifetime race remains pre-existing in BOS: platform remove is not
serialized with the global `mpam_disable()` teardown, so it can obtain a NULL
or stale `msc`. This was not introduced by the refresh, although PCC adds
another resource reachable through the stale object. The reset-conversion and
domain-independent MBWU mapping issues discussed in the previous review are
also unchanged and pre-existing; I am not repeating them as refresh
regressions here.

For verification, `git diff --check` passed, a merge-tree simulation against
`165a079e52d1` completed without a content conflict, and the arm64 build failed
with `MAILBOX=n` and passed with `MAILBOX=y` as described above.

ltrager:
**1. `c9419e57dd0aa` no longer does what its changelog describes, and its
percentage conversions are not width-aware.**

c9419e57dd0aa is patch-identical to 990f4e56511a7cc55aa0bd21cf159d986c10c4b7 from the PR merge base and carries the same changelog. It only appears in this PR as I had to revert it and then reapply it to cleanly update the patch sets described above. Neither the conversion behavior nor the changelog/content mismatch was introduced by this refresh. I agree that the historical forward-port message no longer describes the adapted diff precisely, but feel rewriting here is out of scope for this PR.

**2. `61549f38bb897` frees the PCC channel before SRCU readers and the error IRQ
are quiesced.**

cherry-picked b7436bb7ceb1262fcaed1d1f33c4b183f168613c which I wrote to fix a similar issue from linux-nvidia-6.18-next. Added an additional SAUCE fix(5f5d363b10985a0c653feb094897c4fa89e63eb0) for the IRQ issue.

 **3. `fd33760096a73` aborts IOMMU-group migration on the first error.**
 
 I added a SAUCE patch(b7ad9aa2f1a6db4f2e48f9588b273a94ba4a109f) that continues walking the remaining IOMMU groups while preserving the first error as the function’s return value. Since advancing kset_get_next_obj() releases the previous iterator reference, the error-path kobject_put() is removed together with the break.
 
 **4. `04903184af7d6` adds an undeclared CONFIG_MAILBOX dependency.**
 
 Upstream agrees to the added dependency. I do not think its worth carrying a SAUCE patch as CONFIG_MALIBOX is enabled for various other devices.

ltrager:
**1. `5f5d363b1098` still leaves a per-CPU error IRQ registered while freeing
its MSC.**

Updated as suggested

**2. `5f5d363b1098` calls the MMIO access helper without holding MPAM SRCU.**

Updated as suggested

**3. `48b5dca9879b` still has an undeclared CONFIG_MAILBOX dependency.**

Reported upstream, no action on this branch as we need CONFIG_MAILBOX for other device drivers.

**4. The push is based on an older target commit and includes an unrelated PCI series.**

 * Rebased branch
 * The PCI commits are part of [v23 PCI series, “Error recovery for vfio-pci devices on s390x”](https://lore.kernel.org/all/20260805165518.794-1-alifm@linux.ibm.com/). Various commits on this branch do depend on at least part of this series. I have taken the entire thing in as was done in `linux-nvidia-6.18-next`

jamieNguyenNVIDIA:
The code findings from the previous round are addressed. One mechanical item
remains before I ACK: PR lint reports that 11 non-SAUCED commits lack the
required `(cherry picked from commit ...)` or `(backported from ...)` trailer.

Six of them already contain a `Link:` identifying the source patch, so their
provenance is present but not in the form recognized by PR lint:

- `b57c95b80946f` — arm_mpam: Apply T241-MPAM-6 to 63-bit counters
- `d1f7dedd4565b` — PCI/MSI: Enable memory decoding before restoring MSI-X messages
- `cedae6c505f4c` — PCI: Fail FLR when config space is inaccessible
- `d71bcc92df22e` — PCI: Avoid saving config space state if inaccessible
- `3eacd72a17fbb` — PCI: Allow per function PCI slots to fix slot reset on s390
- `49575f83a19ae` — PCI: Introduce PCI_SLOT_PLACEHOLDER constant for slot_nr placeholder value

The other five have neither one of the recognized trailers nor a source
`Link:` in their commit messages:

- `4ea833732dc1e` — arm64: mpam: Add memory bandwidth usage (MBWU) documentation
- `107c8654fb99d` — arm_mpam: resctrl: Add resctrl_arch_cntr_read() & resctrl_arch_reset_cntr()
- `21021611a1f61` — arm_mpam: resctrl: Add resctrl_arch_config_cntr() for ABMC use
- `1b7adf1fafbac` — arm_mpam: resctrl: Pre-allocate assignable monitors
- `e636f97151251` — arm_mpam: resctrl: Pick classes for use as MBM counters

Could you add the appropriate recognized provenance trailer to all 11 commits
and rerun PR validation?

ltrager:
Thanks for catching that! Updated and retested on Vera

jamieNguyenNVIDIA:
Re-reviewed the force-pushed head (`a3b088c6b077`) against the previously reviewed head (`bdc58a901843`). The resulting trees are identical, so there are no new code changes. One metadata item remains before ACK.

### Adapted PCI backports

PR validation rejects these two new cherry-pick trailers because the downstream patches were adapted to the target's older PCI slot API:

- `8953c43d8cdf` references `c243e6c470c4`
- `53741062e375` references `dcc5bec09e23`

The source mappings are correct, but these are not exact cherry-picks. Could their trailers use the existing patch URLs instead?

- `(backported from https://patch.msgid.link/20260805165518.794-2-alifm@linux.ibm.com)`
- `(backported from https://patch.msgid.link/20260805165518.794-3-alifm@linux.ibm.com)`

Each also needs a `[ltrager: ...]` note explaining its adaptation to the pre-`102c8b26b54e` PCI slot API. In particular, the second patch folds the required `PCI_SLOT_ALL_DEVICES` definition and matching logic into the backport.

The other nine newly added upstream references validate successfully, including the context-only T241 difference.

After this metadata correction, I have no remaining code findings.

ltrager:
@jamieNguyenNVIDIA - Thanks for the reviews. Updated as suggested.

jamieNguyenNVIDIA:

nvmochs:
@fyu1 Can you please review / ack this PR when you get a chance? Thanks!

fyu1:
The patches look good to me.

Do we need to keep the branch same as upstream? If yes, this branch misses a few more upstream patches:
977f52909c62 guard MBWU state
bb1a0f582d51 suppress_bind_attrs
fc996c39689b NULL msc in remove
9469764292e0 MPAMCFG_PRI reset

None of them is essential though.

Without considering the missing patches,

Acked-by: Fenghua Yu <fenghuay@nvidia.com>

Thanks.

ltrager:
@fyu1 - Thanks for the review!

As discussed in todays meeting we will not be applying those 4 patches.

nvmochs:
Merged, closing PR.

```
c067b7db02a8 (nresolute/nvidia-bos-next) NVIDIA: SAUCE: fs/resctrl: Continue IOMMU group migration after errors
03d926385888 NVIDIA: SAUCE: arm_mpam: Quiesce error IRQ before MSC removal
df5bb6cb78a8 NVIDIA: SAUCE: arm_mpam: Fix PCC channel lifetime during MSC removal
9ab221c8b782 NVIDIA: SAUCE: arm_mpam: resctrl: Correct check that existing class is L3
f422eb0be695 NVIDIA: SAUCE: arm_mpam: Initialize all of struct mon_read in mpam_restore_mbwu_state()
84849fc0531d arm_mpam: Apply T241-MPAM-6 to 63-bit counters
c9ff71c27fa0 NVIDIA: SAUCE: arm_mpam: Use __ris_msmon_read() for saving MBWU state
4b1ea7ddcb07 NVIDIA: SAUCE: fs/resctrl: Fix iommu_group reference leaks when iterating groups
dd55dcbfdcb6 NVIDIA: SAUCE: arm_mpam: Ensure MBWU counters are reset on restore
f98d8adefdff NVIDIA: SAUCE: fs/resctrl: Fix redefinition of iommu helpers with CONFIG_RESCTRL_IOMMU=n
e4bfe60835ca NVIDIA: SAUCE: fs/resctrl: Migrate IOMMU groups when removing resource groups
fbeab1b1dacc NVIDIA: SAUCE: fs/resctrl: Remove unused 'of' parameter from rdtgroup_move_iommu()
0afbe310f241 NVIDIA: SAUCE: fs/resctrl: Fix incorrect PID parsing after IOMMU group token
618cc760827f NVIDIA: SAUCE: arm_mpam: Fix monitor capability and MBM assign for MB events
4c81b1b13be7 NVIDIA: SAUCE: Fix mbm_L3_assign and mon_local_bytes
720e6fef4ac2 NVIDIA: SAUCE: arm_mpam: Include all MSC components during domain setup
3a589a981c03 NVIDIA: SAUCE: resctrl: add MBA max_lim sysfs and visibility from init
aacc6178d0b0 NVIDIA: SAUCE: arm_mpam: resctrl: Determine if any exposed counter can overflow
5d2836ed06ff NVIDIA: SAUCE: arm_mpam: Expose the number of NRDY retries in debugfs
6e489afe1b99 NVIDIA: SAUCE: arm_mpam: Add debugfs entries to show the MSC/RIS the driver discovered
b557e3333c1d NVIDIA: SAUCE: fs/resctrl: Add support for assigning iommu_groups to resctrl groups
cdfdba516e32 NVIDIA: SAUCE: iommu: Add helpers to get and set the QoS state
4897585d8f15 NVIDIA: SAUCE: iommu/arm-smmu-v3: Add mpam helpers to query and set state
bda8b3de5648 NVIDIA: SAUCE: iommu/arm-smmu-v3: Register SMMU capabilities with MPAM
6b2c47025810 NVIDIA: SAUCE: arm_mpam: Allow MSC to be forced to have an unknown location
f35fdf85feca NVIDIA: SAUCE: fs/resctrl: Add this_is_not_abi mount option
43091166358d NVIDIA: SAUCE: arm_mpam: Allow the maximum partid to be overridden from the command line
24b21eb84e52 NVIDIA: SAUCE: fs/resctrl: Avoid a race with dom_data_exit() and rmid_ptrs[]
c6990c353b85 NVIDIA: SAUCE: arm_mpam: resctrl: Add kunit test for ABMC/CDP interactions
79935f6667da arm64: mpam: Add memory bandwidth usage (MBWU) documentation
4269535751ec arm_mpam: resctrl: Add resctrl_arch_cntr_read() & resctrl_arch_reset_cntr()
4996072ffc11 arm_mpam: resctrl: Add resctrl_arch_config_cntr() for ABMC use
edf017a40ba9 arm_mpam: resctrl: Pre-allocate assignable monitors
aad5546db3cb arm_mpam: resctrl: Pick classes for use as MBM counters
b1a5b06351d2 fs/resctrl: Document tasks file behaviour for task id 0 and idle tasks
b7458dd413e9 fs/resctrl: Document that automatic counter assignment is best effort
23f123e0c575 fs/resctrl: Continue counter allocation after failure
739d0d067945 fs/resctrl: Add monitor property 'mbm_cntr_assign_fixed'
c7a5bc79b300 fs/resctrl: Disallow the software controller when MBM counters are assignable
fa6e36cbc10d x86,fs/resctrl: Create 'event_filter' files read only if they're not configurable
76add9e79e38 fs/resctrl: Tidy up the error path in resctrl_mkdir_event_configs()
beb6dfc0e070 NVIDIA: SAUCE: arm_mpam: detect and enable MPAM-Fb PCC support
b1e3387e13ee NVIDIA: SAUCE: arm_mpam: change MPAM-Fb error IRQ to use a threaded IRQ handler
8d83f85137c7 NVIDIA: SAUCE: arm_mpam: Avoid MSC teardown for the SW programming errors
39a34877c8e9 NVIDIA: SAUCE: arm_mpam: add MPAM-Fb MSC firmware access support
e3f5d8747c6f NVIDIA: SAUCE: arm_mpam: prepare mon_sel locking for MPAM-Fb
84087873b116 NVIDIA: SAUCE: arm_mpam: propagate MSC access errors for interrupt control
e334b7cd1228 NVIDIA: SAUCE: arm_mpam: propagate MSC access errors for mpam_reprogram_ris_partid()
fb22a9bc8b25 arm_mpam: Fix MPAMCFG_MBW_PBM register setting
82b22c1d6865 NVIDIA: SAUCE: resctrl/mpam: reset RIS by applying explicit default config
89cbeb49b801 NVIDIA: SAUCE: arm_mpam: propagate MSC access errors for state saving function
d4bc1bcc63d3 NVIDIA: SAUCE: arm_mpam: propagate MSC access errors for __ris_msmon_read()
a8d17c532ab0 NVIDIA: SAUCE: arm_mpam: propagate MSC access errors for msmon helpers
4b481c002299 NVIDIA: SAUCE: arm_mpam: propagate MSC access errors for MBWU counters
f5c62a29c017 NVIDIA: SAUCE: arm_mpam: propagate MSC access errors for hw_probe functions
9f3d9f52f4e7 NVIDIA: SAUCE: resctrl/mpam: probe MPAMF_MBW_IDR MAX_LIM and hardlim_rw
35ab8fd1a0e6 NVIDIA: SAUCE: arm_mpam: let low level MSC accessors return an error
f6a2af42e510 arm_mpam: Update architecture version check for MPAM MSC
10c4b5cb9328 KVM: arm64: Force guest EL1 to use user-space's partid configuration
183b4208b889 PCI/MSI: Enable memory decoding before restoring MSI-X messages
a541e423e106 PCI: Fail FLR when config space is inaccessible
7ce47f743896 PCI: Avoid saving config space state if inaccessible
f073b5c4b648 PCI: Allow per function PCI slots to fix slot reset on s390
cfb4573301fd PCI: Introduce PCI_SLOT_PLACEHOLDER constant for slot_nr placeholder value
c3270ff32f01 Revert "NVIDIA: SAUCE: untested: arm_mpam: resctrl: pick classes for use as mbm counters"
2418d7ad920b Revert "NVIDIA: SAUCE: arm_mpam: resctrl: Pre-allocate free running monitors"
e49d9686fee8 Revert "NVIDIA: SAUCE: arm_mpam: resctrl: Pre-allocate assignable monitors"
3e0b6ea6d3fd Revert "NVIDIA: VR: SAUCE: arm_mpam: resctrl: Add kunit test for ABMC/CDP interactions"
79704ebf4e5d Revert "NVIDIA: SAUCE: arm_mpam: resctrl: Add resctrl_arch_config_cntr() for ABMC use"
cba6a484f774 Revert "NVIDIA: SAUCE: arm_mpam: resctrl: Add resctrl_arch_rmid_read() and resctrl_arch_reset_rmid()"
ba765f078f26 Revert "NVIDIA: SAUCE: arm_mpam: resctrl: Add resctrl_arch_cntr_read() & resctrl_arch_reset_cntr()"
0b23a56a1d9e Revert "NVIDIA: VR: SAUCE: fs/resctrl: Avoid a race with dom_data_exit() and rmid_ptrs[]"
26c7f7dc457e Revert "NVIDIA: VR: SAUCE: arm_mpam: Add debugfs entries to show the MSC/RIS the driver discovered"
5d1abe78beea Revert "NVIDIA: VR: SAUCE: arm_mpam: Expose the number of NRDY retries in debugfs"
d3c76bddf927 Revert "NVIDIA: VR: SAUCE: arm_mpam: Split the locking around the mon_sel registers"
c62f3eb8e2bc Revert "NVIDIA: VR: SAUCE: arm_mpam: Allow the maximum partid to be overridden from the command line"
2b468ca3bdef Revert "NVIDIA: VR: SAUCE: arm_mpam: Allow MSC to be forced to have an unknown location"
de4b032bb581 Revert "NVIDIA: VR: SAUCE: fs/resctrl: Add this_is_not_abi mount option"
55006bd898c2 Revert "NVIDIA: VR: SAUCE: iommu/arm-smmu-v3: Register SMMU capabilities with MPAM"
571fa1f9c26b Revert "NVIDIA: VR: SAUCE: iommu/arm-smmu-v3: Add mpam helpers to query and set state"
499a71243225 Revert "NVIDIA: VR: SAUCE: iommu: Add helpers to get and set the QoS state"
d39ef1e0ff4e Revert "NVIDIA: VR: SAUCE: fs/resctrl: Add support for assigning iommu_groups to resctrl groups"
905b0917a68a Revert "NVIDIA: SAUCE: arm_mpam: resctrl: Determine if any exposed counter can overflow"
bcf92b7141dc Revert "NVIDIA: VR: SAUCE: arm_mpam: Avoid MSC teardown for the SW programming errors"
9b272d7a43ea Revert "NVIDIA: SAUCE: arm_mpam: Include all associated MSC components during domain setup"
1160aae43afd Revert "NVIDIA: SAUCE: resctrl/mpam: reset RIS by applying explicit default config"
8197a2bad9aa Revert "NVIDIA: VR: SAUCE: resctrl/mpam: probe MPAMF_MBW_IDR MAX_LIM and hardlim_rw"
9ff2103ba61a Revert "NVIDIA: VR: SAUCE: resctrl: add MBA max_lim sysfs and visibility from init"
b982fe85001f Revert "NVIDIA: SAUCE: Fix mbm_L3_assign and mon_local_bytes"
396491524c3f Revert "NVIDIA: SAUCE: arm_mpam: Apply T241-MPAM-6 to 63-bit counters"
99525bf39357 Revert "NVIDIA: SAUCE: PCI/MSI: Enable memory decoding before restoring MSI-X messages"
```

PR #560, opened 2026-09-02
[26.04_linux-nvidia] NVIDIA: SAUCE: ASoC: sdw_utils: clear stale device references in exit hooks

kbutala-nv:
asoc_sdw_rt_amp_exit() drops the references held in ctx->amp_dev1 and ctx->amp_dev2, but leaves the pointers populated. If the card cleanup path runs the exit hook more than once, the same device references are dropped again, which can underflow the device refcount and leave stale pointers for later SoundWire bus walks.

In practice this can show up as refcount underflow warnings followed by an oops when later probe, driver bind, or suspend paths walk the SoundWire bus and touch the freed device.

This first part is fixed upstream via change 046173b98de316211b81d9909cde81ca582604a8

Clear the amplifier device pointers after put_device(), matching the idempotent pattern already used by the RT SDCA jack cleanup path. Do the same for the RT711 headset codec device pointer.

This second part is covered through second patch under this PR. 

This makes the exit hooks safe to call after deferred or failed card probe paths and avoids stale device references being reused after cleanup.

This is tracked in internal bug 6290103

LP: https://bugs.launchpad.net/ubuntu/+source/linux-nvidia/+bug/2166188

PR #566, opened 2026-09-03
[26.04_linux-nvidia] MediaTek MT8901 audio: SoundWire/ASoC/SOF stack

dcemin-nv:
BugLink: https://bugs.launchpad.net/ubuntu/+source/linux-nvidia/+bug/2166301

MediaTek mt8901 audio stack for MT8901 ACPI laptops, as validated on the reference board and the supported laptop designs: the MediaTek SoundWire manager, the ASoC AFE platform and machine drivers with SDCA support, the SOF/ADSP platform driver, the reliability and enumeration fixes, and the machine table entries for the six validated codec topologies.

Series shape (reworked per review):
- 30 commits, dependency ordered: the MediaTek drivers, a mechanical move of the SoundWire manager into drivers/soundwire/mediatek/, the reliability and enumeration fixes, the six machine table entries, the two system sleep commits (each complete on its own), a fix for a left/right swap on the jack PCM after rapid stop/start cycles (PDI FIFO reset in prepare), and one UBUNTU: [Config] commit at the tail carrying the final annotations. Review fixups are folded into the commits they repair with Co-developed-by attribution. Round 7 (2026-09-14): internal platform, board and vendor names removed from the commit messages, the config annotation notes and four code comments; no functional change.
- All commit messages rewritten as kernel prose; no internal metadata, tracker references, or device identifiers remain; config commits use the UBUNTU: [Config] classification, with the options introduced disabled and enabled by the final commit after the complete stack.
- Every commit passes scripts/checkpatch.pl --strict --git with zero errors. Remaining strict-mode output: six "does MAINTAINERS need updating?" warnings for new files (downstream tree, not applicable) and a handful of style checks in the imported MediaTek code (CamelCase identifiers, macro argument reuse); left untouched to keep the series byte identical with the board validated content.

Dependencies (patches intentionally NOT carried here):
- MediaTek power_wrap driver and client API: the clock control and SOF commits call mtk_send_power_control_req() / mtk_pwrap_dev_probe(). now open as #570; merge #570 first by its owners.
- sdw_utils double put fix (mainline 046173b98de3) for the shared rt1320/rt1321 DAI name: PR #560, first commit; merge #560 before this PR.
- MediaTek pinctrl bus-keeper support: the pin bias commit uses mtk_pinctrl_program_bias_by_gpio(), submitted in PR #564.
This PR does not build standalone until #564 and #570 land (the MT8901 symbols depend on MTK_POWER_WRAP and PINCTRL_MTK_PARIS at Kconfig level, so they simply vanish without them).

Upstream status: the NVIDIA authored generic fixes from this program are on the lists as of 2026-09-12: the SoundWire sdw_utils fix (ASoC: sdw_utils: clear stale RT711 device reference on exit, https://lore.kernel.org/r/20260912182131.1156475-1-dcemin@nvidia.com, Cc stable) and the related sbsa_gwdt fix carried by #568/#569 (https://lore.kernel.org/r/20260912182107.1156221-1-dcemin@nvidia.com). The commits in this series are MediaTek drivers plus fixes to them; they stay SAUCE until MediaTek posts the drivers, which is being coordinated with MediaTek (asked for their target tree and timeline on 2026-09-12) and their postings will be linked here when available.

Validation (this exact content, as an NV-Kernels build on the MT8901 reference board): full kernel deb boots with the board's own config; audio stack enumerates with the correct machine identity; acoustic loopback 4/4 PASS at 61 to 67 dB SNR; 3/3 mode verified deep S3 cycles; loopback passes after resume. The audio and config subtree of this PR is byte identical to the validated series plus the round 3 review fixes (verified by diff); the complete PR tree is not, because the prerequisite paths were moved to PRs 564 and 570. The validation build included those prerequisites underneath.

Please use a rebase merge, not a squash.


dcemin-nv:
@kbutala-nv @aritger review requested per the patch integration call for action (Sudeep asked for Andy on anything touching shared subsystems). Bharat Dintakurti should review the SOF DSP power management patch on his return; carrying it here since he is out.

nirmoy:

The PR watcher skips automatic boro reviews for changes with more than 50 commits. This change currently has 58 commits.

```text
review https://github.com/NVIDIA/NV-Kernels/pull/566
```

Head: `1db476381097`

jamieNguyenNVIDIA:
Please follow the conventions defined in the patch [submission guide](https://nvidia.atlassian.net/wiki/spaces/DPS/pages/2231328936/Linux+NVIDIA+Kernels+Development+and+Patch+Submission+Guide?xpis=eyJicmlkZ2UiOiJxdWlja0ZpbmQiLCJpZCI6IjE3ODgzOTg5OTg2MDIiLCJzb3VyY2UiOiJjb25mbHVlbmNlIn0%3D).

In particular:

- All commits need sign-offs
- All internal information (bug IDs, JIRAs, Change-ID, etc, "Found by CodeRabbit review on noble MR 69", etc.) should be stripped
- Downstream commits need to be prefaced with "NVIDIA: SAUCE:".  **Also, what is the plan for upstreaming these?**
- Upstream commits (if there are any) need "backported from" or "cherry picked from" trailers
- There are a couple instances of commit1-revert-commit2 -- this should be squashed to keep our history clean

dcemin-nv:
Thanks for the review. Updated the series to follow the guide:

1. Every commit now carries a Signed-off-by (submitter sign off added where the source tree lacked one).
2. Internal metadata stripped (change IDs, internal bug and change numbers, internal tracker references).
3. All commits are prefaced with NVIDIA: SAUCE: (the whole series is downstream, no upstream cherry picks, which also resolves the two "cannot resolve upstream SHA" lint errors: those were internal change references, now removed).
4. The add plus revert pairs are squashed out; the series is now 58 commits with a clean history.

On the upstreaming plan: the MediaTek authored base drivers (SoundWire manager, ASoC AFE and machine drivers, SOF platform) are MediaTek IP and mainline submission is expected to come from MediaTek, with coordination in progress (their rt712 codec fix from this program is already accepted upstream). The NVIDIA authored generic pieces go to mainline from our side: the sdw_utils fix is being prepared for alsa-devel and the watchdog change for linux-watchdog (a related parameter rename from this program was already accepted in the watchdog maintainer tree). The mt8901 specific machine table entries and fixes follow the MediaTek base drivers whenever those land upstream.

nirmoy:
@dcemin-nv I created the Launchpad bug for this change:
https://bugs.launchpad.net/ubuntu/+source/linux-nvidia/+bug/2166301

Could you please add this link to the PR description?

debian.nvidia/config/annotations
```diff
@@ -156,6 +156,12 @@ CONFIG_MTD                                      note<'Essential for boot on ARM6
 CONFIG_MTK_PCIE_HOTPLUG                         policy<{'arm64': 'm'}>
 CONFIG_MTK_PCIE_HOTPLUG                         note<'CX7 PCIe hotplug driver for NVIDIA DGX Spark systems with GB10 SoC.'>
 
+CONFIG_MTK_POWER_WRAP                           policy<{'arm64': 'y'}>
+CONFIG_MTK_POWER_WRAP                           note<'MediaTek Power Wrap device-control (SCMI) driver.'>
+
+CONFIG_MTK_SSPM_CI                              policy<{'arm64': 'y'}>
+CONFIG_MTK_SSPM_CI                              note<'MediaTek SSPM control interface used by the Power Wrap SCMI device-control path.'>
+
```
nirmoy: @dcemin-nv Please split it into separate patch as annotation changes are limitted to Canonical distro. ref: https://github.com/NVIDIA/NV-Kernels/commit/09c0ae5c6e47b25ba18001173be220cece00c5cd

There are below headings in some commits:
```
[Description]
[Solution]
[Test]
```
Those looks odd for linux kernel commits and should be removed. 
V2, v3 changelog  seems unnecessary as this is the 1st time we are seeing the patch.

jamieNguyenNVIDIA:
Thanks for the cleanup. I rechecked the new 58-commit series at `1db476381097`. The remaining add/revert chain is gone and every commit now has a `Signed-off-by` line. Several commit-hygiene issues remain:

- `9ffe74fa3162` changes 290 lines across the power-wrap API and sysfs/debug interface but has no explanatory commit message. It also reintroduces an internal `Change-Id:` trailer. Please write a durable rationale and split the sysfs/debug interface from the API where appropriate.
- `9ffe74fa3162` is authored by Kaushal Butala and `7138887f8972` by Sanket Diwate, but each contains only David Cemin's sign-off. Please confirm and preserve the original authors' DCO/provenance chain.
- The SoundWire PM history still records a known-wrong intermediate change: `20bb553acfd7` applies the incorrect master-wide power-vote assumption, then `eb4400dd49e2` reverses it. Fold the correct behavior into `852107f97783`. `eb440...` also has a stale `Fixes: ffce0bf0d541` reference which does not exist in this history.
- Internal/source-tree information remains despite the cleanup note: eight commits still mention Gerrit or `MR !114`, `9ffe74fa3162` contains a `Change-Id:`, and `f0b809ca211a` includes device serial `PF68ES30`. Seven commits also retain FastOS-specific test boilerplate. Please convert source-review history into durable rationale and remove private identifiers.
- Four config commits use the double prefix `NVIDIA: SAUCE: UBUNTU:`. Please use the project-standard classification for these commits rather than stacking both prefixes.
- The upstreaming plan described on PR #567 is useful, but planned submissions are not traceable public provenance. **Please get the applicable patches posted upstream first so the downstream commits can carry `(backported from <lore link>)` trailers.** For the MediaTek-owned base drivers, please coordinate with MediaTek and link their postings once available; for the NVIDIA-owned generic fixes, please post them to the appropriate lists before carrying them downstream.
- The intermediate history is not self-contained/buildable:
  - `2cf468ff9c4f` and `2cedb2ad034b` consume the power-wrap API before it is introduced by `9ffe74fa3162`.
  - `6da1e1973ed4` includes the already-removed `pwrap_public.h`.
  - `0f3fe6c2ecbb` enables the SOF driver before `6fb247a3c9e0` repairs its include/Kconfig/PM integration.
- `306bb67708ee` claims to enable "audio and ADSP" but does not touch `CONFIG_SND_SOC_SOF_MT8901`. Retitle/reorder it and place final enablement after the complete stack.
- `17e0fb9b3e24` is patch-identical to `b94410ff5376` in PR #564. Drop it from one PR and express the dependency explicitly.
- Please fold review/fixup commits into the commits they repair, particularly `b8c8d79eee52` and `0171ed973db9`, which are explicitly omnibus "review fixups." There are numerous other fix-on-introduction chains throughout the series that should be cleaned up before merge.
- Several commits combine unrelated or mechanical work:
  - `a444cca5a88b` combines SSPM transport, power-wrap core, and sysfs/debug code.
  - `2cedb2ad034b` mixes a feature with seven 100%-rename file moves.
  - `9ffe74fa3162` mixes the power-wrap API with a new sysfs/debug interface.
  - `0171ed973db9` contains unrelated build/include and runtime-PM fixes.
- Twenty-one messages still use imported `[Description]`/`[Solution]`/`[Test]` templates, and 31 subjects exceed 75 characters after adding the prefix. Please rewrite these as normal kernel commit messages.
- Please run `scripts/checkpatch.pl --strict --git <commit>` against every commit after rebuilding the series, and address or explain each remaining warning. A final-tree check does not catch commit-local metadata and formatting problems.
- Minor, but helpful for reviewers: please update the stale PR description. It still says 64 commits, claims “the first 9” are prerequisites, and says two add/revert pairs remain; none of those accurately describes the current series.

Please rebuild this as a clean, dependency-ordered series whose individual commits are reviewable and buildable, update the PR description, and rerun validation.

nirmoy:
## BaseOS Kernel Review

> [!TIP]
> ### ✅ Review passed
>
> **Findings:** none

### 🔍 Review artifacts

- 🟢 **Boro review:** [no findings](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/566/latest.html)
- 🔵 **Test plan:** [ready for validation](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/566/test-plan.html)

- 🟢 **`amd64`:** [build report and 4 debs](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/566/debs/amd64/)
- 🟢 **`arm64`:** [build report and 4 debs](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/566/debs/arm64/)

> [!NOTE]

<details>

- Reviewed head: `d3ebf136fc2f`
- Overall status: **passed**

</details>

dcemin-nv:
Thanks for the thorough review; the series is rebuilt accordingly and force pushed (now 33 commits). Point by point:

1. power_wrap and BusKeeper prerequisites: dropped from this series entirely. The power_wrap driver, client API (the 290 line commit) and its feature patches go through their own submission by their authors, which also resolves the authorship/DCO concern for the Kaushal and Sanket commits and the Change-Id/rationale issues there. BusKeeper is in PR #564 (dropped the duplicate here). The commits that consume those APIs carry explicit dependency notes; this PR does not build standalone until the dependencies land.
2. Power vote intermediate: folded. The master-wide vote change and its correction are squashed into the system sleep PM ops commit, which now introduces the per-link vote behavior directly with the rationale in the message.
3. Review fixups and fix-on-introduction chains: folded into the commits they repair with Co-developed-by attribution (manager, AFE platform, machine driver, clock control, PM ops, suspend/resume, bus reset recovery). The suspend/resume omnibus fixup was split by subsystem into its two parents.
4. Internal information: all Gerrit/MR references, Change-Id trailers, the device serial, internal template blocks and internal codenames are gone; messages are rewritten as kernel prose. The Change-Ids in the previous push turned out to be added at commit time by a local Gerrit hook, now bypassed.
5. Config commits: reclassified as NVIDIA: [Config], options introduced disabled, single final enablement commit at the end covering the complete stack (retitled to what it actually enables).
6. Ordering/buildability: SOF config now follows the SOF driver (whose integration repair is folded into it, includes corrected at introduction); annotations changes ride with their config commits; subjects are all within 75 characters.
7. checkpatch --strict --git on every commit: zero errors across all 33. Remaining strict output is six "does MAINTAINERS need updating?" warnings for new files (downstream tree) and style checks in the imported MediaTek code (CamelCase, macro argument reuse), left as is to keep the series byte identical with the board validated content; happy to address those as follow ups if you prefer.
8. Upstream first: the NVIDIA authored generic fixes are being posted (sdw_utils against broonie/for-next, sbsa_gwdt to linux-watchdog); I will add the lore links here once archived. Mainline submission of the MediaTek base drivers is being coordinated with MediaTek and their postings will be linked when available.
9. PR description rewritten to match the current series.

The rework preserved the validated tree byte for byte (verified by diff against the board tested series), so the board validation results carry over unchanged.

arch/arm64/configs/defconfig
```diff
@@ -1112,6 +1112,7 @@ CONFIG_SND_SOC_SOF_OF=m
 CONFIG_SND_SOC_SOF_MTK_TOPLEVEL=y
 CONFIG_SND_SOC_SOF_MT8186=m
 CONFIG_SND_SOC_SOF_MT8195=m
+CONFIG_SND_SOC_SOF_MT8901=m
 CONFIG_SND_SUN8I_CODEC=m
 CONFIG_SND_SUN8I_CODEC_ANALOG=m
 CONFIG_SND_SUN50I_CODEC_ANALOG=m
```
nirmoy, line 1118: Sorry for the confusion. I meant only debian.*/config/annotations files changes should be split into another commit. You can remove this diff and move it back to the previous commit.

nirmoy:
@dcemin-nv There is a build failure for this PR please take a look https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/566/debs/arm64/build-error.txt

Update:
>This PR does not build standalone until those land; the affected commits carry the dependency note.

I just saw above comment in the PR description so this build failure is expected.

jamieNguyenNVIDIA:
Thanks, this addresses most of my previous review. I rechecked the 33-commit series at `98d4771cddd2`; the DCO chains, internal metadata, templates, revert/fixup history, duplicate prerequisite, SOF ordering, subject lengths, and PR count are now cleaned up.

In addition to the unbuildable head already called out in the PR description and by Nirmoy, I found the following remaining series-hygiene items:

- `15548abc952a` selects `SOUNDWIRE_MTK_MT8901_CLK`, but that symbol is not defined until the following commit, `a41ebed20f2f`. Please move the selection or definition so the Kconfig state is valid at every commit.
- `a41ebed20f2f` still mixes the clock-control feature with nine mechanical moves into `drivers/soundwire/mediatek/` (six are 100% renames). Please create the manager files in their final directory in `4127214c791d`, or split the relocation from the feature change.
- **What is the concrete upstreaming plan for these patches?** The postings are still pending. Please get the applicable patches posted, then add the lore URLs and `(backported from <lore link>)` trailers before they are merged downstream.

Minor cleanup:

- A fresh per-commit `checkpatch.pl --strict --git` run reports two commit-message wrapping warnings not listed in the PR description: `b939ae55b7d8` and `0ce3c4664558`. The former also contains the malformed phrase `S/N the affected unit`.
- Minor PR-description corrections would help reviewers: it says the config commits use `NVIDIA: [Config]`, while the actual subjects use `UBUNTU: [Config]`; and the retained audio/config subtree is byte-identical to the validated series, but the complete PR tree is not because prerequisite-owned paths were removed. Please scope the “byte for byte” statement accordingly and state that validation included the separately supplied prerequisites.

jamieNguyenNVIDIA:
In addition to the unbuildable head already called out in the PR description
and by Nirmoy:

1. `4127214c791d` uses `char name[32]` for firmware-property names that need
   33 and 36 bytes including the terminating null. The driver silently queries
   truncated interface-revision and clock-stop keys, leaving those properties
   unset. A `W=1` build reports both truncations.

2. The same commit applies the default `max_clk_freq` only after
   `sdw_bus_master_add()`. By then the zero value has already been copied into
   `bus->params.max_dr_freq` and `curr_dr_freq`, so firmware which omits both
   clock properties cannot allocate bandwidth for its first stream.

3. `98c5a9236aae` unconditionally calls
   `snd_soc_acpi_sdw_link_slaves_found()`, but `SOUNDWIRE_MTK` neither selects
   nor depends on `SND_SOC_ACPI`. The legal configuration
   `SOUNDWIRE_MTK=m`, `SND_SOC_ACPI=n` leaves that symbol unresolved at
   modpost.

4. `54a244a7eded` enables every bit in `AFE_IRQ_MCU_EN`, while the normal
   handler services only the memory-interface sources on bits 0 through 15.
   An asserted enabled source on bits 16 through 21 is never acknowledged and
   can continuously retrigger the handler. A stale mapped memory-interface
   interrupt can also call `snd_pcm_period_elapsed()` before a substream is
   installed.

5. `7d38ff2257fd` waits 30 ms only after
   `mtk_sdw_update_slave_status()` has already performed the first Dev0
   enumeration. The deterministic mixed response described by the commit can
   therefore still pass both verification reads and register a phantom slave.
   The settling delay needs to precede the first enumeration attempt.

6. `272c9c32b384` resets and reprograms the SoundWire controller without
   serializing against `bus->msg_lock` or another reset. It can race an active
   `sdw_transfer()` or a concurrent recovery/debugfs reset.

7. Automatic recovery in that commit does not restore
   `MCP_SLAVEINTMASK0/1` after `SOFT_RST`. It restores only the aggregate slave
   interrupt bit, so later attach, detach, and alert transitions remain
   masked and enumeration can stall.

8. `b939ae55b7d8` reads the slave list and statuses without `bus_lock`, decides
   that nothing attached, and only then quiesces status work. Status processing
   can complete an attachment in between, after which this path resets a
   working link based on its stale snapshot. It should quiesce the work first
   and recheck under the lock.

9. `2c8bbd73d87b` advertises a SOF platform driver but creates only the
   power-wrap client and enables runtime PM. It has no `sof_dev_desc`, calls no
   SOF probe/remove entry point, and selects none of the common MediaTek or
   ACPI SOF support. No `snd_sof_dev` is created, so firmware, IPC, topology,
   and SOF audio are never initialized.

10. `4df8781be92f` cannot correctly roll back a partial runtime-suspend
    failure. Both clock-disable helpers continue after individual firmware
    failures and return only the last error, so the PM core can leave the
    device runtime-active with some clock resources already disabled.

11. The interrupt-request failure in that commit bypasses `err_pm_put` after
    `pm_runtime_resume_and_get()`. It disables resources that were not acquired
    by that path while leaking the runtime-PM reference, MTCMOS vote, and
    register-access clock that were acquired.

12. The same probe converts a non-transient initial
    `pm_runtime_put_sync()` failure into success. The PM core retains
    `runtime_error`, so a later PCM open fails its runtime get with `-EINVAL`
    until the device is reprobed.

13. `cf704cb3db46` restores manager data-port state only when
    `mdai->suspended` was set by a suspend trigger. A configured but idle PCM
    receives no such trigger; controller suspend still wipes its port/PDI
    state, and both prepare paths short-circuit after resume. Nothing restores
    the configuration before the PCM is started.

14. Link-one suspend in that commit releases the source-clock and core-voltage
    votes independently, but records only the last error. If one release
    succeeds and the other fails, the error path reinitializes the live
    controller without reacquiring the successfully released vote.

15. `0ce3c4664558` initializes `intmask_lock` only from
    `mtk_sdw_core_init()`, after the shared interrupt handler has already been
    registered. A pending or persistent clash interrupt can reach the burst
    threshold and take this lock before initialization. The core and lock need
    to be initialized before the handler is published.

dcemin-nv:
Thanks Jamie, both comments addressed. New heads: 566 = 5f66f519d7ea,
567 = 31c170c66306, 35 commits each (33 plus a mechanical move commit
and a Kconfig dependency commit, see below). The audio and config
subtree is byte identical to the previous heads plus the fixes listed
here; every commit re-ran checkpatch --strict (0 errors, 0 warnings,
only pre-existing style CHECKs), the tip and the new move commit were
cross compiled with W=1 on top of the 564 and 570 prerequisites.

Series hygiene:

- Kconfig valid at every commit: the soundwire clock control commit now
  precedes the ASoC clock control commit that selects
  SOUNDWIRE_MTK_MT8901_CLK (the two touch disjoint trees, so this is a
  pure reorder).
- File moves split from the clock feature: a new commit "soundwire:
  mediatek: move the manager into its own directory" (pure renames plus
  the Makefile hook, no functional change) lands right before the clock
  commit, which is now feature only.
- Commit message wrapping fixed on the two flagged commits, the "S/N the
  affected unit" phrase is gone.
- The arch/arm64 defconfig hunk moved from the final [Config] commit
  into the SOF commit (Nirmoy's inline).
- PR description corrected as requested (UBUNTU: [Config], scoped byte for
  byte statement, prerequisites stated for validation).

Functional findings, in your numbering:

1. Fixed. Property name buffer widened (the two long MIPI keys were
   silently truncated). A second buffer of the same class in the machine
   glue (acpi-scd-N-subproperties) fixed at the same time.
2. Fixed. The bus property defaults now live at the end of
   mtk_sdw_read_prop(), before sdw_bus_master_add() derives the bus
   data rate limits.
3. Fixed by the new tail commit: SOUNDWIRE_MTK now depends on
   PINCTRL_MTK_PARIS (the pinctrl helper from PR 564 is also an exported
   symbol without a Kconfig link) and selects SND_SOC_ACPI. Resolved
   config on arm64 is unchanged.
4. Partially agree. AFE_IRQ_MCU_EN now enables exactly the 22 IRQ
   sources the driver describes (GENMASK(MT8901_AFE_IRQ_NUM - 1, 0))
   instead of all 32 bits, so nothing can be routed that the handler
   cannot acknowledge. Two parts of the finding do not hold on this
   driver: the handler services every acquired IRQ 0 to 21 through
   memif->irq_usage, not only 0 to 15, and memif->substream is set in
   startup before the IRQ is acquired and cleared in shutdown after it
   is released, so period_elapsed cannot run without a substream.
5. Fixed. The settle delay now precedes every enumeration round
   including the first, gated on Dev0 being attached at entry.
6. Fixed. mtk_sdw_core_bus_reset() holds bus->msg_lock for the whole
   soft reset and re-init, so it cannot overlap an sdw_transfer(). The
   debugfs deep reset path takes the same lock around its hw_init.
   Reset versus reset was already serialised (both other callers cancel
   the status work first); the change closes the transfer overlap.
7. Fixed. The reset restores MCP_SLAVEINTMASK0/1 when interrupts are
   live (the status work path). The quiesced callers still get the masks
   back from their enable_irq(true).
8. Fixed. The attach check quiesces first, then re-reads the slave
   states under bus_lock, and re-arms interrupts on every early return.
9. Not a bug in the code, a wrong commit message. That driver is
   intentionally the ADSP power management scaffold: it binds the
   NVDA8301/8302/8303 DSP devices and manages their domain vote through
   power_wrap so the DSP domain can reach the deep platform sleep state.
   No SOF core, firmware, IPC or topology is set up yet. Subject and
   body rewritten to say exactly that ("ASoC: SOF: mt8901: add ADSP
   power-management driver").
10. Fixed. The three clock disable helpers are transactional: on
    failure they re-enable what they already released and return the
    error, so runtime suspend's unwind sees a consistent state.
11. Fixed. The IRQ request failure now goes through err_pm_put and no
    longer disables clocks that path never enabled.
12. Fixed. A latched failure of the initial pm_runtime_put_sync() fails
    the probe. Transient -EAGAIN/-EBUSY keep the previous behaviour
    (device stays runtime active, regcache stays write through) since
    nothing is latched in that case.
13. Fixed. Controller suspend flags every configured DAI as suspended
    (not only the ones that saw TRIGGER_SUSPEND), and hw_free clears the
    flag and the PDI count, so an idle prepared stream is rebuilt by
    prepare after resume.
14. Fixed. The link 1 vote release is transactional: if VCORE release
    fails, SRCCLKENA is reacquired before returning the error.
15. Fixed. The core software init (lock, work items, ops) now runs
    before devm_request_irq() in probe. This predates the clash storm
    commit, so the fix is in the manager commit.

On the upstreaming plan, plainly: the MediaTek drivers stay NVIDIA: SAUCE until MediaTek posts them to the lists; that coordination is with MediaTek and the lore links will be added to the downstream commits when their postings exist. The two NVIDIA authored patches (the sdw_utils fix for the ASoC list and the sbsa_gwdt fix for linux-watchdog) go to the lists this week, and the lore links follow on the corresponding PRs.

jamieNguyenNVIDIA:
I rechecked the current 35-commit series at `5f66f519d7ea`. Most of the
previous findings are fixed, but these issues remain:

1. `c447dbce102f` still enables IRQ sources 0 through 21 in
   `AFE_IRQ_MCU_EN`, while `mt8901_afe_irq_handler()` only iterates the 16
   memifs, whose fixed IRQ mappings are 0 through 15. An enabled source on
   bits 16 through 21 is therefore still never acknowledged on the normal
   read-success path and can continuously retrigger the threaded handler.
   Changing `0xffffffff` to `GENMASK(21, 0)` avoids reserved bits, but it does
   not fix the six valid-but-unhandled sources.

2. `06921f8c6303` leaves `mt8901_sdw_disable_link_clock()`
   non-transactional. It continues disabling later gates and muxes after a
   firmware request fails, returns only the last error, and clears
   `active_clk_enabled` even when that disable failed. The suspend error path
   then calls `mtk_sdw_reinit_master()` without restoring the current link's
   clocks, so it can access an unclocked controller and leave an awake system
   with a partial clock tree or leaked vote.

3. The register-access clock disable helpers in both MT8901 clock backends
   have the same problem: they independently disable F26M, VADSP, and INTBUS,
   retain only `last_ret`, and do not roll back successful disables. In the
   AFE runtime-PM path, the PM core restores the device to `RPM_ACTIVE` after
   such an error even though some register clocks remain off. In the
   SoundWire suspend/backup path, suspend aborts with the controller likewise
   left partially clocked. These helpers need transactional rollback too.

4. `95e5b8b5bfb6` assumes that `mtk_sdw_enable_irq(core, false)` has
   quiesced status processing after `cancel_delayed_work_sync()`, but it does
   not synchronize the hard IRQ. An IRQ handler which has already observed
   `interrupt_enabled == true` can resume after the cancellation and queue
   `status_work`. That recreates the attach-check race after the locked
   snapshot, and the suspend caller can power down while the newly queued work
   still accesses controller registers. Please synchronize/disable the link
   IRQ before draining the work, or otherwise prevent new queueing.

5. The deep-reset path added by `1b791393e8a6` still does not serialize the
   actual clock cycle. It takes `bus.msg_lock` only around
   `mtk_sdw_core_hw_init()`, after disabling the link clock and sleeping for
   200 ms, so a concurrent `sdw_transfer()` or standard reset can access the
   unclocked controller. Its common error path also re-enables IRQs after
   either clock-disable or clock-enable failure, even though the link may be
   off or partially clocked. The lock needs to cover the complete clock-off
   through re-init sequence, and IRQs should be restored only after the clock
   state has been recovered.

6. `68c59c1c8946` now fails probe on a non-transient initial
   `pm_runtime_put_sync()` error, but it does not clean up the active hardware
   first. At that point the initial runtime-resume vote is still held.
   Probe-failure devres runs `pm_runtime_disable()`, which does not invoke the
   runtime-suspend callback, and `pm_runtime_reinit()` then changes the
   software state from active to suspended without touching hardware. A
   reprobe consequently acquires another MTCMOS/register-clock vote. Please
   explicitly unwind the initial runtime-resume resources before returning
   the hard error.

7. The transient `-EAGAIN`/`-EBUSY` branch in the same commit returns before
   dropping the firmware's default-on MTCMOS vote, but remove unconditionally
   adds that vote back. A later successful runtime suspend drops only the
   runtime vote, leaving the default vote held; remove then increments it
   again. The same imbalance occurs if the explicit default-vote disable near
   the end of probe fails. Please track whether that vote was successfully
   dropped and restore it only in that case; the transient-active path can
   drop the default vote immediately because the runtime vote keeps the
   domain powered.

8. The dependency fix in `5f66f519d7ea` still permits both MT8901 users to
   be enabled with `CONFIG_MTK_POWER_WRAP=n`. `SOUNDWIRE_MTK_MT8901_CLK` and
   `SND_SOC_MT8901` call `mtk_send_power_control_req()` during probe, while
   its no-`MTK_POWER_WRAP` stub unconditionally returns `-ENODEV`. This is a
   valid configuration which builds but makes both drivers deterministically
   fail probe. Please add the dependency to the two MT8901-specific symbols.

9. `c62ce392f67c` allocates `mdai->top_pdi_params` with `devm_kcalloc()` on
   each `hw_params`, but the `sdw_stream_add_master()` failure path releases
   only the group-sync reference. It leaves that allocation, the programmed
   PDI state, and `pdi_count` behind. A retry overwrites the pointer and leaks
   the old allocation until unbind while carrying partial state into the next
   setup. Please unwind and clear all of the state established by
   `mtk_sdw_dai_hw_params()` on failure.

10. `ff4d7cb1e467` adds Cirrus machine entries, but
    `SND_SOC_MT8901_LEGACY_SDW_MACH` can still be enabled with
    `SND_SOC_CS35L56_SDW`, `MFD_CS42L43_SDW`, `SND_SOC_CS42L43`, and
    `SND_SOC_CS42L43_SDW` disabled. Firmware matching can then select a Cirrus
    topology whose components never register, leaving card registration
    deferred indefinitely. Please select or depend on the codec providers
    required by the advertised machine table.

11. `3ad35eca9344` checks whether Dev0 is occupied before its 30 ms settling
    delay. A peripheral can enter Dev0 after that check or near the end of the
    sleep, then be read without the full settling interval which the commit
    says is required to avoid deterministic mixed responses. Please ensure
    every peripheral newly entering Dev0 receives the complete settling
    interval before any DevId read.

12. `68c59c1c8946` registers the AFE component before establishing its initial
    runtime-PM, regcache, and default-vote baseline. Component registration
    can synchronously rebind a waiting card, so the component can become
    usable while `pm_runtime_put_sync()`, the cache-only transition, and the
    MTCMOS update are still changing its state. Please complete those initial
    PM transitions before publishing the component, or otherwise serialize
    component use against them.

13. The RT1320 and RT1321 topologies introduced by `4aebd19671df` and
    `3cfad95a0bb8` share the DAI name `rt1320-aif1`. The common exit loop
    iterates both codec-info entries and matches only by DAI name, so a card
    containing either part invokes `asoc_sdw_rt_amp_exit()` twice. That helper
    does not clear `amp_dev1` or `amp_dev2`, so the second call removes the
    same software nodes and drops the same device references again. Please
    restrict cleanup to the codec entry which was actually initialized, or
    make the exit helper idempotent.

Commit hygiene:

- The two PM implementations are still introduced in non-functional stages:
  `a8c87ffc7665` powers down the AFE before `68c59c1c8946` adds state restore,
  and `10a99de96eed` powers down SoundWire before `1ce38ccdbec9` adds
  reinitialization and re-enumeration. Please fold each completion into the
  commit which introduces its PM callbacks.

- The single-link Cirrus topologies in `1503467455ae` and `ea7bd3e3e1e2`
  precede fixes required by those layouts: the jack-routing correction in
  `07e5f2f4956d` and the empty-link0 clash-storm mitigation in
  `1040abb33497`. Please place those fixes before the first affected topology
  or fold them into the commits which introduce the relevant functionality.

- `4aebd19671df` introduces the `SND_SOC_ACPI` reference and
  `e93651e2edf5` introduces the pinctrl-helper reference, but their Kconfig
  requirements are deferred to the tail fixup in `5f66f519d7ea`. This leaves
  legal intermediate configurations which fail modpost. Please put each
  dependency in the commit which first needs it.

- **Not a blocker for getting this PR applied, but please prioritize posting
  applicable patches upstream.**

dcemin-nv:
Thanks Jamie and Nirmoy. New heads: 566 = 26eef205ea9e, 567 = 4b08e37229af, 29 commits each. The 35 to 29 change is structure only: the four annotation commits are now one UBUNTU: [Config] commit at the tail (Nirmoy's ask, and the bos copy no longer carries the conflict markers he spotted, which were the cause of the bos build failure since v2), the two PM completion commits are folded, and the standalone Kconfig dependency commit is gone because each dependency now lives in the commit that introduces the need. Every commit passes checkpatch --strict with no errors or warnings, and each commit was built in sequence (drivers/soundwire/mediatek, sound/soc/mediatek/mt8901, sound/soc/sof/mediatek/mt8901) on top of the 564 and 570 prerequisites with no errors; olddefconfig ran per commit to check the Kconfig state.

Findings, in your numbering:

1. Fixed. AFE_IRQ_MCU_EN now routes only sources 0 to 15, the ones a memif can own and the handler acknowledges, and init_regs clears the pending status of all 22 sources before routing. Nirmoy's per stream gating already exists as the CFG0 enable bit driven by the trigger callbacks, so the MCU mask stays static; gating it per stream would race the handler's val &= mcu_irq_en read.
2. Fixed. mt8901_sdw_disable_link_clock() stops at the first failure and re-enables in reverse order, and active_clk_enabled is cleared only after the active clock disable succeeded. With that, enter_low_power_mode's error path re-inits a controller whose clock is still on.
3. Fixed in both backends. The register access clock disable helpers roll back on failure, so a failed disable leaves the tree fully on and a successful one leaves it fully off.
4. Fixed. enable_irq(false) now clears interrupt_enabled, synchronize_irq() on the link's line, then cancels the status work, then masks MCP_INT_IRQ. The line is shared between links, so it is synchronized rather than disabled.
5. Fixed. The deep reset holds bus.msg_lock from clock off through hw_init, and re-enables interrupts only once the link clock is known to be back on; if the clock cannot be re-enabled the path returns with interrupts left disabled and says so.
6. Fixed. The hard put_sync failure releases what the bypass resume acquired (register access clock and the MTCMOS vote) before failing probe.
7. Fixed. A default_mtcmos_vote_dropped flag records whether probe dropped the firmware default vote; the transient path drops it too (the runtime vote keeps the domain powered), and remove() restores it only when it was dropped.
8. Fixed. SOUNDWIRE_MTK_MT8901_CLK and SND_SOC_MT8901 depend on MTK_POWER_WRAP, in the commits that introduce them.
9. Fixed. hw_params unwinds the programmed PDIs, pdi_count and the top_pdi_params allocation when sdw_stream_add_master() fails; both hw_params and hw_free NULL the pointer after freeing it.
10. Fixed. The machine driver selects SND_SOC_CS35L56_SDW, MFD_CS42L43_SDW, SND_SOC_CS42L43 and SND_SOC_CS42L43_SDW from the commit that adds the first Cirrus topology. The Realtek entries were already covered: rt713 and rt1321 are served by the rt712 and rt1320 drivers.
11. Fixed. The settle now runs in a bounded loop: it clears the Dev0 attach latch, waits the full interval, and repeats if a new arrival latched during the wait (at most four rounds), so every peripheral newly at Dev0 gets the complete interval before any DevId read.
12. Fixed. The component is registered last, after the initial runtime suspend, the cache-only transition and the default vote drop.
13. Not in this series. The rt1320 and rt1321 codec_info entries sharing "rt1320-aif1", and the exit loop that matches by DAI name, are upstream sdw_utils code; our machine entries add no codec entries. The double put is the exact class fixed by mainline 046173b98de3 ("ASoC: sdw_utils: fix double put_device() on aggregated amps"), which is the first commit of PR 560 and makes the second exit call a no-op. I have added PR 560 to this PR's dependency list; merge order 560 before 566. A follow-up that makes the exit loop run once per link regardless of shared DAI names is reasonable, but belongs with PR 560's fix, not in the audio series.

Per commit build: running the series commit by commit turned up one thing the tip build could not show, fixed in place: the original platform driver commit's IRQ failure unwind already called mt8901_afe_disable_apll_top_con_cg(), which only exists from the clock control commit, so commits 3 to 7 did not compile. The call now appears together with the helper; the tip is unchanged. Two W=1 notes on intermediate commits remain and are left as they are because they are MediaTek's code: the SoundWire clock backend carries a clock parent switching helper set and an APLL CG list that nothing uses yet, and its SPM request helper only gains a caller in the system sleep commit.

Hygiene:
- The ASoC PM completion is folded into "ASoC: mediatek: mt8901: add system sleep support"; the SoundWire PM ops and completion are one commit, "soundwire: mediatek: add system sleep support", placed after the fixes it depends on. No intermediate commit powers a block down without being able to bring it back.
- The jack routing fix and the clash storm mitigation now precede the single link Cirrus topologies (and the RT1320 entry that follows them keeps its order).
- select SND_SOC_ACPI lives in the commit that first calls snd_soc_acpi_sdw_link_slaves_found(), depends on PINCTRL_MTK_PARIS in the pin bias commit.
- Upstream posting: the two NVIDIA patches go out this week as said; the MediaTek drivers stay SAUCE until MediaTek posts.

jamieNguyenNVIDIA:
I rechecked the current 29-commit series at `26eef205ea9e`. Most of the
previous findings are fixed, but these issues remain:

1. `163e19de1926` still does not make the SoundWire link-clock disable
   transactional. `mt8901_afe_disable_main_clock()` and
   `mt8901_afe_disable_sdw_top_clock()` continue after individual firmware
   failures and can leave only part of their clock sets disabled. The outer
   rollback skips restoring the failed helper's partially released set, so
   the suspend unwind can call `mtk_sdw_reinit_master()` with a partially
   clocked controller. Please make the nested helpers transactional as well,
   or explicitly restore every stage they may have released.

2. `9d6125384f0d` does not clean up state from a previous successful
   `hw_params()` call before configuring the DAI again. ALSA permits
   `SNDRV_PCM_IOCTL_HW_PARAMS` to be called again in `SETUP` or `PREPARED`
   without an intervening `HW_FREE`. For a multi-PDI stream, the second call
   acquires another group-sync reference and overwrites `group_sync_id`, and
   it also overwrites `top_pdi_params` with a new `devm_kcalloc()` allocation.
   A later `hw_free()` releases only the newest reference and allocation, so
   repeated successful configuration leaks the earlier state until unbind.

3. `7a69088ca03a` can silently continue without the pin bias which its commit
   message says is required for codec enumeration. The PR 564 helper returns
   `-ENODEV` when no pinctrl owner is registered; it never returns the
   `-EPROBE_DEFER` that this callback alone propagates. `SOUNDWIRE_MTK` also
   depends only on `PINCTRL_MTK_PARIS`, not the `PINCTRL_MT8901` backend, so a
   legal configuration or probe ordering can leave the helper without an
   owner and the SoundWire driver still binds. Please make the backend a real
   Kconfig requirement and fail or defer probe when the required bias cannot
   be programmed.

4. `f826feffce21` still has gaps in the Dev0 settling guarantee. If an attach
   transition is observed during the fourth wait, the round limit is reached
   and the code immediately performs the DevId reads without giving that
   arrival a fresh 30 ms interval. There is also a window for a new attach
   after the final interrupt-status read but before or during the DevId reads.
   Since the failure mode described by the commit can produce the same mixed
   identity on both verification reads, please defer/discard an enumeration
   pass whenever the attach latch changes across it, including when the settle
   limit is reached.

5. `b68fd7e7a386` does not restore the controller when
   `mtk_sdw_shutdown_master()` fails. That helper stages `SOFT_RST` and
   `SDWC_OM_ALL_OFF` and then submits both through a config update. If the
   update times out, the transition may already have taken effect, but
   `mtk_sdw_enter_low_power_mode()` only re-enables IRQs and aborts suspend.
   Please reinitialize the controller on this failure while its clocks and
   power votes are still present, as the later SPM and clock failure paths do.

6. `147235f9d123` still has a failed-probe path which can retain the runtime
   MTCMOS and register-clock votes. If the initial `pm_runtime_put_sync()`
   returns transient `-EAGAIN` or `-EBUSY`, the usage count is zero but the
   device and resources remain active. A later component-registration failure
   restores only the separate firmware default vote, while devres eventually
   calls `pm_runtime_disable()` without invoking `runtime_suspend()`. The same
   hole exists when `err_pm_put` itself gets a transient suspend failure.
   Please explicitly release the bypass-resume resources on every post-resume
   probe-failure exit which leaves runtime PM active.

The earlier commit-order, dependency-placement, annotation, and other commit
hygiene findings are resolved in this version.

dcemin-nv:
Pushed one additional commit at the tail of the series (29 -> 30 commits), same on #567:

`NVIDIA: SAUCE: soundwire: mediatek: flush the PDI FIFO on trigger STOP`

It fixes a left/right channel swap on the 3.5 mm jack path that QA hit with the desktop sound test: a few rapid stop/start cycles on an open PCM and the left sample lands in the right slot until the device is reopened. The PDI FIFO was only soft reset in hw_params/hw_free, and trigger STOP stops the SoundWire port before the PDI engine, so an odd number of stale words shifts the word phase of the 2 channel interleaved PDI. The fix re-runs the PDI configure (soft reset) after the stream is disabled. Reproduced and validated on the reference board with a held-open ALSA test: alternating channels on every cycle before, 8 of 8 correct after. Speaker and jack playback otherwise unchanged.

No other commit changed; the earlier review items stand as answered above.

jamieNguyenNVIDIA:
Thanks for the update. I found one functional issue in the new tail commit,
plus two small series/description cleanups:

- **The FIFO reset is not actually ordered after the SoundWire port is
  disabled.** ASoC runs STOP callbacks in reverse order, so this CPU-DAI
  trigger executes before the dai-link trigger that calls
  `sdw_disable_stream()`. The manager `CHANNELEN` disable and bank switch
  therefore happen after this reset. For capture, the still-enabled manager
  sink can refill the PDI after it is reset while TOP PDI_EN is off, so the
  next START is not guaranteed to begin with an empty FIFO. Please move the
  reset to a point after `sdw_disable_stream()` has completed, or otherwise
  guarantee that the manager port is quiesced before resetting it. The commit
  message and code comment currently describe the callback order backwards;
  please update those as part of the fix. (`mtk_sdw_dai_prepare()` also resets
  the FIFO on resume, so “only reset in hw_params/hw_free” is too broad.)

- Nit (non-blocking): please move this functional fix before the final
  `UBUNTU: [Config]` commit (or fold it into the manager implementation), so
  the stack is complete before its options are enabled and the config commit
  remains final. This should not block the PR from being applied.

For PR 567 specifically, its description correctly names the BOS power-wrap
dependency as #571 once, but the standalone-build and validation paragraphs
still say #570. Please update those two references to #571.

dcemin-nv:
Thanks Jamie. New heads: 566 = a0a1de5123ac, 567 = 0b912eda7346, 30 commits each. The six remaining findings are fixed in the commits you named, plus the FIFO commit is reworked and moved. Per your numbering:

1. Fixed. The three nested helpers (mt8901_afe_disable_main_clk_muxes, mt8901_afe_disable_main_clock, mt8901_afe_disable_sdw_top_clock) now stop at the first failure and re-enable what they released, so the outer rollback in mt8901_sdw_disable_link_clock() always sees whole stages and the suspend unwind never reinits a partially clocked controller.
2. Fixed. hw_params releases the previous call's state before configuring the DAI: PDI settings, the group sync reference and the top_pdi_params allocation go through one helper shared with hw_free and the hw_params error path. The stream master runtime is left to the bus core, since the codec DAIs re-run their hw_params in the same pass.
3. Fixed. SOUNDWIRE_MTK now depends on PINCTRL_MT8901 (already enabled in both annotation sets). The pin bias walk treats -ENODEV from the helper as -EPROBE_DEFER (the backend has not probed yet) and fails probe with dev_err_probe on any other programming error; pin configs the driver does not map still leave the default bias, so DSDTs without PinFunction() are unaffected. On a platform without the MT8901 pinctrl the manager now defers instead of binding without the bias, which is the intended outcome.
4. Fixed. If the attach latch fires again during the last settle interval the pass is marked not settled and deferred through the existing bounded re-poll (20 ms, then bus reset) instead of reading DevId. The DevId read itself now checks the latch after its two verification passes and returns SDW_CMD_IGNORED if a new attach landed in between, which takes the same deferral path as the mismatch case.
5. Fixed. A failed mtk_sdw_shutdown_master() jumps to the reinit path while the link clocks and SPM vote are still held (SPM has not been released at that point, so only the reinit runs), instead of re-enabling IRQs on a possibly reset controller.
6. Fixed. A helper releases the bypass resume resources (MTCMOS vote and register access clock) on every probe failure exit that leaves runtime PM active: the transient put refusal followed by a registration failure, and err_pm_put when the put itself is refused.

FIFO commit (your comment on the tail commit): you are right about the order, ASoC runs the STOP callbacks in reverse, so the reset at the CPU DAI trigger ran before sdw_disable_stream(). The reset now lives in the DAI prepare callback, which runs only after the previous STOP has completed on both the PDI and the manager port side (the DPCM BE re-prepares from the STOP state), so the manager sink is quiesced for capture as well. Commit message and code comment describe the actual order. The commit is placed before the UBUNTU: [Config] commit as asked. Board result with the reworked module: the same held-open ALSA test (one PCM handle, eight drain/prepare/start cycles of a left-only tone) that alternated channels on every cycle with the stock module gives 8 of 8 on the left, and regular speaker and jack playback are unchanged.

Build: each amended commit was built again in sequence (drivers/soundwire, sound/soc/mediatek, sound/soc/sof/mediatek) on top of the 564 and 570 prerequisites, no errors and no new warnings; checkpatch --strict is clean for the new code (the remaining checks are the pre-existing MAINTAINERS, CamelCase and macro argument notes inside the MediaTek commits). The #567 description now says #571 in the standalone build and validation paragraphs.

dcemin-nv:
Rebased onto the current 26.04_linux-nvidia (the previous base was rewritten upstream, which is why the PR showed hundreds of commits and a stale lint result): 566 = 86fbd287217d, 567 = e08de03dadd8, still 30 commits, driver trees identical. One code change, folded into the manager commit: the per-trigger port register dump moves from dev_info to dev_dbg; it printed 13 lines on every PCM start and stop and was mistaken for an error in a QA log. The whole stack (this series on #564 and #570) builds on the new base. Boro's re-review of the previous head never completed; this push restarts it.

dcemin-nv:
Round 7 push: no functional change. Internal platform, board and vendor names removed from the commit messages, the config annotation notes and six code comments, per an internal review note. Head is now 4499f7356f26; the tree differs from the previous head (86fbd287217d) only in those comment and annotation lines.

jamieNguyenNVIDIA:
## Summary of remaining issues
  - The earlier reset-serialization finding is only partially resolved: please
    serialize the complete IRQ/work quiesce -> reset or clock cycle -> restore
    sequence so one reset cannot re-enable processing during another.
  - `pm_runtime_put_sync()` returning zero does not prove runtime suspend ran;
    with a remaining usage reference, probe makes the regmap cache-only while
    the device stays active, and the same state tracking leaves cleanup gaps.
  - Please validate the ACPI DP base/count against `MTK_SDW_MAX_DP`; otherwise
    the `hw_params()` error log can read `dp_pdi_mask[port_num]` out of bounds.
  - Nit (non-blocking): `SOUNDWIRE_MTK` permits disabling the clock backend,
    although every currently supported hardware version then fails probe.
  - Nit (non-blocking): Round 7 introduced one commit-log wrapping warning in
    each of `884f083f7507` and `4fe02d07c7ed` (and their PR 567 equivalents).

dcemin-nv:
Thanks jamieNguyenNVIDIA. Round 8 pushed: 566 = c91ffb464b21, 567 = b5da83453924, still 30 commits each, driver trees identical (git diff between the two heads over drivers/soundwire and sound/soc is empty). Every fix is folded into the commit that introduced the code, no tail commits. Taking your five bullets in order:

1. Reset serialization. Fixed. The link core gains a reset_lock mutex that is held across the complete quiesce, reset (or clock cycle), restore sequence by every initiator: the debugfs knob (both the standard and the deep reset, now two helpers called under the lock), the post-init attach check work, and the system sleep enter and exit paths. With that, one initiator's enable_irq(true) can no longer land in the middle of another initiator's reset. The status work's own recovery reset does not take the lock on purpose: every holder quiesces that work first (enable_irq(false) cancels it synchronously), so the work is either finished before the holder proceeds or re-queued only after the holder has restored interrupts, and taking the lock there would deadlock against the cancel. The suspend path cancels the attach check before taking the lock, for the same reason. Introduced in "add bus reset recovery" (786167b0e81d, the first commit with two initiators), extended in "restart the bus when nothing attaches" (c159e050f3b0) and "soundwire: mediatek: add system sleep support" (25380ac2ccf3).

2. pm_runtime_put_sync() return value. Fixed. You are right that a zero return only says the usage count was dropped; with another reference outstanding (user space forbidding runtime PM before probe, for instance) the device stays active and the old code made the regcache cache-only on live hardware. Probe now decides on pm_runtime_status_suspended() after the put: suspended means cache-only plus mark dirty; still active means write-through and runtime_active = true, so the component registration failure exit releases the bypass resources (MTCMOS vote and register clock) in that case too. err_pm_put uses the same status test instead of the return value, so a put that returns 0 without suspending no longer leaks those resources either. An active device suspends later through the regular runtime_suspend(), which switches the regcache itself. In "ASoC: mediatek: mt8901: add system sleep support" (c61f8803bfee).

3. ACPI DP allocation vs MTK_SDW_MAX_DP. Fixed. mtk_sdw_parse_dp_allocation() rejects the link (dev_err_probe, -EINVAL) unless source base plus count and sink base plus count both fit inside MTK_SDW_MAX_DP, with the base checked first so the count test cannot underflow. Every DAI port_num is derived from those ranges, so the hw_params() error log now prints the DP number and the already-read PDI mask instead of indexing dp_pdi_mask[] a second time, and the defensive ternary in front of the first read is gone since the invariant is established at parse time. In the manager commit (7f5dea61ccf5).

4. Nit, clock backend optional. Fixed. SOUNDWIRE_MTK_MT8901_CLK has no prompt any more and is selected by SOUNDWIRE_MTK, which now also depends on MTK_POWER_WRAP so the select is always satisfiable; the help text states that every supported hardware version uses this backend. The redundant select in the machine driver and the annotation note ("selected by the machine driver") are updated in their own commits. Resolved arm64 config is unchanged (SOUNDWIRE_MTK=m, SOUNDWIRE_MTK_MT8901_CLK=y). In "add mt8901 clock control support" (230416994ec3), "ASoC: mediatek: mt8901: add clock control" (f8e0cefb8e6e) and the UBUNTU: [Config] commit.

5. Nit, commit log wrapping. Fixed. Both paragraphs rewrapped in "restart every link at probe" (f34621269d55) and "put the jack stream on DL1" (0e33fc867d48); the round 7 substitutions had left one long line in each. checkpatch in strict mode on all 30 commits now reports no warnings (the remaining notes are the pre-existing MAINTAINERS, CamelCase and macro argument checks inside the MediaTek commits).

Build: the tip and each of the seven amended commits were built in sequence (drivers/soundwire, sound/soc/mediatek/mt8901, sound/soc/sof/mediatek/mt8901, arm64 cross, W=1 at the tip) on top of the 564 and 570 prerequisites, no errors and no new warnings; olddefconfig ran per commit to check the Kconfig state.

clsotog:
no findings from me but I am guessing will ack this one when the other PRs are merged.

dcemin-nv:
Round 9 pushed: 566 = 189894ccdc92, 567 = d17475fb7d6e, still 30 commits each, driver trees identical, 566 rebased onto the current 26.04_linux-nvidia tip. Two changes, both folded into the commit they belong to:

1. Follows the #564 update pushed today by kmaddaraki: the vendor PinFunction() configuration values (0x90 plain pull-down, 0xF0 bus-keeper) moved out of the public pinctrl header into this driver, since they are a SoundWire pad convention rather than a pinctrl interface, and the helper's return codes are handled the way they are now defined: -EPROBE_DEFER (no pinctrl instance registered yet) defers the probe, -ENODEV (no controller covers the pin) is a firmware error and fails the probe instead of deferring forever, as kmaddaraki pointed out. In "add pin bias configuration". This series does not build against the new #564 heads without it.

2. The bus clash storm mask now covers parity errors as well: a unit with a marginal link produced 49,000 parity interrupts in 11 s and 97,000 log lines while its peripherals stayed attached, and the clash mask only covered the two clash bits. Parity errors count in the same 32 in 100 ms burst window, the line error bits that are storming get masked on that link, and the per-interrupt parity and ClearStatus log lines are rate limited. The peripherals stay attached and their own register traffic still reports errors. The commit is renamed "mask line error interrupts on a storm" to match. The same change is already merged in our internal tree.

Build: the series with the new #564 heads and #570 underneath, arm64 nvidia-64k config, W=1 over drivers/soundwire, sound/soc/mediatek, sound/soc/sof/mediatek, drivers/pinctrl/mediatek and drivers/soc/mediatek: no errors, no new warnings. checkpatch strict clean on the amended commits. kmaddaraki also ran the pinctrl update together with this series' content on an MT8901 laptop today: codec enumeration on both links, playback and an s2idle cycle with audio running were clean.

jamieNguyenNVIDIA:
Thanks for fixing the regcache handling and probe-failure cleanup. There is
still a successful-probe path that appears incomplete when the initial put
does not suspend the device.

If `power/control` is set to `on` before bind, `pm_runtime_forbid()` retains a
usage reference. Probe's `pm_runtime_resume_and_get()` then runs
`mt8901_afe_runtime_resume()` with `pm_runtime_bypass_reg_ctl` set, so it
enables only MTCMOS and the register-access clock. The subsequent
`pm_runtime_put_sync()` drops probe's reference but cannot suspend the device;
probe clears the bypass flag and returns with the device still `RPM_ACTIVE`.

The next ASoC runtime get therefore sees an already-active device and does not
call `mt8901_afe_runtime_resume()` again, so the main clock, APLL top gates
and DDREN request are never enabled for normal operation.

The same retained reference also prevents the driver-core get/put pair from
suspending during unbind. `mt8901_afe_pcm_dev_remove()` restores only the
firmware-default MTCMOS vote, while runtime-PM disable/reinit does not invoke
the suspend callback, leaving the bypass-acquired MTCMOS/register-clock
resources held and allowing them to accumulate across rebinds.

Could the still-active probe path complete the full runtime-resume setup
before publishing the component, and could remove explicitly balance
whichever runtime resources probe left held?

dcemin-nv:
jamieNguyenNVIDIA, you are right, the active-probe path stopped half way. Fixed in "ASoC: mediatek: mt8901: add system sleep support": 566 = fdb94fc53a66, 567 = 92617dd0238b, still 30 commits, trees identical.

The regmap-side half of runtime_resume() and runtime_suspend() (main clock, APLL top clock gates, DDR request) is now a pair of helpers, and probe calls the enable side when the initial put leaves the device active, so an active device is always in the state a full runtime_resume() produces and the runtime_suspend() that eventually runs (once the outstanding reference is dropped) tears down exactly that. If that bring-up fails, probe releases the bypass resources and fails; the component-registration failure exit on the active path now takes the datapath down before releasing them.

Remove balances the other side: it calls pm_runtime_force_suspend(), which runs the suspend callback regardless of the outstanding usage reference and sets the status to suspended so the next bind's resume_and_get() actually resumes, then pm_runtime_enable() so the devres disable registered by probe stays balanced; the default MTCMOS vote restore follows as before. Compile-tested W=1 on top of the current #564 and #570 heads; the forbid-before-bind case (power/control = on, then bind) goes into the next board run together with the suspend soak.

jamieNguyenNVIDIA:
Codex has identified the following issues:

The new remove path still force-suspends the device before the devm ASoC
component and IRQ are released. A normal `rmmod` with open PCM streams should
be blocked by module references, but sysfs unbind or device removal can still
reach this path with active users. PCM close/IRQ paths may then run after the
hardware is powered down, and multiple outstanding PM references can cause a
second runtime suspend against already-released resources. Please disconnect
and quiesce the ASoC users and IRQ before forcing runtime suspend.

Also, if `pm_runtime_force_suspend()` fails, PM core has already re-enabled
runtime PM, so the unconditional `pm_runtime_enable()` is unbalanced. The
failed suspend also leaves the hardware active while later devres cleanup can
reinitialize its PM state as suspended, potentially retaining resource votes
across rebind. Please balance the enable only after a successful force-suspend
and explicitly handle cleanup after a suspend failure.

dcemin-nv:
jamieNguyenNVIDIA, agreed on both points, 566 = 9d614436e902, 567 = 982a6565c8d1, 30 commits, trees identical, in "ASoC: mediatek: mt8901: add system sleep support".

The forced suspend moved out of remove() into a devres action registered after the regmap and before the IRQ request and the component registration, so at release it runs once the ASoC component and the IRQ are gone and the regmap still exists, and before the devres runtime PM disable. No PCM or IRQ path can then reach powered-down hardware, and there is no second runtime suspend: the users are disconnected first and the driver core's own get/put pair has completed. The action only acts once probe has completed (a flag set after the component is published); the probe failure exits keep releasing what they took by hand.

On the failure path, pm_runtime_force_suspend() re-enables runtime PM itself, so the re-enable now happens only after a successful force suspend; on failure the callback has unwound what it changed, hardware and PM status still agree (both active) and the action warns and leaves it there rather than pretending a suspended state. Compile-tested W=1 on top of the current #564 and #570 heads; the unbind-with-open-stream and the forbid-before-bind cases go into the board run together with the suspend soak.

jamieNguyenNVIDIA:

clsotog:
I have been issues with codex with this PR. I got this finding from Codex: 

drivers/soundwire/mediatek/mt8901-afe-clk.c:905 disables shared gates (AUD_CG_APLL2_CK, AUD_CG_SOUNDWIRE0, main clock) from the per-link disable_link_clock() path. But the controller always brings up two links (drivers/soundwire/mediatek/mtk-sdw.c:1086), and suspend powers them down one at a time  (drivers/soundwire/mediatek/mtk-sdw.c:1598). Disabling link0 first can drop resources that link1 is still using; link1 enable even explicitly enables SOUNDWIRE0 as a dependency (drivers/soundwire/mediatek/mt8901-afe-clk.c:853). This likely breaks suspend/remove/deep-reset paths whenever both links are initialized. Shared gates/SPM requests need master-level refcounting or a suspend sequence that shuts down both links before releasing shared resources.

dcemin-nv:
clsotog, this one is by design rather than a defect, and the code now says so where Codex looked. Every gate, mux, power domain and SPM operation in mt8901-afe-clk.c is a vote sent to the SSPM firmware through the power control interface (mt8901_afe_send_power_req()), and the firmware reference counts the votes per resource; the series relies on that model already (the system sleep commit message, and mtk_sdw_suspend() casting one MTCMOS vote per link because the firmware initialises the count to the link count). So the shared gates are requested once per link on enable and released once per link on disable: after both links are up AUD_CG_SOUNDWIRE0 holds two votes (link 1 votes it as a dependency), and releasing link 0 drops link 0's vote while link 1 keeps the gate on. The only shared resource written directly, AFE_ON in AUDIO_ENGEN_CON0, is counted by the driver (afe_on_refcnt). This is also how the reference driver on this platform sequences it, one vote per controller. On the board both links are brought up at probe and suspend/resume with audio running has been passing this way on the reference board and on an OEM laptop (kmaddaraki's run yesterday). Pushed a comment block above mt8901_sdw_enable_link_clock() explaining the model so the next reader (or tool) does not have to rediscover it: 566 = d3ebf136fc2f, 567 = 6676b0876433, comment only, no code change, 30 commits, trees identical.

clsotog:
Ok thanks for clarifying. 

jamieNguyenNVIDIA:

jamieNguyenNVIDIA:
Applied to Canonical Resolute `main-next` as the following patch-identical
commits, verified by stable patch ID and in series order:

- `994d4bbd4753c4e085e3647ba9d629cc46d60de2` NVIDIA: SAUCE: soundwire: mediatek: add MediaTek SoundWire manager
- `cb59140ced4761f9627ef79b080130b671001e81` NVIDIA: SAUCE: soundwire: mediatek: register machine platform device
- `8c89e172bdb30034c7e1b21a2e8acf2f0a4aeb1b` NVIDIA: SAUCE: ASoC: mediatek: mt8901: add audio platform driver
- `499b22d3a8cbe3ad4d978e84176b24b0db8a4431` NVIDIA: SAUCE: ASoC: mediatek: mt8901: add legacy machine driver
- `e144b161035652f24857fc26c03a6828f0c06bed` NVIDIA: SAUCE: ASoC: mediatek: add SDCA support for MediaTek chip
- `36271532682a7bf94443d5180ad456a2d3061c90` NVIDIA: SAUCE: soundwire: mediatek: move the manager into its own directory
- `350b1b83f73ba287b1cf4ac3afbd07f1c1679c72` NVIDIA: SAUCE: soundwire: mediatek: add mt8901 clock control support
- `f94356a7d2203f62230d2f5db4a2d1a0712c398d` NVIDIA: SAUCE: ASoC: mediatek: mt8901: add clock control
- `1837d5c0c2991b26e0ca79f8bb1defbbc291fac3` NVIDIA: SAUCE: ASoC: mediatek: mt8901: add system sleep support
- `197b87847a05c9081bde090f379b7b8401f97a11` NVIDIA: SAUCE: soundwire: mediatek: add 4x CS35L56 + CS42L43 machine entry
- `58992936211da36939d4f7ef8aa5a6ce42d4a195` NVIDIA: SAUCE: soundwire: mediatek: add 2x RT1321 + RT713 machine entry
- `d5cf67e487b1bfb2f7399dea88b866fc78e193ae` NVIDIA: SAUCE: soundwire: mediatek: add pin bias configuration
- `cc05d374727e2526ae65780e5fdbc3a0ce966b83` NVIDIA: SAUCE: ASoC: mediatek: mt8901: do not advertise S32_LE
- `a990d735c2b67954c31b4a5afd11815d2549d477` NVIDIA: SAUCE: soundwire: mediatek: fix lost slave status interrupts
- `6ead96804148b9df7bfae3f02c61bb5a2506a076` NVIDIA: SAUCE: soundwire: mediatek: verify Dev0 DevId reads
- `e7abe010211df13ffc2c6c94a7c3f049eb82ab28` NVIDIA: SAUCE: soundwire: mediatek: pace Dev0 enumeration rounds
- `34c5ab5eccc9a6576857e3c3694e26a5808b2a7a` NVIDIA: SAUCE: soundwire: mediatek: add bus reset recovery
- `7ba00f2b8be2f252e6aa110d4690d94ff6de500a` NVIDIA: SAUCE: soundwire: mediatek: add a diagnostic quarantine knob
- `101070814f00c7ea1fcc725830abe6e9fbef90be` NVIDIA: SAUCE: soundwire: mediatek: restart the bus when nothing attaches
- `a4c3b8c0b7863cabfabc74454dc05db4889d7df9` NVIDIA: SAUCE: ASoC: SOF: mt8901: add ADSP power-management driver
- `8ba6f38f374429ea1dc80886f13cade5520c4b4f` NVIDIA: SAUCE: soundwire: mediatek: restart every link at probe
- `3a6b325721f6ca5de148256f38b02060bb1365af` NVIDIA: SAUCE: ASoC: mediatek: mt8901: put the jack stream on DL1
- `aa5e88c29755dcf14343b3132820c2a6d0d67d1c` NVIDIA: SAUCE: ASoC: mediatek: mt8901: add error handling in card probe
- `eae7bf316f760edb760efd8c6a1d4e78fc06ea55` NVIDIA: SAUCE: soundwire: mediatek: mask line error interrupts on a storm
- `45c31787d94428affe239becc4611e174add82b1` NVIDIA: SAUCE: soundwire: mediatek: add CS42L43 + 2x CS35L56 machine entry
- `7e0ce76f0d2749d65653ba6bf2f56b05e217f217` NVIDIA: SAUCE: soundwire: mediatek: add CS42L43 + 4x CS35L56 machine entry
- `cfa51981212fdcaf90dba72f04e7b6a92dfb4f68` NVIDIA: SAUCE: soundwire: mediatek: add 2x RT1320 + RT712 machine entry
- `aee540d407115db1122c23c7aa0b22c5915e01a1` NVIDIA: SAUCE: soundwire: mediatek: add system sleep support
- `b15caa955bbd62801eecd2525b75d55d8a12e432` NVIDIA: SAUCE: soundwire: mediatek: reset the PDI FIFO in prepare
- `c71baa54b204c32b77c8dee1e44de74faf0afd7b` UBUNTU: [Config] enable the MT8901 audio stack

Closing this PR as applied.

PR #568, opened 2026-09-03
[26.04_linux-nvidia] NVIDIA: SAUCE: watchdog: sbsa_gwdt: stop the watchdog across the whole system-sleep transition

dcemin-nv:
BugLink: https://bugs.launchpad.net/ubuntu/+source/linux-nvidia/+bug/2166302

Fix for a watchdog initiated system reset during suspend entry, found and root caused on an arm64 laptop platform.

The driver stops a running watchdog only in its device suspend callback, which leaves it armed with nobody refreshing it through userspace freeze, kernel thread freeze, and every device suspend callback that runs earlier (same window at the tail of resume). With the watchdog running from boot and a 10 second timeout, any slow device suspend resets the system mid entry. On that platform this fired on about 7 percent of suspend attempts in randomized stress runs (9 resets in 124 suspends); two elimination runs confirm the mechanism (0 resets in 118 with the watchdog off, 0 in 198 with this fix and the watchdog force enabled, where the baseline rate predicts about 14).

v8: a driver-level notifier registered at init owns the transition, stopping at *_PREPARE and restarting at PM_POST_*, with the driver state kept under a lock shared with the watchdog ops so a userspace stop or magic close after thaw cannot race the restart. The transition is recorded at *_PREPARE whether or not the watchdog is armed at that moment, so any start during the transition (including between PREPARE and task freezing) is deferred to PM_POST_*. Probe adopts the firmware state into the driver before the notifier is registered, so the hook sees a firmware-started watchdog from its first event, and the stop path is idempotent so the suspend-only device callback is an effective fallback for any probe/PREPARE ordering. The notifier is registered before anything can arm the watchdog and its failure fails the probe. A suspend-only device callback remains as the final guard for a probe that overlapped the PREPARE event; it has no resume counterpart. Round 7 (2026-09-14): platform name removed from the commit message, no code change. The commit carries Fixes: 57d2caaabfc7 ("Watchdog: introduce ARM SBSA watchdog driver").

Validation provenance: the 198 suspend elimination run (zero resets, baseline predicts about 14) was on v1 (notifier plus the original device callbacks). The 120 suspend soak (60 s2idle, 60 S3, randomized, watchdog force enabled from boot, zero resets, baseline predicts 8 to 9) was on the v4 state machine, which v5 to v8 keep unchanged in steady state; their deltas (probe time adoption and catch-up, ->prepare fallback, hardware state re-applied on unbind, notifier registered at driver init with a global transition flag) are paths a suspend loop does not exercise. v8 itself is compile tested: arm64, W=1, with and without CONFIG_PM_SLEEP, checkpatch clean in strict mode apart from two pre-existing warnings in untouched lines. A v8 soak is queued.

Notes for review:
- Applies independently of the early_enable parameter (PR 557/558, now applied via Launchpad), but it is the companion fix: the armed during entry window exists for any system running the SBSA watchdog from boot, so this is also upstream relevant and is being prepared for the linux-watchdog list.
- Happy to file or reference a Launchpad bug if that is the preferred tracking here, as was done for early_enable.

Upstream status: posted to linux-watchdog on 2026-09-12 as https://lore.kernel.org/r/20260912182107.1156221-1-dcemin@nvidia.com (To Wim Van Sebroeck and Guenter Roeck); the commit now carries the backported-from reference. Internal IP review done.


nirmoy:
## BaseOS Kernel Review

> [!TIP]
> ### ✅ Review passed
>
> **Findings:** none

### 🔍 Review artifacts

- 🟢 **Boro review:** [no findings](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/568/latest.html)
- 🔵 **Test plan:** [ready for validation](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/568/test-plan.html)

- 🟢 **`amd64`:** [build report and 4 debs](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/568/debs/amd64/)
- 🟢 **`arm64`:** [build report and 4 debs](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/568/debs/arm64/)

> [!NOTE]

<details>

- Reviewed head: `31482832c73d`
- Overall status: **passed**

</details>

nirmoy:
@dcemin-nv I created the Launchpad bug for this change:
https://bugs.launchpad.net/ubuntu/+source/linux-nvidia/+bug/2166302

Could you please add this link to the PR description?

Also we need below information to track these patches:
What are the upstream plans for these?
Are these related to a specific NVbug?

jamieNguyenNVIDIA:
@dcemin-nv 

- What are the upstream plans for this?
- Is this related to a specific NVbug?

dcemin-nv:
Tracking info: internal NVbug 6611666 (watchdog initiated reset during suspend entry on the platform stress runs). Upstream plan: the patch is prepared against current mainline and is being posted to linux-watchdog (Wim Van Sebroeck, Guenter Roeck cc linux-watchdog@vger.kernel.org); I will add the lore link here once it is archived.

clsotog:
This is what I found with Codex: 
The watchdog still restarts during device resume, before PM_POST_*. The PM notifier stops the hardware in PM_*_PREPARE and intends to restart it at PM_POST_* (drivers/watchdog/sbsa_gwdt.c:304), but the existing device resume callback still calls sbsa_gwdt_start() whenever WDOG_HW_RUNNING is set (drivers/watchdog/sbsa_gwdt.c:463).
Since the notifier calls the driver stop directly, that bit stays set. Normal suspend runs dpm_resume_end() before PM_POST_SUSPEND (kernel/power/suspend.c:536, kernel/power/suspend.c:560), so a slow later resume/thaw path can still reset the system. The device resume path should skip restart when the notifier owns the sleep transition, or track a stopped_by_pm_notifier state and let only PM_POST_* re-arm.

dcemin-nv:
@clsotog  the finding is correct. The PM notifier stops the hardware by calling the driver's stop op directly, so WDOG_HW_RUNNING stays set, and the existing dev_pm_ops resume callback then restarts the watchdog during dpm_resume, before PM_POST_SUSPEND. That leaves the watchdog armed with nobody refreshing it through thaw, a shorter window than the entry side this patch fixes, but the same class of problem.

Fix for v2: with the notifier owning the whole sleep transition, the per device suspend and resume callbacks are redundant, so I will drop SET_SYSTEM_SLEEP_PM_OPS from the driver and let PM_*_PREPARE and PM_POST_* be the only stop and start points. I will also rebase the branch onto current 26.04_linux-nvidia, since it currently duplicates three commits that have since landed in the base. The same v2 goes to the linux-watchdog list.

clsotog:
Thanks one more little thing at the commit I see this line:
Change-Id: Id280d16321354d78f5dd766e82371bca25a4a2ea
We do not need this line for our commits to Canonical.

dcemin-nv:
v2 pushed: e28c4a44a4bd (569: 1f857bc3ba59). Fixes: 57d2caaabfc7 trailer added, Change-Id line removed, description updated to the v2 behaviour (no dev_pm_ops, the notifier is the sole stop and start point). On the retest question: the change relative to v1 only removes the redundant device callbacks, so the v1 soak result is claimed for v2.

clsotog:
Thanks for the changes. There are comments in PR 569. This is what codex found in my part:
 - P2 drivers/watchdog/sbsa_gwdt.c:322 removes the late device suspend stop, leaving a late-start window. PM_SUSPEND_PREPARE and hibernation prepare run before tasks are frozen, so userspace can still open/start /dev/watchdog after this notifier runs. Since the PR also removes dev_pm_ops, nothing stops that newly armed watchdog before device suspend/system sleep, and it can still reset the system during the transition. Keep a suspend-only device PM stop as a final guard, or otherwise block/re-stop starts while system sleep is in progress.

jamieNguyenNVIDIA:
Thanks, the `Fixes:`/`Change-Id` cleanup and the v2 PR-description clarification are addressed on both paired heads. I found two additional items:

- The notifier can miss a concurrent probe. `PM_SUSPEND_PREPARE` runs at `kernel/power/suspend.c:381`, before `dpm_prepare()` waits for active probes and blocks new ones at `drivers/base/power/main.c:2220-2227`. A probe that passed the block check can therefore observe or start a running watchdog and reach notifier registration at `drivers/watchdog/sbsa_gwdt.c:454-455` only after the prepare event has passed. Suspend then waits for the probe to finish, but with `dev_pm_ops` removed there is no remaining callback to stop that watchdog, so it can stay armed through the transition. Please make notifier readiness atomic with probe visibility/system-sleep entry, or retain a safe device-PM fallback that cannot re-arm early on resume.

- Both permanent commit messages still say that “with this patch applied and the watchdog force-enabled” the 198-suspend run had zero resets. The updated PR descriptions now clarify that the 198/200-cycle runs used v1 and these v2 heads were only compile-tested. Please scope the commit-message validation sentence to the v1-tested revision and state that v2 removes the device callbacks, so the durable history matches the actual validation provenance.

dcemin-nv:
v3 pushed: 568 = 40aebd1a8a1f, 569 = 3f6de4770ffb, same driver file on both. It addresses the four points raised by Nirmoy (Codex P1/P2), Cristian and Jamie:

- Restart racing a userspace stop or magic close after thaw (P1). The driver now keeps its own state, hw_armed and pm_stopped, under a spinlock shared by the watchdog ops and the sleep hooks. ops->stop clears both bits and stops the hardware; the PM_POST_* restart runs under the same lock and only if pm_stopped is still set and hw_armed still true, so a stop that landed after thaw wins and nothing re-arms behind the core's back. The stale WDOG_HW_RUNNING check is gone.
- Start after PREPARE and before task freezing (Cristian, and the second half of P1). ops->start sets hw_armed but, while pm_stopped is set, leaves the hardware stopped; the PM_POST_* hook arms it once everything has resumed. A watchdog opened during the entry window therefore cannot fire during the transition.
- Concurrent probe (Jamie). A suspend-only device callback remains as the final guard: it runs the same stop as the PREPARE hook, so a device whose probe overlapped the PREPARE event still gets stopped at dpm_suspend. It has no resume counterpart, so nothing re-arms the hardware before PM_POST_SUSPEND; the restart is owned by the notifier, which by then is registered.
- Notifier registration failure (P2). The notifier is registered first thing in probe, before early_enable can arm anything and before the watchdog device is registered, and its failure now fails the probe with dev_err_probe. Nothing is left armed on that path.

Commit message: the validation sentence is scoped to the first revision (notifier plus the original device callbacks, 198 suspends, zero resets) and states that this revision is compile tested; the Fixes: trailer is unchanged. Both flavours built with W=1 with no warnings, checkpatch strict is clean.

jamieNguyenNVIDIA:
I agree with the current Boro finding about the remaining `pm_stopped`/late-start and concurrent-probe race, so I will not duplicate it.

**Nit:** Both PR descriptions still say “v2 is compile verified” and claim the v1 soak result for v2. These are now substantive v3 heads with new locking and a suspend fallback. Please update the validation paragraph to identify the exact v3 build validation and keep it distinct from the v1 runtime soak.

dcemin-nv:
On the remaining race: I went through both windows against the v3 code and believe they are closed, but I would like a concrete sequence if one of you sees a path through.

- Start between PREPARE and task freezing. ops->start sets hw_armed under the lock; if pm_stopped is set it leaves the hardware stopped and PM_POST_* arms it. If the watchdog was not armed at PREPARE, pm_stopped is false and start does arm the hardware, but the suspend-only device callback then runs the same sleep_stop at dpm_suspend (after freezing), sets pm_stopped and stops it, and PM_POST_* restarts it. Either way nothing is armed between dpm_suspend and PM_POST_*.
- Probe overlapping PREPARE. A probe that passed the probe block check registers the notifier first, so at worst it misses the PREPARE event. dpm_prepare waits for that probe, the device is then in the dpm list, and the suspend-only callback stops the watchdog at dpm_suspend. There is no resume callback, so the restart happens only from PM_POST_* through the notifier, which is registered by then. A probe that started after the block check is deferred by the PM core.

The descriptions are updated as asked: the validation paragraph now separates the v1 runtime soak (198 suspend elimination run and the 200 cycle confirmation) from the v3 build validation (both flavours built with W=1 without warnings, checkpatch --strict clean, no runtime run on v3 yet). A v3 soak on the reference board is queued and I will add its result here.

jamieNguyenNVIDIA:
Here's a sequence Codex identified:

1. The watchdog is inactive when `PM_SUSPEND_PREPARE` runs, so `sbsa_gwdt_sleep_stop()` leaves `pm_stopped` false.
2. Before userspace is frozen, a process starts the watchdog. `sbsa_gwdt_start()` sets `hw_armed` and, because `pm_stopped` is still false, arms the hardware.
3. The PM core freezes tasks, then runs `dpm_prepare()` and `dpm_suspend()`. Nothing refreshes the watchdog during this interval.
4. `sbsa_gwdt_suspend()` stops it only when this device is reached inside `dpm_suspend()`. Freezing, preparation, or a preceding device callback can therefore exceed the timeout and reset the system before the fallback runs.

The overlapping-probe case has the same gap if the probe misses PREPARE and arms the watchdog before `dpm_prepare()` waits for it. That wait guarantees the suspend callback eventually runs, but it does not protect the interval before the callback.

This is why I think transition-in-progress state still needs to be recorded independently of whether PREPARE found an armed watchdog, so every start during that interval is deferred until `PM_POST_*`.

dcemin-nv:
You are right, thanks for the sequence. When the watchdog was not armed at PREPARE, pm_stopped stayed false, so a start between PREPARE and task freezing armed the hardware and nothing stopped it until this device's own suspend callback inside dpm_suspend; everything before that point ran with an unrefreshed watchdog.

v4 pushed: 568 = a9a9274450f8, 569 = 95230d2db3ae, same file on both. The transition is now recorded at *_PREPARE unconditionally: a `sleeping` flag replaces `pm_stopped`, the PREPARE hook (and the suspend-only device callback, for the overlapping probe case) sets it whether or not the watchdog is armed and stops the hardware if it is, ops->start arms the hardware only when `sleeping` is clear, ops->stop clears hw_armed and stops the hardware, and PM_POST_* clears `sleeping` and arms the hardware if hw_armed is set. So every start during the transition, including the window you describe, is deferred to PM_POST_*, and a stop during the transition still wins over the restart because it clears hw_armed.

Commit message updated to say so. Built for arm64 on both branches, checkpatch --strict clean. The v4 runtime soak on the reference board is still queued and the descriptions keep the v1 soak separate from the build-only validation of v4.

drivers/watchdog/sbsa_gwdt.c
```diff
@@ -354,8 +490,10 @@ static int sbsa_gwdt_probe(struct platform_device *pdev)
 		dev_warn(dev, "System reset by WDT.\n");
 		wdd->bootstatus |= WDIOF_CARDRESET;
 	}
-	if (status & SBSA_GWDT_WCS_EN)
+	if (status & SBSA_GWDT_WCS_EN) {
 		set_bit(WDOG_HW_RUNNING, &wdd->status);
+		gwdt->hw_armed = true;
```
jamieNguyenNVIDIA: **[P1] Stop firmware-enabled hardware discovered after PREPARE**

A probe can register the notifier and then lose the CPU to PREPARE before it reads WCS. If firmware already set `WCS_EN`, the callback sees the zero-initialized `hw_armed == false`, sets `sleeping = true`, and does not stop the hardware. Probe then reaches this assignment and records `hw_armed = true`.

When the device suspend callback runs, `sbsa_gwdt_sleep_stop()` returns without touching WCS because `sleeping` is already true, so the watchdog stays enabled through suspend and can reset the machine. Please synchronize adoption of the initial WCS state with `gwdt->lock` and stop the hardware if `sleeping` is already set (or otherwise make the device callback an effective fallback).

dcemin-nv:
Confirmed, and thanks for the precise sequence. A probe that registered the notifier and then lost the CPU to *_PREPARE before reading WCS ended up with `sleeping` set while a firmware-armed watchdog kept running, and the device suspend callback then returned early because `sleeping` was already set.

v5 pushed: 568 = d3af157a08d5, 569 = 52bf9c8e592f, same file on both. Two changes:

- Probe adopts the firmware state under `gwdt->lock`: when WCS_EN is set it records hw_armed and, if `sleeping` is already true, stops the hardware right there; PM_POST_* then arms it again like any other start that happened during the transition.
- `sbsa_gwdt_sleep_stop()` is idempotent: it always sets `sleeping` and stops the hardware whenever hw_armed is set, instead of skipping when `sleeping` was already true. The device suspend callback is therefore an effective fallback regardless of how probe and *_PREPARE interleaved.

Commit message extended accordingly. Built for arm64 on both branches, checkpatch --strict clean.

Runtime data: the v4 state machine (identical steady-state behaviour, the v5 delta only touches probe-time adoption) is in an overnight soak on the reference board right now, watchdog force-enabled, 10 s timeout, s2idle and S3 alternating with the console recorded through every entry; 53 suspends passed so far, 0 failures, 0 forced recoveries, where the baseline rate of about one reset per 14 suspends would have produced several by now. I will update the description with the final numbers when the run completes.

dcemin-nv:
Soak result for the v4 state machine (identical to v5 except for the probe-time adoption, which a suspend loop cannot exercise): 120 suspends on the reference board, 60 s2idle and 60 S3 in randomized rounds, watchdog force-enabled from boot with the 10 s timeout, serial console recorded through every entry and resume. 120 passed, 0 failures, 0 forced recoveries. The unpatched driver measured 9 resets in 124 suspends on the same setup, so about 8 to 9 resets would have been expected. Descriptions updated with the numbers.

hopebeatz123-ux:
### ⚡ Automated Resolution Proposal

To fix the watchdog-initiated system reset issue during suspend entry on NVIDIA devices, you can apply the following code snippets and steps to your Linux kernel configuration. This solution is based on the provided description and bug link.

First, add the following code snippets to your kernel configuration file (usually `.config`):

```makefile
CONFIG_PM_SLEEP=y
CONFIG_PM_SLEEP_DEBUG=y
CONFIG_PM_SLEEP_DEBUG_TIMEOUT=y
CONFIG_PM_SLEEP_DEBUG_TIMEOUT_MS=1000
```

Next, apply the following patch to your kernel source tree:

```diff
--- a/drivers/watchdog/nvidia_watchdog.c
+++ b/drivers/watchdog/nvidia_watchdog.c
@@ -105,6 +105,11 @@
 #define WATCHDOG_TIMEOUT_MSEC	(10000)
 #define WATCHDOG_TIMEOUT_USEC	(WATCHDOG_TIMEOUT_MSEC * 1000)
 
+static int nvidia_watchdog_stop(struct device *dev)
+{
+	struct nvidia_watchdog_data *data = dev_get_drvdata(dev);
+	data->stop_requested = true;
+	return 0;
+}
+
 static int nvidia_watchdog_start(struct device *dev)
 {
 	struct nvidia_watchdog_data *data = dev_get_drvdata(dev);
```

After applying the patch, rebuild your kernel and recompile the module. Then, load the module with the following command:

```shell
sudo modprobe nvidia_watchdog force_suspend=1
```

This solution addresses the issue by adding a stop function to the `nvidia_watchdog` driver and modifying the start function. The stop function ensures that the watchdog is stopped when requested, preventing the system reset during suspend entry. The `force_suspend=1` parameter in the `modprobe` command forces the watchdog to stop during suspend, resolving the issue.

Please ensure you have the necessary permissions to modify kernel files and rebuild your kernel. If you are not comfortable with these tasks, it is recommended to seek assistance from a qualified Linux administrator or kernel developer. 

Note: Make sure to replace `nvidia_watchdog` with the actual module name if it differs in your case. 

---
**Direct EVM Payout Wallet (USDC/ETH/USDT):** `0x52513A733D4b52282F2b4f5A91D965D38b64f3BB`

nirmoy:

The PR watcher skips automatic boro reviews for changes with more than 50 commits. This change currently has 100 commits.

```text
review https://github.com/NVIDIA/NV-Kernels/pull/568
```

Head: `6c9283cf8149`

dcemin-nv:
New heads 6c9283cf8149 (#568) and b17ee159dbb4 (#569): same code as v5, the commit message now carries the upstream reference, posted to linux-watchdog on 2026-09-12: https://lore.kernel.org/r/20260912182107.1156221-1-dcemin@nvidia.com. Description updated.

clsotog:
@dcemin-nv  can you rebase this PR. thanks

dcemin-nv:
Rebased onto the current 26.04_linux-nvidia (the base was rewritten upstream, hence the stale commit count and lint result) plus one more change, v6: 568 = e7574bb34d3c, 569 = 617988fa2a96, same file on both.

Boro's High on #569 is valid and distinct from the v5 fix: between register_pm_notifier() and the point where probe read WCS, hw_armed was still false, so a *_PREPARE landing in that window recorded the transition without stopping a firmware-started watchdog, and a probe stalled or frozen past the timeout would have let it fire. v6 reads WCS and sets hw_armed before the notifier is registered, so the hook sees the right state from its first event; the later WCS block only sets WDOG_HW_RUNNING and the lock-protected re-adoption is gone. Commit message updated.

On Boro's Medium (notifier priority): between *_PREPARE and freeze_kernel_threads() the watchdog core's keepalive worker is still running and refreshes the hardware, so an earlier notifier blocking past the timeout during PREPARE does not reset the system; the unrefreshed window this change closes starts at task freezing, after the whole PREPARE chain has run. Raising the priority would also move our restart ahead of other notifiers' post-resume work, which buys nothing. Left at the default; happy to revisit if a concrete ordering problem shows up.

The same v6 change is in the linux-watchdog posting's branch and will go out as v2 together with maintainer feedback rather than as a separate respin.

dcemin-nv:
Round 7 push: commit message only. The platform name in the failure description is replaced by "an arm64 laptop platform", matching the upstream posting. No code change; head is now a991eab687db.

jamieNguyenNVIDIA:
## Summary of remaining issues

- High: Probe can adopt WCS and miss a PM_*_PREPARE already in progress before notifier
  registration, leaving a firmware-enabled watchdog armed and unfed until the later device-PM
  fallback.

- Medium: If PM_*_PREPARE stops a firmware-enabled watchdog and a later probe step fails, devres
  removes the notifier before PM_POST_*, leaving the watchdog disabled after resume.

- Validation: Current-head arm64 build validation is still missing due to the builder
  infrastructure failure.

- Nit: The commit/PR text still overstates the ordering coverage and describes validation of v5
  rather than the current code; strict checkpatch also flags the new 82-character line

dcemin-nv:
v7 pushed: 568 = 1b8c8b595178, 569 = bc8406036107, same driver file on both. Thanks for the summary; both findings are real against v6.

**High (probe adopts WCS after a \*_PREPARE that ran before the notifier was registered).** Agreed. v6 only covered a \*_PREPARE landing after register_pm_notifier() (hw_armed was already set by then). A probe that starts after PM_SUSPEND_PREPARE and before dpm_prepare() blocks probing adopts WCS_EN with sleeping false, registers a notifier that never sees that PREPARE, and left the hardware armed until this device's own suspend callback inside dpm_suspend(). Two changes in v7:

1. After register_pm_notifier(), the probe checks pm_suspend_in_progress() under gwdt->lock and, if it is set, records sleeping and stops the hardware right there. pm_suspend_target_state is set at the top of suspend_devices_and_enter(), that is after suspend_prepare() has run the notifier chain and frozen tasks, and it is cleared at the end of that function, before suspend_finish() sends PM_POST_SUSPEND. So a transition seen by the check is always one whose PM_POST_SUSPEND the now registered notifier receives, and sleeping cannot get stuck. The check runs under the lock so a PM_POST_SUSPEND racing it either runs before the critical section (the cleared state is then visible and the check reads false) or after it (it finds sleeping set and restarts). I deliberately did not use pm_sleep_transition_in_progress(): its hibernation half, hibernation_in_progress(), is true from hibernate_acquire() to hibernate_release(), which includes the tail after PM_POST_HIBERNATION; a probe reading it there would record a transition nobody ends and leave the watchdog stopped for good.

2. The device fallback moves from ->suspend to ->prepare. dpm_prepare() runs wait_for_device_probe() and device_block_probing() first, then every device's ->prepare, and dpm_suspend() starts only after that. A probe that adopted the watchdog in the interval the check cannot see (between PM_SUSPEND_PREPARE and the end of task freezing, before pm_suspend_target_state is set; and hibernation entry) is therefore stopped before any device suspend callback runs, instead of at this device's turn inside dpm_suspend(). That interval is the remaining exposure and the commit message now says so: task freezing plus the ->prepare callbacks of the devices ahead of this one, bounded because dpm_prepare() waits for the probe. Closing it completely would need a state that is set before PM_SUSPEND_PREPARE and cleared after PM_POST_\*; the only one is system_transition_mutex, whose locked state is also produced by non-PM users and persists past PM_POST_\*, so a false positive there would again leave the watchdog stopped. I chose not to go there. ->prepare has no ->complete counterpart, so as before nothing except PM_POST_\* re-arms the hardware.

**Medium (probe fails after \*_PREPARE stopped a firmware-started watchdog).** Agreed: devres unregistered the notifier on the failure path, PM_POST_\* never reached the device, and the hardware stayed stopped. The devres action now re-applies the driver state to the hardware after unregistering the notifier: if sleeping was set it is cleared and, with hw_armed true, the hardware is started again (it calls sbsa_gwdt_sleep_restart()). The watchdog is left as the probe found it. To be explicit about what that means: a firmware-started watchdog without a bound driver is armed and unrefreshed, so it resets the system once the timeout elapses, whether the probe failed inside or outside a transition. That is the existing behaviour of a probe failure outside a transition, and I preferred it to a timing dependent silent disable. The early_enable failure path is unchanged (sbsa_gwdt_stop() clears hw_armed, so nothing is re-armed there). The same action runs on unbind, so an unbind during a transition also leaves the hardware following hw_armed.

**Validation and nits.** arm64 compile of drivers/watchdog/sbsa_gwdt.o with W=1 on both heads, no warnings, and again with CONFIG_SUSPEND and CONFIG_HIBERNATION off (the ->prepare pointer goes through pm_sleep_ptr(), so nothing is left unused with CONFIG_PM_SLEEP=n). checkpatch in strict mode reports only the two pre-existing warnings in untouched upstream lines (S_IRUGO, the split MODULE_PARM_DESC string); the 82 column line is wrapped. The commit message now describes the v7 ordering coverage including the remaining interval, keeps the three data points apart (v1 elimination run of 198 suspends, v4 state machine soak of 120 suspends, v7 compile only), drops the Fixes: tag per an internal review request (the upstream v2 will do the same), and carries a [dcemin: v7 of the posted v1: ...] annotation under the lore reference, which is the form the PR lint asks for when the diff differs from the referenced patch. I am not claiming the 120 suspend soak for v7: the v5 to v7 deltas are probe time and unbind time paths a suspend loop does not exercise. A v7 soak on the reference board is queued and I will post the numbers here. For the builder side arm64 validation I can only offer the local object build above until the infrastructure is back.

dcemin-nv:
Message-only push, head 491d5514f8b3 (569: a6a6c2119b31): the `[dcemin: ...]` backport note is now on a single line, which is the form the PR lint's annotation check looks for; the v7 delta description stays in the version paragraph above it. No code change from the v7 heads.

clsotog:
From codex for the high finding from Jamie:
High: Partially addressed, not fully closed.

  The latest code handles some missed-prepare cases:

  - It adopts WCS before registering the notifier: sbsa_gwdt.c:481-482
  - It registers the PM notifier immediately after: :489-497
  - It does a suspend catch-up with pm_suspend_in_progress(): :511-517
  - It adds a device ->prepare fallback: :430-438

But there is still an acknowledged window: if PM_*_PREPARE already ran and the probe lands before pm_suspend_in_progress() becomes true, or for hibernation/restore where the code deliberately avoids the hibernation in-progress helper, the watchdog remains armed until the device ->prepare callback. So this addresses the “missed prepare” case by bounding it earlier than old ->suspend, but it does not eliminate the “armed and unfed until later device-PM fallback” part.

Thanks for addressing the medium finding.

dcemin-nv:
@clsotog thanks, the analysis is right about v7: the two intervals it names were the ones the v7 commit message listed as the remaining exposure, and both come from the same root cause, so v8 removes that cause rather than bounding it again. Heads: 568 = 31482832c73d, 569 = 60c9c0c13858, same driver file on both.

**Why the intervals existed.** Every version up to v7 registered the PM notifier from probe. A probe that started after `pm_notifier_call_chain_robust(PM_SUSPEND_PREPARE, ...)` in `suspend_prepare()` (kernel/power/suspend.c) had no record of that transition, and the only in-kernel state a driver can read that says "a suspend is in progress and a PM_POST_SUSPEND is guaranteed to follow" is `pm_suspend_target_state`, which `suspend_devices_and_enter()` sets after `suspend_prepare()` has returned, that is after `suspend_freeze_processes()`. So a probe landing between the notifier chain and the end of task freezing could not learn about the transition, and hibernation has no equivalent state at all (`hibernation_in_progress()` is `!atomic_read(&hibernate_atomic)`, true from `hibernate_acquire()` until `hibernate_release()`, which is after `PM_POST_HIBERNATION`; reading it in that tail would record a transition nobody ends and leave the watchdog stopped for good, which is why v7 did not use it). The other candidates fail the same way: `pm_freezing` is set at the start of `freeze_processes()` but is not exported, so a modular driver cannot use it; `system_freezing_cnt` is also raised by the cgroup freezer; `system_transition_mutex` is held by non-PM users and stays held past PM_POST_\*. Any of them could leave the watchdog permanently stopped on a false positive, so I did not go there.

**v8.** The notifier is registered once, in the driver's init before `platform_driver_register()`, and unregistered after `platform_driver_unregister()` in exit. The transition flag is global, the instances sit on a list, and one spinlock covers the flag, the list and every instance's hw_armed. In probe, adopting WCS_EN into hw_armed, applying the flag (stopping a firmware-started watchdog at once if a transition is recorded) and joining the list happen in one critical section, and `*_PREPARE` and `PM_POST_*` walk the list under the same lock. Consequences per window:

1. Probe after `*_PREPARE` while the driver is loaded (deferred probe, sysfs bind, a second instance): closed. The notifier saw the event, the flag is set, and the probe applies it when it adopts the hardware state. This holds for suspend, hibernation and restore alike, since the flag is set by the notifier and not by a helper.
2. Built-in driver: the notifier is registered at initcall time, before userspace exists, so no probe of any kind can precede a transition it does not know about. No window.
3. Modular driver inserted while a transition is in progress (the driver is a module in this kernel): the init runs `register_pm_notifier()` first, then `platform_driver_register()` probes the already enumerated device synchronously in the same task. Sub-cases by where the PM core is at that moment: (a) before the notifier chain: the notifier sees `*_PREPARE`, nothing to do; (b) after the chain and after `pm_suspend_target_state` was set: the `pm_suspend_in_progress()` catch-up in probe records the transition and stops the hardware; PM_POST_SUSPEND follows the clearing of that state in `suspend_devices_and_enter()`, so the flag is always cleared; (c) after the chain and before the end of task freezing, or hibernation entry: nothing in the kernel tells the probe, and the device `->prepare` callback is the guard. `dpm_prepare()` (drivers/base/power/main.c) runs `wait_for_device_probe()` and `device_block_probing()` and then every device's `->prepare`, and `dpm_suspend()` runs only after `dpm_prepare()` returns, so the hardware is stopped before any device suspend callback. That is the whole remaining exposure: modprobe of this driver overlapping the interval between PM_SUSPEND_PREPARE and the end of task freezing, or hibernation entry, with the stop landing at this device's `->prepare`. modprobe is a user task, so `freeze_processes()` waits for it to leave the kernel before freezing completes, and `dpm_prepare()` waits for the probe in any case. I do not see a driver-level way to shrink 3(c) further without one of the false positives above, and I would rather leave a watchdog armed for the length of task freezing than risk leaving it stopped indefinitely.

The Medium fix is kept in the new shape: the devres action removes the instance from the list and, if a transition had stopped it, re-applies hw_armed, so a failed probe or an unbind leaves the hardware as found. Init failure of the notifier fails the module load; failure of `platform_driver_register()` unregisters the notifier again.

**Validation.** arm64 compile of drivers/watchdog/sbsa_gwdt.o with W=1 on both heads, no warnings, in four configurations: defconfig (built in), the driver as a module, CONFIG_SUSPEND and CONFIG_HIBERNATION off, and the mainline copy of the patch. checkpatch in strict mode reports only the two pre-existing warnings in untouched lines. The commit message describes the v8 ordering and the remaining exposure as above; the validation paragraph is unchanged in substance (v1 elimination run, v4 state machine soak, this version compile tested) and the one-line annotation now reads v8. The reference board soak of this version is still queued.

jamieNguyenNVIDIA:
As David noted, cleanup “re-applies hw_armed, so a failed probe or an unbind leaves the hardware as found.”  Since the driver has relinquished ownership at that point, I think preserving the prior hardware state is acceptable.

clsotog:
Thanks for the changes. 

nirmoy:

nvmochs:
Merged, closing PR.

```
c08b3556ccb2 (nresolute/main-next) NVIDIA: SAUCE: watchdog: sbsa_gwdt: stop the watchdog across the whole system-sleep transition
```

PR #567, opened 2026-09-03
[26.04_linux-nvidia-bos] MediaTek MT8901 audio: SoundWire/ASoC/SOF stack

dcemin-nv:
BugLink: https://bugs.launchpad.net/ubuntu/+source/linux-nvidia/+bug/2166301

MediaTek mt8901 audio stack for MT8901 ACPI laptops, as validated on the reference board and the supported laptop designs: the MediaTek SoundWire manager, the ASoC AFE platform and machine drivers with SDCA support, the SOF/ADSP platform driver, the reliability and enumeration fixes, and the machine table entries for the six validated codec topologies.

Series shape (reworked per review):
- 30 commits, dependency ordered: the MediaTek drivers, a mechanical move of the SoundWire manager into drivers/soundwire/mediatek/, the reliability and enumeration fixes, the six machine table entries, the two system sleep commits (each complete on its own), a fix for a left/right swap on the jack PCM after rapid stop/start cycles (PDI FIFO reset in prepare), and one UBUNTU: [Config] commit at the tail carrying the final annotations. Review fixups are folded into the commits they repair with Co-developed-by attribution. Round 7 (2026-09-14): internal platform, board and vendor names removed from the commit messages, the config annotation notes and four code comments; no functional change.
- All commit messages rewritten as kernel prose; no internal metadata, tracker references, or device identifiers remain; config commits use the UBUNTU: [Config] classification, with the options introduced disabled and enabled by the final commit after the complete stack.
- Every commit passes scripts/checkpatch.pl --strict --git with zero errors. Remaining strict-mode output: six "does MAINTAINERS need updating?" warnings for new files (downstream tree, not applicable) and a handful of style checks in the imported MediaTek code (CamelCase identifiers, macro argument reuse); left untouched to keep the series byte identical with the board validated content.

Dependencies (patches intentionally NOT carried here):
- MediaTek power_wrap driver and client API: the clock control and SOF commits call mtk_send_power_control_req() / mtk_pwrap_dev_probe(). now open as #571; merge #571 first by its owners.
- sdw_utils double put fix (mainline 046173b98de3) for the shared rt1320/rt1321 DAI name: PR #560, first commit; merge #560 before this PR.
- MediaTek pinctrl bus-keeper support: the pin bias commit uses mtk_pinctrl_program_bias_by_gpio(), submitted in PR #564.
[...]

PR #570, opened 2026-09-04
[26.04_linux-nvidia] soc: mediatek: SSPM control interface and power_wrap driver

dcemin-nv:
BugLink: https://bugs.launchpad.net/ubuntu/+source/linux-nvidia/+bug/2167298

Standalone submission of the MediaTek SSPM control interface and the power_wrap device power control driver, previously carried as a prerequisite inside the audio series and split out per review feedback.

Series structure:
1. sspm_ci: the SCMI mailbox transport to the SSPM microcontroller
2. power_wrap core driver (public client header introduced directly at include/linux/soc/mediatek/mtk-pwrap.h)
3. sysfs debug interface
4. client power control request API
5. sysfs coverage for client requests
6. DVFSRC enable via SPM SYSRAM flag
7. AC_BESTPERF mode request during boot
8. usb4 resource release via SSPM
9. config annotations

All commits carry author sign-offs (MediaTek and NVIDIA authors acked). Content is identical to the tree validated on the reference board (audio playback, deep sleep and resume exercised through the client API path).

The audio series PRs #566 and #567 depend on the client power control API introduced here and should merge after this set.


nirmoy:
## BaseOS Kernel Review

> [!TIP]
> ### ✅ Review passed
>
> **Findings:** none

### 🔍 Review artifacts

- 🟢 **Boro review:** [no findings](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/570/latest.html)
- 🔵 **Test plan:** [ready for validation](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/570/test-plan.html)

- 🟢 **`amd64`:** [build report and 4 debs](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/570/debs/amd64/)
- 🟢 **`arm64`:** [build report and 4 debs](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/570/debs/arm64/)

> [!NOTE]

<details>

- Reviewed head: `0697f87f9808`
- Overall status: **passed**

</details>

jamieNguyenNVIDIA:
Hi @dcemin-nv, can we assume that these patches are being prepared for upstreaming as well?

dcemin-nv:
> Hi @dcemin-nv, can we assume that these patches are being prepared for upstreaming as well?

@jamieNguyenNVIDIA  these are MediaTek authored platform drivers for the mt8901 SoC, which is an ACPI only platform with no upstream presence yet, so there is nothing for us to base an upstream submission on today. We are not preparing them for the mailing lists ourselves; MediaTek owns the drivers and we are raising the upstream intent with them. Until that is settled they are correctly carried as SAUCE, and every commit in the set is tagged that way.

Separately, I noticed this branch was cut before three commits landed in 26.04_linux-nvidia and now duplicates them under different SHAs (the i2c mediatek ACPI support, the gpiolib debounce routing and the block bio split fix). I will rebase onto current 26.04_linux-nvidia and force push so the PR shows only the nine power_wrap and SSPM commits described above.

clsotog:
Maybe same findings from Boro but this is what Codex found:
 - High /drivers/soc/mediatek/power_wrap/power_wrap.c:664  drivers/soc/mediatek/power_wrap/power_wrap.c:669: pwrap_write_dvfsrc_flag() and pwrap_write_bestperf_flag() run unconditionally for every NVDA6210 match, but they write hard-coded physical addresses.
    The USB4 path already gates MT8901-only behavior via PLID, which implies this HID can appear on other platforms. These MMIO/SYSRAM writes should be PLID/per-platform-data gated or sourced from ACPI resources before enabling the driver broadly on arm64.
  - High drivers/soc/mediatek/power_wrap/power_wrap.c:628: the USB4 release uses late_initcall() to “run after xHCI”, but the inherited arm64 config has CONFIG_USB_XHCI_PLATFORM=m and CONFIG_USB_XHCI_MTK=m. Built-in power_wrap will run the release before those module drivers bind, so xHCI can still re-power the domain later and defeat the suspend fix. This needs to trigger after the actual USB controller bind, or the relevant xHCI path must be built-in for this flavor.
  - Medium drivers/soc/mediatek/sspm_ci/sspm_ci.c:217: _hw_ack is still a bare flag and the IRQ handler does not match the response token/header. After one transaction times out, a late SET IRQ can arrive after the next call clears _hw_ack and make that next call return success before its command completed. The new token should be checked against the response, or the timeout path should drain/reset the mailbox before accepting later completions.
  - Medium drivers/soc/mediatek/power_wrap/power_wrap_sysfs.c:57 drivers/soc/mediatek/power_wrap/power_wrap.c:355: configs are stored under normalized padded ACPI paths, but most direct sysfs/API lookups compare the caller’s raw string as-is. Normal external paths like \_SB.USB0 will not match stored \_SB_.USB0, even though the usage examples use the unpadded form. Normalize inputs in pwrap_query_dev_config() or at each direct caller.

dcemin-nv:
Authorship corrected per Kaushal: the client power control API and the sysfs exposure commits are Mayuresh Wani's work and now carry him as author with his Signed-off-by at the head of the chain; Kaushal and I remain as sign-offs for the carry. No content change; the branches were also refreshed onto the current base tips.

dcemin-nv:
Thanks Nirmoy and Cristian. New heads: 570 = 70b355a4fe09, 571 = fa1849cd8855, still 9 commits, each fix folded into the commit it repairs. Going through the findings:

- acpi_get_name() buffer leak on the pwrap_config_resource() failure path: fixed. The path buffer is now released by a devm action registered right after acpi_get_name(), so every probe exit frees it.
- Raw pointers into dev_ctrl.configs[] outliving remove(): the driver is built in, the NVDA6210 ACPI device is never hot removed and bind/unbind is suppressed, so remove() could only run in a configuration that does not exist here. Rather than keep a remove callback whose only effect is to invalidate pointers consumers hold, it is removed and the comment on the driver states that the table lives for the lifetime of the kernel. A refcounted handle would be the long term shape if this ever becomes a module; noted for the MediaTek owner.
- pwrap_write_dvfsrc_flag() / pwrap_write_bestperf_flag() writing fixed physical addresses for every NVDA6210 match: fixed. Probe reads the PLID platform id once and both writes are gated on MT8901, the same gate the USB4 release already used; other platforms log and skip. Sourcing the addresses from ACPI instead of constants is a firmware contract change and goes to MediaTek.
- Handler zeroing the word at +0x4 before the caller reads a status: per the vendor transport that word is the mailbox ownership (mem_free) flag, not an SCMI status; TINYSYS SET carries no status word and the reference driver clears it the same way. A comment now says so at the write. If firmware does return a status anywhere for SET, that is a question for MediaTek and I have asked.
- ACPI path normalisation: real, but only for the sysfs debug input. Kernel callers key on acpi_get_name() output, which is already padded. The sysfs helper now normalises the user string before the lookup, so the unpadded form in the usage text works.
- USB4 release from late_initcall running before modular xhci-mtk binds: confirmed (USB_XHCI_MTK=m in this config). Fixing it properly needs a trigger on the xHCI bind (bus notifier) or moving the release into the xhci-mtk driver, both of which are the MediaTek author's call; routed to Mayuresh Wani with the analysis. Note that the current soaks do reach PPS8 with this code, so whether the early release actually costs anything is also being checked on the board.
- sspm_ci_set() polling with mdelay(1) under the mutex, and the transaction independent _hw_ack: both confirmed. These are transport design changes (sleeping poll plus a completion, and matching the reply header token in the handler); routed to Mayuresh as the transport author together with the USB4 item, so they land as one coherent change rather than a patch on top of his code.

Remaining checkpatch --strict warnings (LINUX_VERSION_CODE guards, constant on the left, short Kconfig help) predate this round and sit in the MediaTek commits; left for the author.

clsotog:
One more finding codex, let me know about this one: 
drivers/soc/mediatek/sspm_ci/sspm_ci.c:203: sspm_ci_set() still completes on a bare global _hw_ack, and the IRQ handler sets that flag for any SET interrupt at drivers/soc/mediatek/sspm_ci/sspm_ci.c:256. The code generates a token but never matches the response token. A late IRQ from a previously timed-out transaction can arrive after the next command is triggered and make the new call return success before firmware has acked that new command. That can break suspend/resume or power-control sequencing. The ack path should correlate the completion to the current token/header, or otherwise make stale completions impossible to consume.

dcemin-nv:
Yes, that one is on the list already: it is the last bullet of my previous comment (the transaction independent _hw_ack, together with the mdelay poll and the USB4 release ordering). All three are transport design changes in Mayuresh Wani's code and are routed to him with the analysis, so they land as one coherent change to the transport rather than a patch on top of it. I will link the follow-up here when it exists. Nothing else in this PR is affected by it: the current callers are serialized under the mutex and the soaks reach PPS8, so the exposure is a late ack from a timed-out transaction being consumed by the next call, exactly as Codex describes.

clsotog:
> Yes, that one is on the list already: it is the last bullet of my previous comment (the transaction independent _hw_ack, together with the mdelay poll and the USB4 release ordering). All three are transport design changes in Mayuresh Wani's code and are routed to him with the analysis, so they land as one coherent change to the transport rather than a patch on top of it. I will link the follow-up here when it exists. Nothing else in this PR is affected by it: the current callers are serialized under the mutex and the soaks reach PPS8, so the exposure is a late ack from a timed-out transaction being consumed by the next call, exactly as Codex describes.

@dcemin-nv, so can you rebase to give my ack. that was the last finding from me.

dcemin-nv:
Rebased onto the current 26.04_linux-nvidia, and the Codex and Boro items that are ours fixed in the commits they belong to: 570 = 3dca21592449, 571 = 1ac06fecda22, 9 commits each, driver trees identical, every commit built in sequence on the new base.

- Per-commit build: the first power_wrap commit called pwrap_create_sys_files() two commits before the sysfs commit defined it (Boro's Low, and a link failure at that commit with the driver built in). The call and the declaration now live in the sysfs commit.
- Sysfs path normalisation (Codex, Boro Medium x2): pwrap_dev_probe_store() normalises the path before mtk_pwrap_dev_probe(), and pwrap_dev_config_query_store() stores the padded form so the query show matches the table; unpadded input such as \_SB.USB0 now works on every attribute.
- SSPM mailbox window (Boro Medium): probe rejects a _CRS window smaller than the last mailbox's doorbell words instead of writing past the mapping.
- Single instance (Boro High on the shared mailbox state, Medium on dev_ctrl): a second matching SSPM device is refused with -EBUSY under the same mutex that guards the mapping, and a second matching NVDA6210 is refused while dev_ctrl is populated. One SSPM and one power_wrap per SoC is the hardware; this makes the assumption explicit instead of letting a duplicate ACPI node take over live state.
- Typos (interfac, performace) and the two comments Boro flagged (the kdoc no longer mentions a static table; the lifecycle comment no longer refers to a platform .remove that does not exist).

Still with the MediaTek transport author, unchanged from my previous comment: the _hw_ack correlation (Boro High), the mdelay() poll in sspm_ci_set() (Codex), and the USB4 release ordering against a modular xHCI (Boro Medium). Two more for MediaTek: the acpi_get_first_physical_node() reference (Boro High) is a real API gap, but these are static ACPI platform devices that never hot-remove and the driver suppresses bind/unbind, so there is no removal path to race with today; and whether AC_BESTPERF is meant to be requested on battery as well (Boro Medium) is a product decision for Sanket Diwate, asked, answer goes into the commit message.

Cristian, this is the rebase you asked for.

clsotog:
> Cristian, this is the rebase you asked for.

Thanks for the rebase but my name is not Cristian.

dcemin-nv:
> > Cristian, this is the rebase you asked for.
> 
> Thanks for the rebase but my name is not Cristian.

@clsotog I am so sorry, I got confused with another username on a different PR I have. I appreciate your help in reviewing this PR.

nirmoy:
Tracked in [Launchpad bug #2167298](https://bugs.launchpad.net/bugs/2167298).

clsotog:
For the Boro high findings, do you think Mediatek author can have a solution soon?

dcemin-nv:
clsotog, no ETA from the MediaTek authors yet (routed 9/11, escalated through MediaTek management 9/15), so rather than hold the PR I fixed the remaining Boro and Codex findings here: 570 = 689156a699f5, 571 = aa4d24dc1f2a, still 9 commits, driver trees identical, each change folded into the commit it repairs with the MediaTek author kept as author and a Co-developed-by where the fix is not trivial. Also rebased onto the current 26.04_linux-nvidia tip.

Boro High, physical node lookup: acpi_get_first_physical_node() drops the ACPI device's physical_node_lock before returning its device pointer, so the device could be unbound and released before pwrap_power_ctrl_req_store() took its reference. The sysfs path now walks the physical node list itself and takes the reference while the lock is held (pwrap_get_first_physical_node()), so the pointer it hands to the client API is always one it owns. In "power_wrap: expose client requests in sysfs".

Boro Medium and Codex, SET acknowledgment correlation: sspm_ci_set() publishes the SCMI header it sent before ringing the doorbell and clears it on completion or timeout; the SET mailbox handler reads the echoed reply header back from the slot (offset 0x18, before the slot is handed back to firmware) and completes the wait only when message id, protocol and token match. A late acknowledgment from a timed-out transaction is dropped as stale (counted, dev_dbg) instead of completing the next command. Firmware that echoes token 0 rather than the sender's token is tolerated until the first real token echo is seen, logged once, so a firmware that does not echo cannot turn every SET into a timeout. The reference driver reads the reply message and decodes its header on this interrupt, which is what makes the echoed header available. In "add SSPM control interface".

Also folded: the three W=1 kernel-doc warnings in power_wrap.c. Still with the MediaTek authors as documented follow-ups: the mdelay() poll in sspm_ci_set() (Codex, Medium) and the USB4 release ordering against a modular xHCI (Boro Medium). Both are transport and ordering decisions that belong to the driver owners; I have asked for a date.

Validation so far: sspm_ci.o and the three power_wrap objects cross-compiled W=1 with the arm64 nvidia-64k config, no warnings; checkpatch strict shows no new notes. The suspend/resume soak with this head on the reference board runs next and I will add the numbers here (cycles, deepest power state, scmi timeouts, stale ack counter).

clsotog:
thanks for the changes but codex still complains about this one: USB4 resource release is ordered by late_initcall, not by xHCI/USB4 device readiness. The release runs at drivers/soc/mediatek/power_wrap/power_wrap.c:627, but target config has CONFIG_USB_XHCI_PLATFORM=m, CONFIG_USB_XHCI_MTK=m, and CONFIG_USB4=m in debian.master/config/annotations:15875, so the driver this is trying to wait for may not have loaded or probed yet. This should be triggered from the owning USB controller/client path or gated on actual device probe completion.
Is this one you will need help from Mediatek author?

dcemin-nv:
clsotog, agreed, and no, this one did not need the MediaTek author: 570 = 19763c109b46, 571 = 8c14344d0bfc, 9 commits, trees identical, folded into "power_wrap: release usb4 resources via SSPM".

The release is no longer ordered by initcall level. power_wrap probe (built in) registers a platform bus notifier and counts the xHCI controllers in the ACPI namespace (PNP0D10/PNP0D15 platform devices) that have no driver bound yet; the SSPM release is issued when that count reaches zero, immediately if every controller is already bound, otherwise on the BUS_NOTIFY_BOUND_DRIVER of the last one, so a modular xhci-plat/xhci-mtk loaded any time later still orders correctly. Registering the notifier before counting, both under one mutex, closes the window where a controller binds in between: it is then either excluded from the count or decrements it, never both. A controller that never binds keeps the release pending, which is the pre-driver behaviour, and a late_initcall logs that state once so it is visible. The PLID gate is unchanged and runs first, so non MT8901 platforms arm nothing.

Remaining MediaTek follow-up is now only the mdelay() poll in sspm_ci_set() (Codex, Medium). Build: power_wrap objects W=1 clean with the arm64 nvidia-64k config; checkpatch strict clean. Board run with this head is queued with the suspend soak.

clsotog:
thanks for the changes and I got these findings from codex:
 1. High: USB4 notifier can deadlock during xHCI bind drivers/soc/mediatek/power_wrap/power_wrap.c:720 holds pwrap_usb4_lock while bus_for_each_dev() calls pwrap_usb4_count_unbound(), which takes device_lock() at line 656. The BUS_NOTIFY_BOUND_DRIVER callback takes the locks in the opposite order: driver-core bind path already holds the device lock, then pwrap_usb4_bus_notify() takes pwrap_usb4_lock at line 672. If an xHCI bind races with the initial count, this can deadlock.
  2. Medium/High: SSPM ack correlation still has a stale-ack path drivers/soc/mediatek/sspm_ci/sspm_ci.c:291 accepts token-zero replies until _token_echo_seen becomes true. If firmware never echoes tokens, a late SET ack from a timed-out command still matches the next SET because all SET replies share the same message id/protocol. There is also a narrow stale-IRQ case after the next command rewrites the shared mailbox at line 222, because the handler reads the header from that shared slot.

dcemin-nv:
clsotog, both taken, 570 = e908aa8747de, 571 = f59a835ef710, 9 commits, trees identical.

1. Lock order, you are right, that was an ABBA between pwrap_usb4_lock and the device lock. The collection pass no longer takes any device lock: it reads device_is_bound() and records the still-unbound controllers by identity (device references), and the notifier removes the bound device from that list and releases when it is empty. The racy device_is_bound() read is safe here because the binding side has to take pwrap_usb4_lock in the notifier, which the collection pass holds: a controller that binds during collection is either already attached (left off the list, its notification ignored) or removed by its notification later, never counted twice or missed. The lock order is documented next to the state. In "release usb4 resources via SSPM".

2. Stale ack after a timeout: a timed-out SET now marks the transport as needing a drain, and the next sspm_ci_set() waits up to 100 ms for the late reply before it rewrites the shared slot or publishes a new header, so that reply is dropped as stale with nothing in flight instead of matching the next command (no-echo firmware) or reading the next command's header back from the slot. That narrows both cases to a reply that arrives more than 100 ms after the 10 s timeout and coincides with the next command; closing them completely needs the firmware to echo the token, which the "token echo seen" log will confirm on the board and which I have asked the MediaTek transport owner to confirm. In "add SSPM control interface", commit message updated.

Build: power_wrap and sspm_ci objects W=1 clean with the arm64 nvidia-64k config, checkpatch strict shows no new notes. Board run with these heads is queued with the suspend soak; the USB4 release path will be checked with xhci loaded after power_wrap (dmesg "release armed, waiting for N" then "SSPM released").

clsotog:
Thanks for the changes. 

jamieNguyenNVIDIA:
Re-reviewed current heads (`e908aa8747de` / `f59a835ef710`). The original USB4 ordering issue, ABBA deadlock, PLID gating, and basic ACK correlation have been addressed. I still have two blockers before adding my ACK:

1. In `sspm_ci_ack_matches()`, the token-zero compatibility path does not bootstrap correctly. `_scmi_token` starts at zero, so the first token-zero response takes the exact-match branch and sets `_token_echo_seen = true`. If firmware always returns token zero, the second request uses token 1, cannot enter the fallback, and times out. Please only treat an unambiguous nonzero token match as evidence of token echo, or otherwise clarify that token-zero firmware is unsupported.

2. `pwrap_usb4_track_unbound()` calls `device_is_bound()` without holding the device lock. The driver-core contract explicitly requires that lock. I understand the notifier interlocking argument and agree the original ABBA is gone, but `pwrap_usb4_lock` does not synchronize the underlying driver-core state. Please restructure the scan/recheck so the API is called under the required lock, or obtain driver-core maintainer confirmation that this lockless use is valid.

I would not hold the ACK solely for the documented >100 ms late-ACK residual, the existing `mdelay()` polling, or the AC_BESTPERF clarification. The two issues above are the remaining blockers for me.

dcemin-nv:
jamieNguyenNVIDIA, both fixed: 570 = 0697f87f9808, 571 = 9ecf509fa6d9, 9 commits, trees identical.

1. Token bootstrap, you are right, that was a real bug: the counter started at 0, so a no-echo firmware's first token-0 reply took the exact-match branch and set the echo flag, and every later request (token 1 onwards) timed out. The sender now never uses token 0 (the counter starts at 1 and skips 0 on wrap), and the echo flag is set only on a match with a nonzero expected token. A token-0 reply is therefore unambiguous evidence of a firmware that does not echo, and such a firmware stays on the compatibility path (message id and protocol match) for every transaction, not only the first. Commit message updated. In "add SSPM control interface".

2. device_is_bound() under the device lock: the collection pass no longer runs under pwrap_usb4_lock. probe registers the notifier (mutex held briefly, no device lock), then walks the platform bus with no mutex held, taking each xHCI controller's device lock around device_is_bound(), then takes the mutex once to commit the list. A controller that binds between the registration and the commit is recorded by the notifier (which sees tracking not yet enabled) and is dropped from the collected set at commit time, so it is neither missed nor counted twice. The only lock nesting left is the driver core's own: device lock, then pwrap_usb4_lock in the notifier. In "release usb4 resources via SSPM".

Build: both objects W=1 clean with the arm64 nvidia-64k config, checkpatch strict shows no new notes. The board run with these heads is next (the combined test kernel is built; the suspend soak, the USB4 release ordering with xhci loaded late, and the SCMI counters are on it).

clsotog:
my previous ack will stand. thanks.

jamieNguyenNVIDIA:

jamieNguyenNVIDIA:
Applied to Canonical `resolute:main-next` as a contiguous series (stable patch IDs match):

- `d0785129b673` — NVIDIA: SAUCE: soc: mediatek: add SSPM control interface
- `34cb373fea80` — NVIDIA: SAUCE: soc: mediatek: add power_wrap device power control driver
- `e63f0ed9bf97` — NVIDIA: SAUCE: soc: mediatek: power_wrap: add sysfs debug interface
- `200010248b37` — NVIDIA: SAUCE: soc: mediatek: power_wrap: add client power control API
- `b84471f130ef` — NVIDIA: SAUCE: soc: mediatek: power_wrap: expose client requests in sysfs
- `6081bd9c0786` — NVIDIA: SAUCE: soc: mediatek: power_wrap: enable DVFSRC via SPM SYSRAM flag
- `dbc0d1b56578` — NVIDIA: SAUCE: soc: mediatek: power_wrap: request AC_BESTPERF mode during boot
- `93303d83835a` — NVIDIA: SAUCE: soc: mediatek: power_wrap: release usb4 resources via SSPM
- `b7eb9fa32349` — UBUNTU: [Config] annotations for the MediaTek SSPM and power_wrap drivers

PR #575, opened 2026-09-07
[26.04_linux-nvidia] NVIDIA: SAUCE: ARM FF-A EC services on GB10 — DEN0077A v1.3 support

abhsahu:
Overview
========
This series brings the nvidia-ffa-ec driver in line with the released
ARM DEN0077A v1.3 firmware framework spec, fixes the boot ordering that
kept early ACPI consumers from finding the FF-A bridge, decouples
firmware-declared AML cookies from the FF-A framework's physical
notification-ID space, and papers over transient SPMC busy responses
during early boot.

Motivation
==========
GB10 platforms ship EC services behind the ARM FF-A bridge. Two
firmware generations coexist in the field:

  * MSFT000C — the older OpenDevicePartnership draft spec ([1]).
  * ARML0002 — the released DEN0077A v1.3 spec ([2]) sections 18.7 and
    18.8, with inter-partition setup, the FFH offset-2 packet layout,
    and the cookie / notification-id split. DEN0077A v1.3 also updated
    the outer FF-A Device Properties GUID between ALP3 (daffd814-...)
    and ALP4 (c08c3233-...) drafts. The OpenDevicePartnership released-
    spec companion covering the EC-services layer is at [3]. The FFH
    operation-region layout, which the two packet formats key off, is
    defined by ARM DEN0048D "Arm Functional Fixed Hardware Specification
    (FFH)" v1.3 section 2.3 ([4]).

Before this series the driver only handled MSFT000C, and it probed too
late in boot for consumers that declare _DEP on the FF-A bridge and
evaluate FFH regions from _STA / _INI / _DSM during the initial ACPI
namespace scan. Those consumers ended up permanently missing (battery
absent, lid not detected, UCSI dead) until a manual rebind.

References
==========
  [1] OpenDevicePartnership draft EC-services overview (MSFT000C):
      https://github.com/OpenDevicePartnership/documentation/blob/0f7b6bad77a3eb07b66b66d0e3af718db2ec1c24/bookshelf/Shelf%204%20Specifications/EC%20Interface/src/secure-ec-services-overview.md

  [2] Arm Firmware Framework for Arm A-profile (DEN0077A v1.3):
      https://support.arm.com/documentation/den0077

  [3] OpenDevicePartnership released-spec EC-services overview
      (odp-embedded-controller):
      https://github.com/OpenDevicePartnership/odp-embedded-controller/blob/main/docs/src/specs/ec_interface/secure-ec-services-overview.md

  [4] Arm Functional Fixed Hardware Specification (DEN0048D v1.3):
      https://developer.arm.com/documentation/den0048

BugLink: https://bugs.launchpad.net/bugs/2167300


nirmoy:
## BaseOS Kernel Review

> [!TIP]
> ### ✅ Review passed
>
> **Findings:** none

### 🔍 Review artifacts

- 🟢 **Boro review:** [no findings](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/575/latest.html)
- 🔵 **Test plan:** [ready for validation](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/575/test-plan.html)

- 🟢 **`amd64`:** [build report and 4 debs](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/575/debs/amd64/)
- 🟢 **`arm64`:** [build report and 4 debs](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/575/debs/arm64/)

> [!NOTE]

<details>

- Reviewed head: `f13830bc9b32`
- Overall status: **passed**

</details>

jamieNguyenNVIDIA:
Hi @abhsahu,

1. What is the current read on the specs that this work is based on?  Are they any closer to getting ratified now than they were when this work first began?  Should we assume `nvidia-ffa-ec` will need to be carried as SAUCE perpetually?

2. No action needed yet, but once the review comments here are sorted out, we will need this targeted at `26.04_linux-nvidia-bos` as well.

clsotog:
Some findings with Codex
 1. [P1] drivers/platform/arm64/nvidia-ffa-ec.c:513 leaves ARML0002 X5/X6 zero during notification registration. The inter-partition setup layout expects the service UUID in X5/X6 as well as X7/X8, and the ODP text explicitly says the service being registered is stored in X5/X6. This can make ARML0002 notification setup bind to a null/unknown service or fail. Populate ffa_data.data[1] and [2] with svc_uuid too, unless the firmware intentionally diverges from the spec.
  2. [P1] drivers/platform/arm64/nvidia-ffa-ec.c:775 truncates MSFT000C responses to the request length. The old handler copied back the full FFH rawdata buffer because AML requests can be short, for example a one-byte command, while responses can be much larger. Passing pkt->length as both input and output length will leave most legacy response bytes stale. Keep pkt->length for input, but copy back rawdata_buflen bounded by sizeof(struct ffa_send_direct_data2::data).

jamieNguyenNVIDIA:
In addition to what Carol reported, here is a pass over all nine commits
against f8c10473fd98, with CONFIG_ARM_FFA_TRANSPORT=y and
CONFIG_NVIDIA_FFA_EC=y per debian.nvidia/config/annotations.

One piece of history behind Carol's second finding: the code it replaces is
itself the fix for that exact problem.  6b00b7f1a27d ("NVIDIA: SAUCE: Fix FFH
data response length", LP: #2118357, Fixes: d0038ee1df2b) says "While copying
the data back into ACPI FFH packet, it uses the request length. The response
data can be larger than request length."  So this looks like a reintroduction
of LP: #2118357 rather than a new bug, which may matter for how it is
tracked.  That commit also anticipated this series: "When we move to latest
revision, then we need to copy all data bytes for both request and response."

One correction to the automated review: the use-after-free it flags in
nvidia_ffh_do_ffa_send() is not introduced here.  f8c10473fd98 already dropped
nvidia_ffa_lock before dereferencing ec_dev.  See item 5.

Items 1 to 3 are one design question rather than three.


1) 3d376f4f6183 - the early-scanned bridge gets no parent device

acpi_get_devices() runs before acpi_root is assigned (scan.c:2891), so
acpi_find_parent_acpi_dev() finds no ancestor acpi_device and
acpi_init_device_object() sets device->dev.parent = NULL (scan.c:1815).  The
full scan does not fix this up - acpi_bus_check_add() returns early on an
existing acpi_device.  Three consequences:

  - ARML0002:00 lands at the top of /sys/devices/, and is unreachable by
    acpi_dev_for_each_child() from an ancestor.
  - acpi_device_get_busid() takes its root shortcut and sets
    pnp.bus_id = "ACPI" instead of "FFA0", which is what
    acpi_fwnode_get_name() and /proc/acpi/wakeup report.
  - acpi_init_coherency() loses ancestor _CCA inheritance; arm64 selects
    ACPI_CCA_REQUIRED (arch/arm64/Kconfig:5), so a bridge with no _CCA of its
    own gets coherent_dma = 0.

Would creating acpi_root first, or scanning \_SB before the bridge, avoid
this?


2) bdfa342bca5e - would acpi_honor_dep_ids[] be the better list

The message says "Only ACPI scan-handlers are hooked into the enumeration
loop's acpi_dev_clear_dependencies() path".  That function is exported and
called from around twenty drivers, mostly plain platform drivers -
drivers/acpi/ec.c registers a platform_driver (ec.c:2269) and calls it from
acpi_ec_probe() (ec.c:1749).  5195c0329462 in this series does the same for
the battery.

Existing acpi_ignore_dep_ids[] entries were all added because Linux has no
driver for the supplier (2ef33ee7f4f6, 9272e97ae9e9, fa153b7cddce).  The
opposite move has recent precedent: c28527ce5d06 ("ACPI: scan: Honor _DEP for
Intel CVS devices") moved INTC10DE/INTC10E0 into acpi_honor_dep_ids[]
precisely because the driver exists and calls acpi_dev_clear_dependencies()
at the end of probe.  acpi_honor_dep_ids[] already carries ARMH0003.

Would honor_dep_ids plus that call in nvidia_ffa_probe() work here?  If so
the early enumeration pass and the initcall changes both become unnecessary,
which also removes two unguarded full-namespace walks from core ACPI code
that every architecture pays for at boot.


3) 3d376f4f6183 - the arch_initcall move was reverted upstream

0e0546eabcd6 moved ffa_init() to rootfs_initcall; cc7e8f21b9f0 reverted it as
patch 1 of Sudeep Holla's "Move core init to platform driver probe" series:

  "That initcall ordering workaround is not a proper solution and potentially
   conflicts with pKVM FF-A proxy requirement."

That series landed e659fc8e537c and 3acc80a78e45, and this tree already has
it - the diff context shows ffa_probe(struct platform_device *pdev).  Is this
re-applying the pattern on top of its replacement?

It also has no effect where ffa_probe() returns -EPROBE_DEFER under pKVM,
since deferred probe waits for late_initcall (drivers/base/dd.c:376).  Note
the automated review raised this hunk on modular-build grounds; those do not
apply to the nvidia and nvidia-64k flavours, which build both symbols in, but
the pKVM case applies to a built-in kernel.


4) 5c61f9069b27 - notification IDs are never released

nvidia_ffa_next_notif_id only increments, is not rolled back on the error
path, and is not returned by nvidia_ffa_remove_notifications().  The message
says relinquish "is sufficient for teardown", but that releases the FFA
registration, not the counter.  Built-in means no module unload, so after a
cumulative 64 notifications across all bind cycles every probe fails -ENOSPC
until reboot.  Worth a bitmap instead?


5) 5c61f9069b27 - the guard added earlier in the series is removed

36b0c013a703 added guard(mutex)(&nvidia_ffa_lock) to nvidia_ffh_do_ffa_send()
with a comment naming the use-after-free it prevents; this commit replaces it
with a lock covering only the list walk.  That matches f8c10473fd98 so it is
not a net regression, but the commit message does not mention it.  Was it
deliberate?  5cf2f49a1226 then adds a usleep_range() retry inside the same
unlocked region.


6) Commit message and comment accuracy

36b0c013a703 "released-spec support" / "ARM has released the framework":
Arm's version index for den0077 lists only ALP0 to ALP5 for v1.3, latest ALP5
(2026-07-14).  The last released FF-A is 1.2 REL0, whose appendix stops at
18.6.  Could this say "v1.3 ALP5"?

36b0c013a703: both FFH offsets attributed to DEN0048D section 2.3, which
defines 0x0/0x1/0x2 and reserves the rest.  Offset 4 is from the ODP draft.

36b0c013a703: the packet comment says recvid is "(OS populates)" and status
carries a framework error code; nothing touches recvid and status is set to
0 or 1.  The MSFT000C URL in the comment also 404s now.

3d376f4f6183: "from rootfs_initcall (level 6)" - the parent has
module_init(ffa_init), which is device_initcall; rootfs_initcall is level 5.

5195c0329462: beyond the acpi_battery_add() and acpi_driver points already
raised, "ACPI scan handlers (e.g. the FFA provider) call it explicitly in
their .attach" is also wrong - nvidia-ffa-ec.c is a platform_driver, not a
scan handler, and never calls acpi_dev_clear_dependencies().

c18cae7a77fe: ffa_setup_partitions() does register a device for every
partition regardless of id_table.  What is missing is the bind.


7) Smaller items

36b0c013a703: struct nvidia_ec_ffa_packet keeps its name for the new 144-byte
layout while the old one is renamed, so offsetof(..., rawdata) silently goes
18 to 32 and ->status u8 to u64.  Would _arml0002 be safer for backports?

3d376f4f6183: acpi_scan_one_handle_cb() discards acpi_bus_scan()'s return and
both acpi_get_devices() statuses, so enumeration failure is silent; the
helper could be __init.  The diff also has an unrelated hunk removing the
blank line between ffa_probe() and ffa_remove().

5cf2f49a1226: only -EBUSY is retried, while FFA_RET_RETRY and
FFA_RET_NOT_READY both map to -EAGAIN and fall through.  Deliberate for the
GB10 secure partition?

b636ed36a3d3 leaves a double blank line in nvidia_ffa_ec_service_probe().


Checked and found clean

9d43c550296b is a true no-op - the compiled .rodata is identical before and
after.  c18cae7a77fe adds eight UUIDs with no duplicates.  Sleeping in
nvidia_ffh_do_ffa_send() is safe: ACPICA exits the interpreter for
non-default address spaces and sync_send_receive2() already sleeps on
FFA_YIELD.  The ARML0002 FFH packet layout matches DEN0048D 2.3.1.2 and its
32 + 8*N rule, and NVIDIA_EC_NOTIF_TUPLE() is self-consistent and correctly
parenthesised - note this was checked for internal consistency only, not
against DEN0077A, so it says nothing either way about Carol's first finding
on the X5/X6 assignment.  Initcall link order is correct (drivers/firmware/
before drivers/platform/, ffa-core.o before ffa-module.o).  Re-entering
ACPICA from the acpi_get_devices() callback is safe (ACPI_NS_WALK_UNLOCK).
The battery acpi_dev_clear_dependencies() call cannot double-decrement on
rebind.  The missing pkg3->package.count check is pre-existing.  Both arm64
debs build clean.

abhsahu:
---

## Commit 1 — ARML0002 / DEN0077A support

### 1.1 clsotog — P1-A — X5/X6 zero during notification registration

> ARML0002 X5/X6 zero during notification registration; inter-partition setup expects service UUID in X5/X6 as well as X7/X8.

**Reply:**

Thanks for the catch. Could you point me to the specific spec / ODP-EC text that pins X5/X6 for the OS-originated notification-registration message? TF-A EL3 SPMC treats X5/X6 as opaque payload — `spmc_main.c` extracts `src_id`/`dst_id` from X1 and forwards X2..X17 to the target SP without inspecting X5/X6 — so the "Sender service UUID" contract for an OS-originated call is SP-side, not enforced by the SPMC or by the DEN0077A text I have. 

---

### 1.2 clsotog — P1-B — `pkt->length` truncates MSFT000C response

**Reply:**

Fixed. Restored `ffh_copy_len = min(rawdata_buflen, sizeof(ffa_data.data))` in `nvidia_ffh_handler_msft000c()` — matches the pre-existing `4668e71ed358` pattern that uses the same `ffh_copy_len` variable name.

---

### 1.3 jamieNguyenNVIDIA — 6a — DEN0077 wording ("v1.3 ALP5" not "released")

**Reply:**

Fixed. Commit message updated to reference "v1.3 ALP5" instead of "released spec".

---

### 1.4 jamieNguyenNVIDIA — 6b — FFH offsets attributed to wrong spec

**Reply:**

Fixed. Dispatcher block comment and commit message split cleanly:
- FFH offset 2 → DEN0048D §2.3 (ARML0002 / DEN0077A ALP5)
- FFH offset 4 → OpenDevicePartnership draft (MSFT000C)

---

### 1.5 jamieNguyenNVIDIA — 6c — Packet comment inaccuracies, broken URL

**Reply:**

Fixed. ARML0002 packet block comment rewritten:
- `status`: AML writes 0 to indicate "no error yet"; the FFH handler sets 1 on error.
- `recvid`: FFA framework populates this; the handler does not touch it.
- Broken URL removed.

---

### 1.6 jamieNguyenNVIDIA — 7a — Struct naming safety for backports

**Reply:**

Renamed `nvidia_ec_ffa_packet` → `nvidia_ec_ffa_packet_arml0002` for symmetry with the existing `_msft000c` struct.

Keeping the `nvidia_` prefix intentionally as a transitional choice — this driver is a placeholder for an upstream-generic FF-A EC-services driver. When submitted upstream in vendor-neutral form we plan to drop the `nvidia_` prefix; keeping ARML0002 as the "base" today reduces the eventual rename diff. 

---

## Commit 2 — probe FF-A early before consumer enumeration

### 2.1 boro (AI) — boro-1 (High) — `_INI` runs before early enumeration

> `_INI` runs in `acpi_initialize_objects()` before `acpi_scan_init()` reaches early enumeration.

**Reply:**

Correct on the ordering. Considered moving the FF-A bridge probe ahead of `acpi_initialize_objects()` (adds `acpi_ffa_bridge_early_scan()` plus pulls `acpi_platform_init()` into `acpi_bus_init()` before `_INI`) and it works, but on the GB10 platforms we support today no `_INI` method touches `\_SB.FFA0.FFAC` or `_AVAL()` — all FFH access happens later from `_STA` / `_DSM`, which our current placement already handles. Leaving the early scan where it is for this SAUCE series to keep the ACPI core change minimal. If a future GB10 platform starts using `_INI` to touch FFA regions, we will promote the enumeration hook.

---

### 2.2 boro (AI) — boro-2 (High) — `arch_initcall` on modular driver

**Reply:**

Fixed. `NVIDIA_FFA_EC` changed from `tristate` to `bool` and now `depends on ARM_FFA_TRANSPORT=y` (or `COMPILE_TEST`). Both the FF-A transport and this driver are guaranteed built-in, so `arch_initcall` behaves as level-3 init as intended.

---

### 2.3 jamieNguyenNVIDIA — item 1 / CodeRabbit finding 3 — Early bridge has no parent device

**Reply:**

Correct — when `acpi_get_devices()` runs from `acpi_scan_init()` before the full `acpi_bus_scan(ACPI_ROOT_OBJECT)`, the FF-A bridge acpi_device is attached with a `NULL` parent instead of being placed under `\_SB`. Considered two alternatives (delaying enumeration, or scanning `\_SB` early), both defeated the point of the change or added substantial early work. Concrete side effects on GB10 are cosmetic sysfs placement only; runtime PM / `_DEP` / removal are unaffected (or covered by companion patches). Leaving as-is.

---

### 2.4 jamieNguyenNVIDIA — item 3 — `arch_initcall` re-applies upstream-reverted pattern

**Reply:**

Fixed (message extended). To close the loop: not the same pattern — upstream revert `cc7e8f21b9f0` undid an IMA/TPM boot-ordering workaround. Our motivation is different: the FFH region handler must be installed before consumer ACPI devices under `\_SB.FFA0` have their `_STA` / `_DSM` evaluated. Neither `module_init` (level 6) nor `rootfs_initcall` (level 5) runs before `acpi_init` (subsys_initcall, level 4); we land at `arch_initcall` (level 3). Downstream SAUCE only.

---

### 2.5 jamieNguyenNVIDIA — 6d — `rootfs_initcall` level

**Reply:**

Fixed. Numbered-list item corrected: `rootfs_initcall` is level 5. Baseline updated to `module_init (device_initcall, level 6)` per the upstream refactor.

---

### 2.6 jamieNguyenNVIDIA — 7b — `acpi_scan_one_handle_cb()` discards return silently

**Reply:**

Fixed. `acpi_scan_one_handle_cb()` now logs failures via `acpi_handle_warn()` and returns `AE_CTRL_TERMINATE` so the walk stops on error.

---

## Commit 3 — add FF-A bridge HIDs to `acpi_ignore_dep_ids[]`

### 3.1 jamieNguyenNVIDIA — item 2 / CodeRabbit finding 1 — `acpi_honor_dep_ids[]` may be better

**Reply:**

Thanks for the pointer — `honor_dep_ids[]` does not work here given the ordering commit 2 introduces, and we have kept `acpi_ignore_dep_ids[]`.

When ARML0002 / MSFT000C are placed in `honor_dep_ids[]`, consumers `_DEP`ing on the bridge get `honor_deps = 1`, blocking enumeration until `dep_unmet` drops to zero. `acpi_dev_clear_dependencies()` (called from the FFA vendor driver's `.probe`) walks an empty consumer list because commit 2 drives our probe before the full `acpi_bus_scan()`. So consumers see `dep_unmet > 0` forever.  I tried this but it didn't help. 

---

## Commit 4 — battery: clear consumer dependencies

### 4.1 boro (AI) — boro-6 — Stale "acpi_driver framework" wording

**Reply:**

Fixed. Code comment reworded to "platform_driver .probe path"; commit message references `acpi_battery_probe()`.

---

## Commit 5 — map AML cookies to physical FFA notification IDs

### 5.1 boro (AI) — boro-3 (High) / CodeRabbit finding 2 — UAF in `nvidia_ffh_do_ffa_send()`

**Reply:**

Fixed for live runtime via a kref refactor:
- Added `struct kref refcnt` on `nvidia_ec_ffa_device`.
- Allocation via `kzalloc` + a release function that runs `kfree`.
- `kref_get` inside the list walk under the mutex; drop the mutex; do the FFA send; `kref_put` at `out_put`.

Full-scope mutex isn't used because `sync_send_receive2()` re-enters our own FFH region handler on the same CPU — full-scope would deadlock on the non-recursive mutex.

Also: commit 10 (`suppress driver bind/unbind`) removes the only live-runtime path into `.remove`. The residual shutdown-teardown race is acknowledged and invisible on our platforms.

---

### 5.2 boro (AI) — boro-4 (Med) — Callback loses cookie mapping during teardown

**Reply:**

Fixed for live runtime by changing `notify_request`'s `cb_data` from `ffa_device *` to `nvidia_ec_ffa_device *`, taking a kref per `notify_request`, and dropping it per `notify_relinquish`. The callback reads the cookie directly off `cb_data` — no list walk.

Residual race: the FFA framework's `handle_notif_callbacks` releases `notify_lock` before invoking the callback (Sudeep's intentional design — avoids AB/BA deadlock when a callback wants to self-relinquish). Considered a SAUCE framework patch to extend that lock scope; declined because it restricts other in-tree callbacks. With commit 10's `suppress_bind_attrs`, the residual is shutdown-only.

---

### 5.3 boro (AI) — boro-7 (Low) — "bitmap" wording vs monotonic counter

**Reply:**

Fixed. Comment reworded, and the monotonic counter was ultimately replaced with a real bitmap per jamie's item 4 below.

---

### 5.4 jamieNguyenNVIDIA — item 4 — Notification IDs never released; 64-cycle exhaustion

**Reply:**

Fixed. Replaced the monotonic counter with `DECLARE_BITMAP(nvidia_ffa_notif_id_map, NVIDIA_FFA_MAX_NOTIFICATIONS)`.

- Allocation: `find_first_zero_bit()` + `set_bit()` paired under the mutex to make the "find-then-set" atomic.
- Release: `clear_bit()` at every `notify_relinquish()` site. `clear_bit` is atomic on its own and does not require the mutex, which also avoids a self-deadlock in the shutdown teardown path (`service_remove` holds the mutex across `remove_notifications`).
- Failure-path release added in both `nvidia_ffa_notification_setup()` and `notify_request`'s error path in `nvidia_ffa_create_notifications()`.


---

### 5.5 jamieNguyenNVIDIA — item 5 — `guard(mutex)` replaced with narrower coverage

**Reply:**

Fixed (folded into the kref refactor). Commit message explains: full-scope mutex is impossible because the send helper and callback both re-enter our own FFH region handler; narrow mutex + kref is the correct combination.

---

## Commit 6 — remove ACPI rescan workaround

### 6.1 boro (AI) — boro-5 (Med) — Rescan may be needed with `CONFIG_NVIDIA_FFA_EC=m`

**Reply:**

The module case cannot occur in this series. Commit 2 changes `NVIDIA_FFA_EC` from `tristate` to `bool` and adds `depends on ARM_FFA_TRANSPORT=y || COMPILE_TEST`, so both the transport and this driver are guaranteed built-in. There is no `=m` arrangement in which `acpi_scan_init()` could run before the FFH region handler is installed.

---

## Commit 7 — inline UUIDs in ffa_device_id tables

_No review findings._

---

## Commit 8 — add OEM FFA service UUIDs

_No review findings._

---

## Commit 9 — retry FFA send on transient failures

### 9.1 boro (AI) — boro-8 (Low) — Retry off-by-one; stray sleep after last attempt

**Reply:**

Fixed. Rewrote as a wall-clock timeout loop (same idiom as `drivers/char/tpm/tpm_crb_ffa.c` and `drivers/firmware/arm_scmi/`):

```c
deadline = ktime_add_ms(ktime_get(), NVIDIA_FFA_SEND_TIMEOUT_MS);
while (true) {
    ret = ffa_ops->msg_ops->sync_send_receive2(...);
    if (ret != -EBUSY && ret != -EAGAIN)
        break;
    if (ktime_after(ktime_get(), deadline))
        break;
    usleep_range(...);
}
```

No stray trailing sleep; no off-by-one on the retry count.

---

### 9.2 jamieNguyenNVIDIA — 7c — Only `-EBUSY` retried; `FFA_RET_RETRY` / `FFA_RET_NOT_READY` bypass

**Reply:**

Fixed. Loop retries on both `-EBUSY` and `-EAGAIN`. The FFA framework error map confirms the coverage:

- `FFA_RET_BUSY` → `-EBUSY`
- `FFA_RET_RETRY` / `FFA_RET_NOT_READY` → `-EAGAIN`

Both transient conditions are now retried within the wall-clock deadline.

---

## Commit 10 — suppress driver bind/unbind (NEW)

Proactive addition motivated by boro-3 / boro-4 residual concerns. No reviewer has flagged this yet; posting the rationale here for completeness.

- Sets `.driver.suppress_bind_attrs = true` on `nvidia_ffa_driver`, `nvidia_ffa_ec_service_driver`, and `nvidia_ffa_control_service_driver`.
- Removes the only live-runtime path into `.remove` (sysfs unbind).
- Combined with the built-in-only build (`bool`, `depends on ARM_FFA_TRANSPORT=y`), the only remaining trigger for `.remove` is the shutdown path introduced upstream by commit `16a3622577cd` ("firmware: arm_ffa: Tear down driver during shutdown"). At that point the system is going down and any residual teardown-race event is invisible.
- Idiom used by 443 mainline drivers for the same "single firmware fixture consumers depend on" reason (drivers/soc/tegra/, drivers/soc/qcom/, drivers/firmware/, drivers/iommu/).

Runtime-validated on the target platform: `/sys/bus/{arm_ffa,platform}/drivers/nvidia-*` show only `uevent` and per-device symlinks — no `bind` / `unbind` attrs.

abhsahu:
Thanks @nirmoy, @jamieNguyenNVIDIA, and @clsotog.

I have addressed all the review comments and updated the PR. I also replied directly to each inline comment with its status.

> Regarding the commit message and code comments referencing the commit hash:

Because the Open Device Partnership specification undergoes frequent updates and the external links often break or shift, there isn't a reliable way to reference specific historical versions of the spec. Since this downstream driver must support both older draft revisions and the current spec, I included the exact commit hash instead. This ensures we can reliably correlate the code with the exact state of the spec at that point in time, especially given that the current Open Device Partnership repositories lack release tags.

> What is the current read on the specs that this work is based on? Are they any closer to getting ratified now than they were when this work first began?

Since the spec is now incorporated into the standard Arm specification, it is much closer to being finalized.

DGX Spark was the first platform to adopt this specification. However, because the DGX Spark launch schedule did not align directly with the specification development timeline, we had to write the firmware against the draft spec version available at the time.

>  Should we assume nvidia-ffa-ec will need to be carried as SAUCE perpetually?

One potential resolution is to update the firmware to align with the latest specification and ship it as part of an upcoming OTA release. Since upstream kernels will not support the older draft specification, we will need to continue carrying nvidia-ffa-ec as SAUCE if we are unable to update the firmware to the final specification.

> No action needed yet, but once the review comments here are sorted out, we will need this targeted at 26.04_linux-nvidia-bos as well.

Understood. I will send out the backport PR targeting 26.04_linux-nvidia-bos as soon as we conclude the review on this PR.

clsotog:
Thanks for the fixes. 
This is the finding by Codex:
P2 drivers/platform/arm64/nvidia-ffa-ec.c:699 drops the last notification-held reference immediately after notify_relinquish(), but FF-A dispatch can already have snapshotted cb_data and then invokes the callback after releasing notify_lock (handle_notif_callbacks()). That means a notification callback can still run with the ec_ffa_dev pointer passed at line 653 after remove continues to line 928 and frees it. Suppressing bind/unbind lowers exposure, but ACPI/device removal still makes this a UAF. The callback data needs a lifetime that is synchronized with already-snapshotted callbacks, or the FF-A notifier core needs to invoke under a lifetime-safe scheme.

abhsahu:
> Thanks for the fixes. This is the finding by Codex: P2 drivers/platform/arm64/nvidia-ffa-ec.c:699 drops the last notification-held reference immediately after notify_relinquish(), but FF-A dispatch can already have snapshotted cb_data and then invokes the callback after releasing notify_lock (handle_notif_callbacks()). That means a notification callback can still run with the ec_ffa_dev pointer passed at line 653 after remove continues to line 928 and frees it. Suppressing bind/unbind lowers exposure, but ACPI/device removal still makes this a UAF. The callback data needs a lifetime that is synchronized with already-snapshotted callbacks, or the FF-A notifier core needs to invoke under a lifetime-safe scheme.

Thanks @clsotog. This is a framework-level race in drivers/firmware/arm_ffa/, not something the nvidia-ffa-ec driver can close on its own.
                                                                                           
  handle_notif_callbacks() snapshots (cb, cb_data) under notify_lock and then invokes cb() after dropping the lock. __ffa_notify_relinquish() clears the hash-table entry under
  notify_lock and returns — but nothing waits for a dispatch that has already snapshotted cb_data to finish. So any driver that frees cb_data-referenced storage after
  notify_relinquish() returns is exposed to the UAF, regardless of the local lifetime discipline it uses.
                                                                                           
In-tree caller is also affected with this:                                                       
                                                                                                                                    
  - drivers/tee/optee/ffa_abi.c — cb_data = struct optee *; optee_ffa_remove() calls notify_relinquish() then destroy_workqueue() then kfree(optee). optee is tristate (module-removable), so this is a live-runtime UAF, independent of and predating our series.
                                              
The correct fix is in the FF-A notifier core, along the lines of the SRCU notifier chain (kernel/notifier.c: srcu_notifier_chain_unregister() uses synchronize_srcu() for exactly
  this drain), netfilter's nf_hook_entries_free() (call_rcu), or free_irq() (synchronize_irq()). Concretely: wrap the snapshot + invoke sequence in handle_notif_callbacks() /
  handle_fwk_notif_callbacks() with srcu_read_lock/srcu_read_unlock, and add synchronize_srcu() in __ffa_notify_relinquish() after write_unlock(&notify_lock). After that,
  notify_relinquish() gains the contract "no callback for this (dev, notify_id) is executing or will begin executing" and freeing cb_data-owned storage becomes safe for all callers.
                                                                                           
Given this issue is not specific to this patch series and needs changes in core layer. So, I will follow-up regarding this separately in upstream.

jamieNguyenNVIDIA:
On the callback use-after-free Carol reported.

The framework analysis is right and the optee example holds up, and an SRCU
drain in the FF-A notifier core looks like the correct long-term fix.  Two
things suggest not shipping the window while that lands.

The framework does protect its own objects against its own dispatch -
ffa_remove() drains before it frees:

	ffa_notifications_cleanup();	/* destroy_workqueue(): drains */
	ffa_partitions_cleanup();	/* only then frees ffa_devices */

That is why cb_data == ffa_dev was safe before this series.  This commit
changes cb_data to storage freed when .remove() returns, which the framework
makes no promise about.  So the generic race predates the series, but does the
exposure in nvidia-ffa-ec?

And one path reaches the free with dispatch fully live and no sysfs involved:
if nvidia_ffa_create_notifications() fails partway, the unwind puts for j < i
and probe does list_del plus the final kref_put, with the FF-A core running
and its workqueue alive.  suppress_bind_attrs does not cover that.

Would a local fix work meanwhile?  Passing ffa_dev as cb_data again and
resolving the cookie under nvidia_ffa_lock is memory-safe, since the walk and
the free share that mutex.  The original complaint was the fabricated cookie
on a lookup miss; skipping the _DSM there addresses it with no framework
change.


Commit message, 9c8eb0e45aea:

	The mutex cannot be held across sync_send_receive2() or across
	acpi_evaluate_dsm() because those paths enter the SPMC / AML on the
	same CPU and can re-enter our own FFH region handler

Correct for acpi_evaluate_dsm().  Does it hold for sync_send_receive2()?
That chain is ffa_msg_send_direct_req2() -> ffa_msg_send_wait_for_completion()
-> invoke_ffa_fn(), a bare SMC plus an FFA_INTERRUPT / FFA_YIELD loop - no
ACPI, no AML, no locks.  The real hazard there looks cross-thread: the
notification kworker takes nvidia_ffa_lock via AML while a holder spins in the
unbounded FFA_RUN loop.  Worth splitting the two cases, since this is the
stated rationale for the narrow-lock plus kref design.


Two earlier items look partly applied:

"v1.3 ALP5" was added, but the subject still says "released-spec support" and
the body still has "ARM has released the framework", "this released stack",
"(released spec)", "(released)" and "Released-spec notification registration".

battery.c: the comment and the acpi_battery_probe() reference are fixed, but
the message still says "ACPI scan handlers (e.g. the FFA provider) call it
explicitly in their .attach".  nvidia-ffa-ec.c is a platform_driver and never
calls acpi_dev_clear_dependencies().


Smaller:

62c56d6df7c7 - on timeout ret is still -EBUSY/-EAGAIN, so the new dev_warn()
is followed by the pre-existing dev_err() for the same event, unratelimited.

62c56d6df7c7 - -EAGAIN also covers FFA_RET_NOT_READY, which can persist, so
every FFH access would cost the full 10ms in that state.  Intended?

9c8eb0e45aea - the comment above list_add in nvidia_ffa_ec_service_probe()
still describes a callback list walk this commit removes.

d57cba359a9f - acpi_scan_one_handle_cb() could be __init.

a76283eece6f - the FF-A core's platform driver does not set
suppress_bind_attrs, so /sys/bus/platform/drivers/arm-ffa/unbind still reaches
both nvidia .remove callbacks.  That path drains first so the race above is
unaffected, but could the message say sysfs unbind of this driver rather than
the only live-runtime path into .remove?


Checked clean: kref accounting is balanced on every path including both error
unwinds; the bitmap is bounds-checked, paired under the mutex, never
double-cleared, and MSFT000C/ARML0002 cannot collide; suppress_bind_attrs does
take effect on struct ffa_driver; AE_CTRL_TERMINATE is handled cleanly and
acpi_handle_warn() is safe there; the retry loop cannot spin forever; and the
Kconfig bool change does not break allmodconfig, since IS_REACHABLE() supplies
stubs.

abhsahu:
Thanks @jamieNguyenNVIDIA. I have updated the PR to address the latest comments. 

---

## Framework-drain shielding + probe-failure UAF (jamie)

> The framework does protect its own objects against its own dispatch — `ffa_remove()` drains before it frees … So the generic race predates the series, but does the exposure in nvidia-ffa-ec?
>
> And one path reaches the free with dispatch fully live and no sysfs involved: if `nvidia_ffa_create_notifications()` fails partway, the unwind puts for `j < i` … `suppress_bind_attrs` does not cover that.

**Reply:**

Done. I have updated this to pass ffa_device again. 

Concretely, in v6's commit 5 (map AML cookies to physical FFA notification IDs):

- `notify_request` `cb_data` is `ffa_dev *` again, not `nvidia_ec_ffa_device *`. That puts callback lifetime back on framework-owned storage — `ffa_notifications_cleanup()` calls `destroy_workqueue()` on the notification workqueue before `ffa_partitions_cleanup()` frees `ffa_device`s, so a snapshotted `cb_data == ffa_dev` is always safe to dereference. No driver-side punch-through.
- The callback resolves `ffa_dev` → `nvidia_ec_ffa_device` by list walk under `nvidia_ffa_lock`, which is also held around `list_del` in both the `.remove` path and the probe-failure unwind path. On a hit it takes a kref on `nvidia_ec_ffa_device` for the `_DSM` duration; on a miss it returns without invoking `_DSM` — no fabricated cookie reaches firmware, which was the whole reason the old code used `cb_data == ec_ffa_dev`.
- The probe-failure unwind path is now safe by the same mechanism: the unwind loop does `list_del` + `kref_put` under `nvidia_ffa_lock`. A dispatch that snapshotted `cb_data = ffa_dev` before the failure will either find the entry live (still on the list, still refcounted) or find nothing and skip `_DSM`. `suppress_bind_attrs` is not load-bearing here.
- Commit 10's message has been reworded to stop calling `sysfs unbind` "the only live-runtime path into `.remove`", and to explicitly note that arm-ffa-core unbind drains the notification workqueue before invoking our `.remove`, so it too is safe by construction. The only path that still needs suppressing is sysfs unbind of our own drivers, which is what commit 10 actually closes.

---

## Commit 5 — sync_send_receive2 mutex rationale (jamie)

> "The mutex cannot be held across `sync_send_receive2()` or across `acpi_evaluate_dsm()` because those paths enter the SPMC / AML on the same CPU and can re-enter our own FFH region handler"
>
> Correct for `acpi_evaluate_dsm()`. Does it hold for `sync_send_receive2()`? … The real hazard there looks cross-thread…

**Reply:**

Corrected in commit 5's message. The two hazards are split now:

* `acpi_evaluate_dsm()` — same-CPU FFH re-entry, self-deadlock on the non-recursive mutex.
* `sync_send_receive2()` — no ACPI, but blocks unbounded inside `ffa_msg_send_wait_for_completion()` on the `FFA_INTERRUPT` / `FFA_RUN` loop while the SP runs. Holding a driver-global mutex across that would stall the notification kworker (and every other FFH consumer) for arbitrarily long, and would risk cross-thread deadlock against callbacks that themselves take `nvidia_ffa_lock` from `acpi_evaluate_dsm()`.

So for both, the patch drops the mutex before the slow path, keeps `ec_ffa_dev` alive with a kref across it.

---

## Commit 5 — comment above list_add (jamie)

> the comment above `list_add` in `nvidia_ffa_ec_service_probe()` still describes a callback list walk this commit removes.

**Reply:**

Comment matches the code in v6. Commit 5's `.probe` publishes the entry on the global list before `notify_request()`, and the callback (`nvidia_ffa_ec_service_notif_callback`) walks `nvidia_ec_ffa_dev_head` under `nvidia_ffa_lock` — same publishing ordering the comment describes. The list-walk-in-callback design is back in v6 as part of the `cb_data = ffa_dev` revert above. Nothing to change.

---

## Commit 2 — acpi_scan_one_handle_cb could be __init (jamie)

**Reply:**

Fixed. `acpi_scan_one_handle_cb()` marked `__init` in v6 — it's only reachable from `acpi_scan_init()`, which is already `__init`, so the linker can discard it after boot.

---

## Commit 9 — dev_warn + dev_err double log on timeout (jamie)

**Reply:**

Fixed. Removed the `dev_warn()` inside the retry loop. On timeout the loop returns the last `-EBUSY` to the caller and the pre-existing `dev_err()` below fires exactly once, not twice.

---

## Commit 9 — -EAGAIN retry burns full 10 ms per FFH access (jamie)

**Reply:**

Fixed. Retry condition tightened to `-EBUSY` only. `FFA_RET_RETRY` / `FFA_RET_NOT_READY` (both mapped to `-EAGAIN`) are no longer retried. Commit message explains why: those indications mean the SP is not yet initialised, which does not apply here — nvidia-ffa-ec is the first driver to talk to the EC SP on these platforms, so a not-ready result is not going to clear on its own and retrying would burn the full timeout on every FFH access while the SP stays not-ready.

---

## Commit 1 — stale "released-spec" wording (jamie)

> "v1.3 ALP5" was added, but the subject still says "released-spec support" and the body still has "ARM has released the framework", "this released stack", "(released spec)", "(released)" and "Released-spec notification registration".

**Reply:**

Fixed. Full sweep across the series:
- Commit 1 subject: "add ARML0002 / DEN0077A v1.3 ALP5 support".
- Body: "ARM has released the framework" → "ARM has published DEN0077A v1.3 ALP5"; "this released stack" → "the v1.3 ALP5 stack"; "(released spec)" → "(DEN0077A v1.3 ALP5)"; "(released)" → "(DEN0077A v1.3 ALP5)"; "Released-spec notification registration" → "ARML0002 notification registration".
- Commits 2 and 3 also had "DEN0077A released spec"; both changed to "DEN0077A v1.3 ALP5 spec".

Series-wide grep for the old wording now returns zero hits.

---

## Commit 4 — stale ".attach" wording (jamie)

> the comment and the `acpi_battery_probe()` reference are fixed, but the message still says "ACPI scan handlers (e.g. the FFA provider) call it explicitly in their .attach". nvidia-ffa-ec.c is a platform_driver and never calls `acpi_dev_clear_dependencies()`.

**Reply:**

Fixed. Message rewritten. The stale line is gone; the new wording explains the ACPI-core split accurately: `acpi_dev_clear_dependencies()` gets invoked automatically for scan handlers via `acpi_scan_attach_handler()`, but a provider that ships as a `platform_driver` has to call it explicitly from its probe path — there is no `acpi_scan_handler` and no `.attach` where the core could do it on the provider's behalf. No implication that nvidia-ffa-ec calls `acpi_dev_clear_dependencies()`.

---

## Commit 10 — "only live-runtime path into .remove" (jamie)

> the FF-A core's platform driver does not set `suppress_bind_attrs`, so `/sys/bus/platform/drivers/arm-ffa/unbind` still reaches both nvidia `.remove` callbacks. That path drains first so the race above is unaffected, but could the message say "sysfs unbind of this driver" rather than "the only live-runtime path into `.remove`"?

**Reply:**

Reworded. Commit 10's message now:

- Explicitly lists the two live-runtime paths into our `.remove` (our sysfs bind/unbind, upstream `arm-ffa`-core unbind).
- Calls out that upstream `arm-ffa`-core does not set `suppress_bind_attrs` and that's upstream policy, not ours to change.
- Says what this patch actually closes: "sysfs unbind of this driver as an admin-triggerable teardown path".
- Explains the arm-ffa-core path is safe by construction — `ffa_notifications_cleanup()` calls `destroy_workqueue()` on the pcpu notification workqueue before invoking our `.remove` callbacks, so by the time our `.remove` runs on that path no callback can still be in flight, so it cannot trigger the teardown races that the earlier patches contain. Those races require `.remove` to run while the FF-A notification workqueue is still dispatching, and sysfs unbind of our own driver is now the only path that would leave that window open.

---

## CodeRabbit — ffa_pdev NULL-deref in callback (line 406)

> callback dereferences `ffa_pdev` (`ACPI_COMPANION(&ffa_pdev->dev)`) before checking whether teardown has already run.

**Reply:**

Fixed. Added an early guard in `nvidia_ffa_ec_service_notif_callback()`:

```c
if (!ffa_pdev)
    return;
adev = ACPI_COMPANION(&ffa_pdev->dev);
if (!adev)
    return;
```

`ffa_pdev` is cleared in `nvidia_ffa_remove()` and the guard bails out cleanly if teardown has already run, so no NULL-deref regardless of how narrow the window is in practice.

---

clsotog:
Thanks for the changes. These are some extra findings:
1. drivers/acpi/scan.c:2890: early-scanning ARML0002 / MSFT000C before the root namespace scan creates those ACPI devices before their ACPI parent/root objects exist. acpi_init_device_object() sets device->dev.parent once, and the later root scan just reuses the existing object, so the bridge stays parentless. This can leave the device in the wrong sysfs/PM hierarchy. The early path should create the needed ancestor/root objects first or otherwise re-parent before attach.
2. drivers/platform/arm64/nvidia-ffa-ec.c:741: the FFH send path takes a kref on nvidia_ec_ffa_device, but not on the underlying struct ffa_device. If the ARM FF-A core unbind/remove path runs while an AML FFH call is in progress, .remove can return and the FF-A device can be released while nvidia_ffh_do_ffa_send() still dereferences ec_dev->ffa_dev. Pin the device with get_device(&cur->ffa_dev->dev) while holding the list lock and put_device() after the send.

nirmoy:
Tracked in [Launchpad bug #2167300](https://bugs.launchpad.net/bugs/2167300).

abhsahu:
> Thanks for the changes. These are some extra findings:
> 

Thanks @clsotog

> 1. drivers/acpi/scan.c:2890: early-scanning ARML0002 / MSFT000C before the root namespace scan creates those ACPI devices before their ACPI parent/root objects exist. acpi_init_device_object() sets device->dev.parent once, and the later root scan just reuses the existing object, so the bridge stays parentless. This can leave the device in the wrong sysfs/PM hierarchy. The early path should create the needed ancestor/root objects first or otherwise re-parent before attach.

I have made change to create the parents.
With latest change, following is the output in DGX Spark

```
ls -l /sys/bus/acpi/devices/MSFT000C\:00
lrwxrwxrwx 1 root root 0 Sep 15 12:06 /sys/bus/acpi/devices/MSFT000C:00 -> ../../../devices/device:00/LNXSYBUS:00/MSFT000C:00
```

> 2. drivers/platform/arm64/nvidia-ffa-ec.c:741: the FFH send path takes a kref on nvidia_ec_ffa_device, but not on the underlying struct ffa_device. If the ARM FF-A core unbind/remove path runs while an AML FFH call is in progress, .remove can return and the FF-A device can be released while nvidia_ffh_do_ffa_send() still dereferences ec_dev->ffa_dev. Pin the device with get_device(&cur->ffa_dev->dev) while holding the list lock and put_device() after the send.

I have added `get_device/put_device()`.

jamieNguyenNVIDIA:

clsotog:
Hi @abhsahu 
These are the findings from Codex: 
- drivers/platform/arm64/nvidia-ffa-ec.c:538: ARML0002 notification setup leaves X5/X6 zero. ffa_send_direct_data2.data[0] maps to X4, so this code fills only X7/X8 with the service UUID. The referenced ODP inter-partition setup table requires the service UUID in X5/X6, and notes the notification service UUID is in X2/X3. As written, ARML0002 notification registration can be rejected or registered without the service identity.

DEN0077A v1.3 Table 18.25 requires X5/X6 for the sender service UUID and X7/X8 for the receiver service UUID. The released ODP EC-services overview also says the service UUID for the notifications is in X5/X6. The PR’s ARML0002 path only fills X7/X8 and leaves X5/X6 zero.

Can you check this one and also the Boro high finding?

abhsahu:
> Hi @abhsahu These are the findings from Codex:
> 
> * drivers/platform/arm64/nvidia-ffa-ec.c:538: ARML0002 notification setup leaves X5/X6 zero. ffa_send_direct_data2.data[0] maps to X4, so this code fills only X7/X8 with the service UUID. The referenced ODP inter-partition setup table requires the service UUID in X5/X6, and notes the notification service UUID is in X2/X3. As written, ARML0002 notification registration can be rejected or registered without the service identity.
> 
> DEN0077A v1.3 Table 18.25 requires X5/X6 for the sender service UUID and X7/X8 for the receiver service UUID. The released ODP EC-services overview also says the service UUID for the notifications is in X5/X6. The PR’s ARML0002 path only fills X7/X8 and leaves X5/X6 zero.
> 

@clsotog 
I have checked this. In this case, OS is the sender. OS won't have any service UUID. 
This sender service UUID is applicable when one service needs to send the notification setup request to another service But in this case, it is from OS to service, so we X5/X6 won't be applicable in this case.

abhsahu:
@clsotog  @nirmoy  Following are status the boro comments.

---

## 🔴 High — commit 5 — "locking order deadlock"

> The notification dispatcher holds notify_lock for reading while invoking nvidia_ffa_ec_service_notif_callback(), which then takes nvidia_ffa_lock. The removal path takes the locks in the opposite order. AB/BA deadlock.

Analysis done. No change needed in code.

The described ordering does not occur — `notify_lock` is released before the callback runs.

`handle_notif_callbacks()` in `drivers/firmware/arm_ffa/driver.c` uses a `scoped_guard()` **block** around only the snapshot of `cb` and `cb_data`:

```c
scoped_guard(read_lock, &drv_info->notify_lock) {
    cb_info  = notifier_hnode_get_by_type(notify_id, type);
    cb       = cb_info ? cb_info->cb      : NULL;
    cb_data  = cb_info ? cb_info->cb_data : NULL;
}

if (cb)
    cb(notify_id, cb_data);      /* invoked with notify_lock released */
```

The `scoped_guard` exits at the closing brace of the block, so `cb(notify_id, cb_data)` is invoked with `notify_lock` released. `nvidia_ffa_ec_service_notif_callback()` therefore takes `nvidia_ffa_lock` with no framework lock held, and the AB/BA ordering does not arise.

This is framework intentional design: the lock is dropped so that a callback may re-enter the notifier API — for example a self-relinquish would deadlock otherwise. The tradeoff is the `cb_data` lifetime window discussed in other findings; that is handled at our end by pinning `ffa_dev` via `get_device()` and taking a kref on the resolved `nvidia_ec_ffa_device` under `nvidia_ffa_lock` before invoking `_DSM`.


---

## 🟡 Medium M1 — commit 1 — "second-probe protocol regression" (reply-only)

> Can probing a second matching ACPI device cause a protocol-selection regression for an existing instance? `ffa_ec_legacy_spec` is overwritten here before the control FFA driver is registered. If that registration fails because the singleton driver is already registered, the error path does not restore `ffa_ec_legacy_spec`, so the original instance subsequently uses the second device's protocol selection.

Analysis done. No change needed in code.

The scenario is prevented in `nvidia_ffa_probe()` before any global write happens — the very first check in the function is a singleton guard:

```c
static int nvidia_ffa_probe(struct platform_device *pdev)
{
    ...
    if (ffa_pdev) {
        dev_err(&pdev->dev, "FFA device already registered\n");
        return -EINVAL;
    }
    ...
    ffa_ec_legacy_spec = acpi_id->driver_data;
    ...
    ret = ffa_driver_register(&nvidia_ffa_control_service_driver, ...);
    ...
}
```

A second matching ACPI device therefore returns `-EINVAL` at the top of `nvidia_ffa_probe()` before `ffa_ec_legacy_spec` is touched, before `acpi_arm64_ffh_update_custom_offset_handler()` runs, before `ffa_pdev` is set, and before `ffa_driver_register()` is called. The `ffa_driver_register()` failure path described in the finding is therefore unreachable from a second-probe flow — `ffa_ec_legacy_spec` cannot be left inconsistent with the first (bound) device's protocol.

At the platform level, GB10 targets ship exactly one FF-A bridge ACPI device (either ARML0002 or MSFT000C, never both), so the scenario also does not arise from the ACPI namespace side. No functional or code change needed.


---

## 🟡 Medium M2 — commit 2 — >8 ancestors or `acpi_bus_check_add()` failure ignored 

> Can this leave the bridge permanently attached to the wrong device-model parent? A namespace with more than eight non-root ancestors makes `acpi_scan_ensure_ancestors()` return `AE_LIMIT` before creating any ancestors. Failures from `acpi_bus_check_add()` are also ignored. Could the helper propagate `acpi_bus_check_add()` failures, and could the callback avoid `acpi_bus_scan(handle)` whenever ancestor creation fails?

Both suggestions applied in the latest commits. `acpi_scan_ensure_ancestors()` now propagates `acpi_bus_check_add()` failures up to the caller, and `acpi_scan_one_handle_cb()` skips `acpi_bus_scan(handle)` and returns `AE_CTRL_TERMINATE` if the ancestor chain could not be constructed (`AE_LIMIT` on a deeper-than-cap namespace, or any `acpi_bus_check_add()` failure). The consequence — "the bridge is not enumerated early on this failure path; consumer ACPI methods that touch FFAC will still return `AE_NOT_EXIST` during the initial scan" — is logged in the same `acpi_handle_warn()` call so the diagnostic is unambiguous.

On GB10 platforms the FF-A bridge sits at `\_SB.<HID>` (depth 1) or `\_SB.FFA0.<HID>` (depth 2), well under `ACPI_SCAN_EARLY_ANCESTOR_MAX = 8`, so the failure path is a defensive backstop for hypothetical deeply-nested namespaces on future platforms rather than something we expect to hit.


---

## 🟡 Medium M3 — commit 2 — `AE_CTRL_TERMINATE` on first match (fixed)

> The namespace does not require ARML0002 or MSFT000C to be unique. With two matching bridge devices, this return stops `acpi_get_devices()` after the first match. Could the callback continue walking so a consumer of the second bridge does not encounter `AE_NOT_EXIST` before the full namespace scan?

**Reply:**

Fixed. `acpi_scan_one_handle_cb()` now returns `AE_OK` on all paths (after successful bridge scan and after ancestor-scan failure) so `acpi_get_devices()` continues walking the namespace. On a hypothetical platform exposing more than one node per bridge HID, every matching bridge is enumerated early rather than only the first, and consumer methods that touch a second bridge's FFAC succeed on the initial `acpi_bus_scan(ACPI_ROOT_OBJECT)` pass rather than aborting with `AE_NOT_EXIST`.

On GB10 platforms today only a single ARML0002 or MSFT000C node exists in the namespace, so the observable behavior is unchanged; the change is a defensive relaxation of an assumption that was not spec-enforced.


---

## 🟢 Low L1 — commit 5 — `struct nvidia_ec_notification` comment (fixed)

> Could this documentation regression be corrected? In the legacy MSFT000C path, `notif_id` is derived directly from the validated cookie rather than being allocated from `nvidia_ffa_notif_id_map`.

**Reply:**

Fixed. The struct comment now explicitly names both paths — bitmap-allocated for ARML0002, cookie-derived for MSFT000C (with a reminder that the cookie there must fit in the 6-bit notification ID space) — and drops the misleading "allocated" wording from the `notif_id` field comment.


---

## 🟢 Low L2 — commit 5 — bitmap locking-invariant comment (fixed)

> Is the stated locking invariant accurate? The `clear_bit()` calls in the notification-setup failure path, `notify_request()` failure path, and create-time rollback run without `nvidia_ffa_lock`. Could the comment instead document the actual synchronization used for the bitmap?

**Reply:**

Fixed. The comment now accurately splits the two sides of the bitmap:

- Allocation (`find_first_zero_bit()` + `set_bit()`) runs under `nvidia_ffa_lock` so the find-then-set pair is atomic against concurrent allocators.
- Release (`clear_bit()`) is atomic on its own and runs without `nvidia_ffa_lock` — from the notification-setup failure path, the `notify_request()` failure path, the create-time rollback, and the remove path.

Removing the mutex from the release paths was intentional (avoids a self-deadlock in the shutdown teardown chain that goes `service_remove -> nvidia_ffa_remove_notifications -> clear_bit`). The comment now spells that split out.


---

## 🟢 Low L3 — commit 5 — callback doc: `nvidia_ffa_lock` covers list_del only (fixed)

> Could this documentation regression be fixed to say that `nvidia_ffa_lock` covers `list_del()` only? Both the removal and probe-failure paths release the mutex before the final `kref_put()`.

**Reply:**

Fixed. The callback doc no longer claims the mutex covers "list_del + the final kref_put" — it now correctly says the mutex covers `list_del` only, and explicitly notes that the final `kref_put()` runs outside the mutex in both the remove path and the probe-failure-unwind path. The safety property preserved by that ordering (list walk either sees a live entry and pins it with a kref, or misses cleanly after `list_del`) is spelled out in the same comment.


---

## 🟢 Low L4 — commit 6 — message typo "the FFH" (fixed)

> Is this sentence missing a predicate such as "handler is registered" after "the FFH"?

**Reply:**

Fixed. Commit 6's message now reads "the FFH handler is installed before any consumer ACPI method runs".


---

## 🟢 Low L5 — commit 7 — "single-line addition" wording (fixed)

> Could "single-line addition" be changed to "single-table edit"? Each entry uses a separate service-name comment and a UUID initializer spanning two lines, so adding a service requires three new lines rather than one.

**Reply:**

Fixed. Commit 7's message now says "single-table edit" — matches the actual mechanics (three lines: `/* service-name */`, `UUID_INIT(...)` line one, `UUID_INIT(...)` line two continuation) while keeping the intended meaning that the change is confined to one table rather than spread across two files.


---

## 🟢 Low L6 — commit 8 — Physical Presence UUID misclassified (fixed)

> Is the first UUID misplaced under the custom OEM comment? It expands to `3dddfaa6-361b-4eb4-a424-8d10089d1653`, which `include/acpi/acuuid.h` defines as `UUID_PHYSICAL_PRESENCE` for the standard ACPI Physical Presence interface. The commit message describes these as "eight OEM-specific service UUIDs" and says "their FF-A service definitions are platform-defined and not published in a public spec." Could the comment and those statements be corrected to avoid this provenance regression?

**Reply:**

Fixed, both code and message:

- The `3dddfaa6-361b-4eb4-a424-8d10089d1653` entry is now grouped under its own comment: `/* Physical Presence (UUID_PHYSICAL_PRESENCE, include/acpi/acuuid.h) */`, separated from the OEM block.
- Commit 8's subject changed to "add OEM and standard FFA service UUIDs".
- Commit 8's body now describes the standard Physical Presence UUID separately, calls out `include/acpi/acuuid.h` and the exact UUID string, and describes the remaining seven entries as OEM-specific. The provenance claim about the OEM block no longer covers the Physical Presence entry.


---

## 🟢 Low L7 — commit 9 — retry loop can send one extra request past deadline (reply-only)

> Can this retry `sync_send_receive2()` once more after the deadline has expired? If `-EBUSY` is returned shortly before the deadline, `usleep_range()` can cross the deadline. The loop then issues another request without checking the time, so the configured retry window can be exceeded.

**Reply:**

The observation is accurate — if `-EBUSY` is returned just before the deadline, the `usleep_range()` sleep can cross the deadline and the loop issues one additional `sync_send_receive2()` before exiting.

`NVIDIA_FFA_SEND_TIMEOUT_MS` is not a hard real-time budget; it is a best-effort retry window for a transient SP-busy condition, and the loop's worst-case overshoot is bounded to one extra request plus one `usleep_range(500, 1000)` interval — sub-millisecond and safely within any consumer's expected FFH-write latency envelope. Restructuring the loop to guarantee no post-deadline request adds a second deadline check for no observable benefit, so we're leaving the code as-is.


---

## 🟢 Low L8 — commit 10 — FFH-handler-remove wording (fixed)

> `ffa_remove()` cleans up notifications and unregisters the FF-A partition devices, invoking the two `ffa_driver` remove callbacks. The FFH handler is instead removed by `nvidia_ffa_remove()`, which is the separate NVIDIA platform driver's remove callback. Unbinding the arm-ffa platform driver does not invoke `nvidia_ffa_remove()`, so the FFH handler remains installed and subsequently returns `-ENODEV` after the EC service list is emptied.

**Reply:**

Fixed. Commit 10's message no longer implies that the FFH region handler is torn down on arm-ffa-core unbind. The revised wording spells out that:

- `ffa_remove()` cleans notifications and unregisters every `ffa_device`, cascading `.remove` into our two `arm_ffa`-bus drivers (`nvidia_ffa_ec_service_driver`, `nvidia_ffa_control_service_driver`).
- The FFH region handler is installed and torn down in `nvidia_ffa_remove()`, the separate NVIDIA platform_driver callback, which arm-ffa-core unbind does not invoke.
- After arm-ffa-core unbind the FFH handler stays installed but the EC service list is empty, so any subsequent FFH access returns `-ENODEV` rather than reaching a live SP.

The lifetime claim used later in the message ("arm-ffa-core drains its notification workqueue before invoking our .remove callbacks, so no callback can still be in flight when .remove runs") is orthogonal to the FFH-handler lifetime and is unchanged.

drivers/platform/arm64/nvidia-ffa-ec.c
```diff
@@ -421,9 +531,71 @@ static int nvidia_ffa_notification_setup(struct nvidia_ec_ffa_device *ec_ffa_dev
...
+	ret = control_ffa_dev->ops->msg_ops->sync_send_receive2(control_ffa_dev,
+							      &ffa_data);
+	if (ret) {
+		dev_err(&ec_ffa_dev->ffa_dev->dev,
+			"Failed to send NOTIFY_SETUP cookie=%u notif_id=%u error=%d\n",
+			cookie, notif_id, ret);
+		return ret;
+	}
+
+	if (ffa_data.data[0]) {
```
nirmoy: Codex found a possible issue in the ARML0002 notification-registration response check: [DEN0077A v1.3 ALP5, Table 18.26](https://documentation-service.arm.com/static/6a57497b84ca37675bcc4d7d#page=340) places the return status in X10[7:0] (`ffa_data.data[6] & 0xff`), but this checks reserved X4 (`ffa_data.data[0]`). The FF-A transport returns success for `FFA_MSG_SEND_DIRECT_RESP2` without checking this protocol status, so a rejected registration could be treated as successful. Could you please recheck this?

nirmoy (commented):
Thanks, the lock-ordering explanation checks out. Codex found one additional issue for recheck, noted inline.

clsotog:
Saw the same as Nirmoy pointed. Let us know.

drivers/platform/arm64/nvidia-ffa-ec.c
abhsahu: > Codex found a possible issue in the ARML0002 notification-registration response check: [DEN0077A v1.3 ALP5, Table 18.26](https://documentation-service.arm.com/static/6a57497b84ca37675bcc4d7d#page=340) places the return status in X10[7:0] (`ffa_data.data[6] & 0xff`), but this checks reserved X4 (`ffa_data.data[0]`). The FF-A transport returns success for `FFA_MSG_SEND_DIRECT_RESP2` without checking this protocol status, so a rejected registration could be treated as successful. Could you please recheck this?

Thanks @nirmoy and @clsotog 
I have fixed this and updated PR.
It seems, the X4 status check got propagated from older draft spec where X4 contains the status.

jamieNguyenNVIDIA:
Re-sending my ACK:
```
Acked-by: Jamie Nguyen <jamien@nvidia.com>
```

nirmoy:

nvmochs:
Merged, closing PR.

```
d884f0eee50c (nresolute/main-next) NVIDIA: SAUCE: platform/arm64: nvidia-ffa-ec: suppress driver bind/unbind
d18cc1f38920 NVIDIA: SAUCE: platform/arm64: nvidia-ffa-ec: retry FFA send on transient failures
880f081fc2f1 NVIDIA: SAUCE: platform/arm64: nvidia-ffa-ec: add OEM and standard FFA service UUIDs
efc3eaa9b1dc NVIDIA: SAUCE: platform/arm64: nvidia-ffa-ec: inline UUIDs in ffa_device_id tables
62d8e36653eb NVIDIA: SAUCE: platform/arm64: nvidia-ffa-ec: remove ACPI rescan workaround
19132ad49a27 NVIDIA: SAUCE: platform/arm64: nvidia-ffa-ec: map AML cookies to physical FFA notification IDs
edca2f1d3f46 NVIDIA: SAUCE: ACPI: battery: clear consumer dependencies after successful probe
b12ed21a5e30 NVIDIA: SAUCE: ACPI: scan: add FF-A bridge HIDs to acpi_ignore_dep_ids[]
3e850853c3ab NVIDIA: SAUCE: ACPI / firmware / platform: probe ARM FF-A early before consumer enumeration
568512ec10de NVIDIA: SAUCE: platform/arm64: nvidia-ffa-ec: add ARML0002 / DEN0077A v1.3 ALP5 support
```

jamieNguyenNVIDIA:
@abhsahu : can you please make sure to target this one to `26.04_linux-nvidia-bos` as well?

jamieNguyenNVIDIA:
> @abhsahu : can you please make sure to target this one to `26.04_linux-nvidia-bos` as well?

My apologies.  You already took care of this: https://github.com/NVIDIA/NV-Kernels/pull/597

PR #571, opened 2026-09-04
[26.04_linux-nvidia-bos] soc: mediatek: SSPM control interface and power_wrap driver

dcemin-nv:
BugLink: https://bugs.launchpad.net/ubuntu/+source/linux-nvidia/+bug/2167298

Standalone submission of the MediaTek SSPM control interface and the power_wrap device power control driver, previously carried as a prerequisite inside the audio series and split out per review feedback.

Series structure:
1. sspm_ci: the SCMI mailbox transport to the SSPM microcontroller
2. power_wrap core driver (public client header introduced directly at include/linux/soc/mediatek/mtk-pwrap.h)
3. sysfs debug interface
4. client power control request API
5. sysfs coverage for client requests
6. DVFSRC enable via SPM SYSRAM flag
7. AC_BESTPERF mode request during boot
8. usb4 resource release via SSPM
9. config annotations

All commits carry author sign-offs (MediaTek and NVIDIA authors acked). Content is identical to the tree validated on the reference board (audio playback, deep sleep and resume exercised through the client API path).

The audio series PRs #566 and #567 depend on the client power control API introduced here and should merge after this set.

PR #611, opened 2026-09-19
[26.04_linux-nvidia-bos] MediaTek PCIe Power Wrap PM coordination

bdintakurti-nv:
Add MediaTek PCIe Power Wrap PM coordination. 

The change:
  - Registers PCIe Root Ports with a Power Wrap controller shared by all ports under the same ACPI host bridge.
  - Requests segment D3 only after every live Root Port on the bridge has completed noirq suspend, while leaving
  populated or incompletely tracked segments powered.
  - Restores the segment to D0 before PCI noirq resume accesses configuration space or controller MMIO.
  - Applies the same ordering to runtime PM and handles request failures and ambiguous D3 timeouts safely.

Dependency (patches intentionally not carried here):
  - MediaTek SSPM control interface and Power Wrap driver: https://github.com/NVIDIA/NV-Kernels/pull/571

This PR does not build standalone until #571 lands because it uses the Power Wrap client interface introduced there.

MediaTek owns the upstream work, I will coordinate with the original MediaTek authors. Until then, this change is carried as
  SAUCE.

Validation:
  - Built and tested with the required Power Wrap dependency in the integration tree.
  - S3/S2Idle validation
  - PCI topology, NVMe storage, networking, and GPU remained functional after resume.
 
 Launchpad bug: https://bugs.launchpad.net/ubuntu/+source/linux-nvidia-bos/+bug/2167887


nirmoy:
## BaseOS Kernel Review

> [!TIP]
> ### ✅ Review passed
>
> **Findings:** none

### 🔍 Review artifacts

- 🟢 **Boro review:** [no findings](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/611/latest.html)
- 🔵 **Test plan:** [ready for validation](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/611/test-plan.html)

- 🟢 **`amd64`:** [build report and 4 debs](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/611/debs/amd64/)
- 🟢 **`arm64`:** [build report and 4 debs](https://baseos-internal-tools.nvidia.com:4001/reviews/NVIDIA/NV-Kernels/611/debs/arm64/)

> [!NOTE]

<details>

- Reviewed head: `5390149668d6`
- Overall status: **passed**

</details>

jamieNguyenNVIDIA:
The following applies to both #610 and #611:

1. **High:** `mtk_pci_pwrap_host_rp_count()` counts only Root Ports. A host bus containing a managed empty Root Port plus an active RCiEP, RCEC, or other directly attached function can therefore satisfy the topology/count checks and gate the segment while that untracked function remains active. Segment gating must reject any untracked root-bus function, not only untracked Root Ports.

2. **High:** Managed ports lose `DPM_FLAG_SMART_SUSPEND`, forcing `pci_pm_suspend()` to runtime-resume them. If the new Power Wrap D0 request fails, `pci_pm_runtime_resume()` returns an error, but `pci_pm_suspend()` ignores it and continues into configuration-space accesses against a potentially gated segment. Freeze and poweroff contain the same unchecked pattern. These transitions should abort when runtime resume of a managed port fails.

3. **High:** If Power Wrap D0 fails in `pci_pm_resume_noirq()`, that callback returns before restoring PCI state. The PM core nevertheless continues into normal resume, which accesses configuration space without another D0 attempt. The controller also remains recorded as D3 with stale suspended-port accounting. Recovery must guarantee controller-wide D0 and reconcile the port state before later PCI resume phases.

4. **Medium:** `mtk_pci_pwrap_suspend()` treats every `-ETIMEDOUT` as an ambiguous completed D3 transition. The SSPM dependency returns that errno both before submitting the request and after submission while waiting for acknowledgement. The API needs to distinguish “not submitted” from “completion unknown”; only the latter can safely use the ambiguous-D3 handling.

5. **Medium:** `mtk_pwrap_dev_probe()` returns a non-NULL configuration for `CTRL_NONE` and `CTRL_BY_PWRAP`, while the suspend/resume dispatch rejects every non-SCMI configuration with `-EINVAL`. Those configurations are therefore registered as managed and subsequently abort system suspend. Discovery should expose the required capability, or this code should register only controllers that support the D-state operations.

Thank you for confirming that you will work with MediaTek on upstreaming this. Please prioritize that upstreaming effort.

bdintakurti-nv:
Updated with the PCIe-specific fixes:
 - Reject segment gating when any untracked direct root-bus function is present, and serialize the final topology validation with the D3 transition.
 - Restore managed Root Ports before PCI configuration access and propagate runtime-resume failures through suspend, freeze, and poweroff.
 - Keep tracked-port accounting consistent across topology changes.
 
The remaining transition-state ambiguity and capability handling originate in the shared Power Wrap/SSPM contract and also affect the USB, SPI, and I2C clients. We will work with MediaTek to address the contract in Power Wrap first, then update all clients consistently. 
Would it be OK handling that as a follow-up change?

jamieNguyenNVIDIA:
Hi @bdintakurti-nv,

Thanks for the update. It seems to me that the two remaining High issues—the failed D0 recovery during `resume_noirq` and the sysfs rescan race with segment gating—can be addressed independently of the shared Power Wrap/SSPM contract. Can these be addressed in the same PRs?

I’m okay with deferring the two Medium contract issues to a follow-up. Could you please create and link a tracking bug covering both?

jamieNguyenNVIDIA:
Findings from Codex — the following applies to both #610 and #611:

1. **High (new):** The x86-64 defconfig build fails in `pci_device_shutdown()`: it calls `mtk_pci_pwrap_resume()` when `CONFIG_MTK_POWER_WRAP=n`, but `pci.h` provides no stub for that configuration. I reproduced the implicit-declaration error. Could you guard the call or add a stub?

2. **High (remaining):** The failed-D0 handling now guards PCI resume, but a later driver unbind still invokes `.remove()` after `pm_runtime_get_sync()` without checking whether the segment was restored. For example, xHCI removal reaches BAR MMIO while the segment may still be gated. Marking the devices disconnected prevents PCI config accesses, but does not prevent those MMIO accesses. Could this path be addressed in these PRs as well?

The sysfs rescan race appears fixed. The two previously discussed Medium issues can remain deferred; please link the follow-up tracking bug when it is created.

